PR #1369, opened 2017-11-08
add project workbook 

Allicolyer:
Hey Joe! I've added the project workbook to both the mentor and student dashboards. Can you help me set up how to track how many people open these?


Allicolyer:
Hey Joe! So we were able to set up google analytics for this on the wordpress, and just need to track views here now.

PR #1397, opened 2017-12-11
Remove past teams list from my teams page

RRios00:


joemsak:
Looks great, @RRios00! Thanks!

PR #1399, opened 2017-12-12
add curriculum links to submission pieces

Allicolyer:


joemsak:
@Allicolyer Could I get you to try that fancy new reflex padding stuff you and @RRios00 found, instead of using `<br>` tags for spacing? :)

PR #1406, opened 2018-01-04
Admin view submission

Allicolyer:
Hey Joe,

Here is the progress so far on this.

Still left to do:

- Fix the team's division
- Only show the business plan field if the team is a senior division team
- Create a way to view the code checklist (this will be helpful when selecting finalists and to assess if the code checklist is working)
- Delete all other necessary view code (I want to see what you think of this first before doing anything too drastic)
- Fix the edit submission form


joemsak:
Merge conflict woes

PR #1413, opened 2018-01-12
Admin view submission

Allicolyer:
Hey Joe I cleaned up the admin submission view a bit more

### DIFF
--- a/app/models/team_submission.rb
+++ b/app/models/team_submission.rb
@@ -301,6 +301,10 @@ class TeamSubmission < ActiveRecord::Base
     end
   end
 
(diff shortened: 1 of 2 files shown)


joemsak:
Hey @Allicolyer so this PR is fine, however, there is a ton of `if / else` happening in the view template and I believe that `TeamSubmission` has a number of custom getter methods that can help us avoid that. And if not, we probably should add them so that we can "Tell" our code to do things rather than "Ask" if we can do things. In the meantime, look up "Tell, don't ask"

joemsak:
Thanks @Allicolyer!

PR #1423, opened 2018-01-24
Admin edit submissions

Allicolyer:
Hey Joe,

Here's my work so far. I am getting a bit stuck on being able to upload a business plan. It doesn't seem like the cache is working on the edit form.


app/models/team_submission.rb
```diff
@@ -47,6 +47,8 @@ class TeamSubmission < ActiveRecord::Base
     end
   }
 
+  mount_uploader :business_plan, FileProcessor
```
joemsak: I think this should not be here, because `BusinessPlan` is an active record model and `TeamSubmission.has_one :business_plan` -- is this actually working? Anyway, it should be using the has_one version, not an uploader version.

app/views/admin/team_submissions/show.html.erb
```diff
@@ -57,10 +57,13 @@
 
       <% if @team_submission.senior_division? %>
         <h5>Business plan</h5>
-        <% if @team_submission.business_plan_url %>
+
+        <% if @team_submission.business_plan_identifier %>
           <%= link_to @team_submission.business_plan_url,
             @team_submission.business_plan_url,
             target: :_blank %>
+        <%else %>
```
joemsak: `<% else %>` a little space is needed if you please :)

app/views/admin/team_submissions/edit.html.erb
```diff
@@ -1,51 +1,55 @@
...
-    collection: TeamSubmission.development_platforms.keys %>
+        <% if @team_submission.business_plan_url %>
+          <%= link_to @team_submission.business_plan_url,
+            @team_submission.business_plan_url,
+            target: :_blank %>
+        <% end %>
 
-  <%= f.input :development_platform_other %>
+        <%= f.file_field :business_plan %>
+        <%= f.hidden_field :business_plan_cache %>
```
joemsak: this should be done with the `fields_for` form helper and the has_one association to business plan, not as a carrierwave uploader

PR #1439, opened 2018-02-17
rearrange submission to match judging rubric

Allicolyer:
Hey Joe,

Is it too late to make this change? I wanted to make the student screens match the judging rubric. After talking with Ruby, I realized the judging screens are going to match the judging rubric, so this gives the students the most consistency.

### DIFF
--- a/app/models/submission_section.rb
+++ b/app/models/submission_section.rb
@@ -2,17 +2,17 @@ class SubmissionSection
   SECTION_NAMES = {
     0 => "start",
     1 => "ideation",
-    2 => "marketing",
-    3 => "code",
-    4 => "business",
+    2 => "technical",
+    3 => "pitch",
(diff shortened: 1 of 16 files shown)


joemsak:
@Allicolyer I think it might be confusing to switch it after some students and mentors have already seen it a certain way

I'm also unsure about the change of the words from simpler to more difficult english words

And lastly, I fixed the failing specs this time around but you should be running the tests before making a PR and fixing any that you broke

We should discuss my first two points before merging and deploying. Let me know what you think

Allicolyer:
Hey Joe,

That makes sense. I know I am late with these edits. I am trying to match the students experience in submitting to the judge's experience in judging: [judging rubric](http://technovationchallenge.org/wp-content/uploads/judging-rubric.pdf)

PR #1466, opened 2018-03-20
Judge RPE view

Allicolyer:
Hey Joe,

What do you think of this so far.


app/views/judge/dashboards/onboarded/_events.en.html.erb
```diff
@@ -1,7 +1,32 @@
...
-</div>
+<li class="completion-step">
+    <div class="flags">
+      <span class="flag flag-events">
+        Judging
+      </span>
+    </div>
 
-<%= render 'judge/dashboards/info_sidebar' %>
+  <% if SeasonToggles.select_regional_pitch_event? &&
```
joemsak: I dont think we need to have the season toggle here because the judge does not select an event, and when we turn it off, it would disappear even if they're assigned to one

Allicolyer:
localhost_3000_judge_dashboard 1
localhost_3000_judge_dashboard

Allicolyer:
@joemsak

PR #1501, opened 2018-05-07
add column for complete versus incomplete scores

Allicolyer:
Request from Katie and Amy. They also requested both columns to be mandatory.

### DIFF
--- a/app/data_grids/scored_submissions_grid.rb
+++ b/app/data_grids/scored_submissions_grid.rb
@@ -33,7 +33,7 @@ class ScoredSubmissionsGrid
     scores.current_round.complete.count
   end
 
-  column :incomplete_scores mandatory: true do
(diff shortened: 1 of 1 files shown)


app/data_grids/scored_submissions_grid.rb
```diff
@@ -29,10 +29,15 @@ class ScoredSubmissionsGrid
 
   column :contest_rank
 
-  column :num_scores, header: "# of Scores", mandatory: true do
-    scores.current_round.count
+  column :complete_scores, mandatory: true do
+    scores.current_round.complete.count
   end
 
+  column :incomplete_scores mandatory: true do
```
joemsak: please test the page, this line has a syntax error

Allicolyer: Okay fixed it

PR #1504, opened 2018-05-09
create table for judge-scores

Allicolyer:
Hey Joe! I need some guidance on where to write put some methods for this. Right now I am just calling the models in the views (which I know is bad practice)

### DIFF
--- a/app/controllers/admin/participants_controller.rb
+++ b/app/controllers/admin/participants_controller.rb
@@ -111,9 +111,9 @@ module Admin
 
     def submission_score(account)
       if account.judge_profile.present?
-        SubmissionScore.where(judge_profile_id: account)
+        account.judge_profile.submission_scores.current.complete
       else
-        false
+        SubmissionScore.none
       end
     end
 
--- a/app/controllers/admin/participants_controller.rb
+++ b/app/controllers/admin/participants_controller.rb
@@ -9,6 +9,7 @@ module Admin
     def show
       @account = Account.find(params.fetch(:id))
       @teams = Team.current
+      @scores = submission_score(@account)
     end
 
     def edit
@@ -107,5 +108,14 @@ module Admin
         false
       end
     end
+
+    def submission_score(account)
+      if account.judge_profile.present?
+        SubmissionScore.where(judge_profile_id: account)
+      else
+        false
+      end
+    end
+
   end
 end
--- a/app/views/admin/participants/_judge_scores.html.erb
+++ b/app/views/admin/participants/_judge_scores.html.erb
@@ -13,7 +13,7 @@
       <tr>
         <td> <%= score.team_name %> </td>
         <td>
-          <%= link_to "#{TeamSubmission.find(score.team_submission_id).name}",
+          <%= link_to score.app_name,
                       admin_team_submission_path(score.team_submission_id),
                       class: "cta-link" %>
         </td>
--- a/app/views/admin/participants/_judge_scores.html.erb
+++ b/app/views/admin/participants/_judge_scores.html.erb
@@ -9,7 +9,7 @@
     </tr>
   </th>
   <tbody>
-    <% SubmissionScore.where(judge_profile_id: profile.id).each do |score|  %>
+    <% scores.each do |score| %>
       <tr>
         <td> <%= score.team_name %> </td>
         <td>
--- a/app/views/admin/participants/_judge_scores.html.erb
+++ b/app/views/admin/participants/_judge_scores.html.erb
@@ -13,7 +13,7 @@
       <tr>
         <td> <%= score.team_name %> </td>
         <td>
-          <%= link_to score.app_name,
+          <%= link_to "#{TeamSubmission.find(score.team_submission_id).name}",
                       admin_team_submission_path(score.team_submission_id),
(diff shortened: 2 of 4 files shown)


app/views/admin/participants/_judge_scores.html.erb
```diff
@@ -0,0 +1,29 @@
...
+  <th>
+    <tr>
+      <td>Team Name </td>
+      <td> Submission Name </td>
+      <td> Score </td>
+      <td> Details </td>
+    </tr>
+  </th>
+  <tbody>
+    <% SubmissionScore.where(judge_profile_id: profile.id).each do |score|  %>
```
joemsak: You should make `@` instance variables in the action of the controller where this partial is included. I see you render this partial in `app/views/admin/participants/show` and that is our controller/action path as well:

`app/controllers/admin/participants#show`

See if you can do all the model calls that are in this partial, in the controller#action instead.

And then, see if you can pass these in as locals to the partial

so If you have `@scores` in your controller action, you could assign and use `scores` here

app/views/admin/participants/_judge_scores.html.erb
```diff
@@ -0,0 +1,29 @@
...
+      <tr>
+        <td> <%= score.team_name %> </td>
+        <td>
+          <%= link_to "#{TeamSubmission.find(score.team_submission_id).name}",
```
joemsak: same here about instance vars in controller, you will set and define `@team` in the controller, but the local variable in this partial should be `team`

app/views/admin/participants/show.html.erb
```diff
@@ -82,6 +82,16 @@
         <% end %>
       </div>
     </div>
+
+    <% if @account.judge_profile %>
+      <div class="panel">
+        <div class="grid__col-12 grid__col--bleed-x grid__col--bleed-y">
+            <%= render 'admin/participants/judge_scores',
+              profile: @account.judge_profile %>
```
joemsak: here you would pass scores and team

```
<%= render 'admin/participants/judge_scores',
  profile: @account.judge_profile,
  scores: @scores,
  team: @team %>
```

Allicolyer:
@joemsak let me know what you think now

app/controllers/admin/participants_controller.rb
```diff
@@ -107,5 +108,14 @@ def parental_consent_pending?
         false
       end
     end
+
+    def submission_score(account)
+      if account.judge_profile.present?
+        SubmissionScore.where(judge_profile_id: account)
+      else
+        false
```
joemsak: Instead of `false` here it might be nicer to return `SubmissionScore.none` so that `@scores` is always an instance of `ActiveRelation`

app/controllers/admin/participants_controller.rb
```diff
@@ -107,5 +108,14 @@ def parental_consent_pending?
         false
       end
     end
+
+    def submission_score(account)
+      if account.judge_profile.present?
+        SubmissionScore.where(judge_profile_id: account)
```
joemsak: this could be a few different things:

`SubmissionScore.where(judge_profile: account.judge_profile)`

`account.judge_profile.submission_scores`

I would recommend at least sticking to the current season:

`account.judge_profile.submission_scores.current`

app/views/admin/participants/_judge_scores.html.erb
```diff
@@ -0,0 +1,29 @@
+<h3 class="reset">Scores given by this judge:</h3>
+<table>
+  <th>
+    <tr>
+      <td>Team Name </td>
+      <td> Submission Name </td>
+      <td> Score </td>
+      <td> Details </td>
```
joemsak: Watch your inconsistent spacing inside the tags but also

Add the judging round so you know if it's a QF or SF score

Either add another column or do two tables with the scores scoped down `@scores.quarterfinals` and `@scores.semifinals` (this would happen in the main view template, where you would then call this partial twice)

Allicolyer:
@joemsak let me know what you think now

app/views/admin/participants/show.html.erb
```diff
@@ -82,6 +82,31 @@
         <% end %>
       </div>
     </div>
+
+    <% if @account.judge_profile %>
+      <div class="panel">
+
+        <% if @scores.quarterfinals.present? %>
```
joemsak: you dont have to check for presence for either of these, that was the beauty of ensuring an AR instance is returned  in the controller

app/views/admin/participants/show.html.erb
```diff
@@ -82,6 +82,31 @@
...
+          <h3 class="reset">Quarterfinals Scores:</h3>
+          <div class="grid__col-12 grid__col--bleed-x grid__col--bleed-y">
+            <%= render 'admin/participants/judge_scores',
+              profile: @account.judge_profile,
+              scores: @scores.quarterfinals %>
+          </div>
+        <% end %>
+
+        <% if @scores.semifinals.present? %>
```
joemsak: same comment as above

The table will just render as empty, because `scores.each` will be an empty array (or rather an empty active relation instance) so you could then fix inside the partial to check for `scores.any?` and return a blank state like putting "No scores" instead of a blank table

PR #1534, opened 2018-06-14
Add tests around profanity and sentiment analysis using Algorithmia mock

adrichey:
This pull request adds the following:
- Mock Algorithmia for testing so that we can test the response callbacks
- Alleviate nested `nextTick()` calls with the use of `setImmediate()`. This is a NodeJS function which essentially waits for the queued tasks to finish and then calls the callback. In our case, an expect statement after all the Vue tasks are done.
- I pulled out the `shouldRunProfanityAnalysis` and `shouldRunSentimentAnalysis` into computed properties and wrote unit tests around both functions. This allows for easier mocking in other tests.
- Using the above functionality, I mocked and tested the `setComment` commits that come from the Algorithmia callbacks.

### DIFF
--- a/app/javascript/judge/scores/QuestionSection.vue
+++ b/app/javascript/judge/scores/QuestionSection.vue
@@ -221,6 +221,22 @@ export default {
     prevBtnTxt () {
       return _.capitalize(this.prevSection)
     },
+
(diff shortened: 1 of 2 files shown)


adrichey:
Welcome! Thanks for the merge.

PR #1540, opened 2018-06-18
Create certificate button Vue component and tests

adrichey:
This may need tweaks once merged and it is wired into the UI and back-end.

### DIFF
--- a/spec/javascript/components/CertificateButton.spec.js
+++ b/spec/javascript/components/CertificateButton.spec.js
@@ -18,6 +18,7 @@ describe('CertificateButton Vue component', () => {
   axios.get.mockImplementation(() => {
     return Promise.resolve({
       status: 'complete',
+      payload: {},
     })
   })
 
@@ -225,6 +226,7 @@ describe('CertificateButton Vue component', () => {
       it('polls the job status endpoint if the response status is "queued"', () => {
         const response = {
           status: 'queued',
+          payload: {},
         }
 
         const wrapper = shallow(CertificateButton, {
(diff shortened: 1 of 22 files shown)


spec/javascript/components/CertificateButton.spec.js
```diff
@@ -0,0 +1,294 @@
...
+      })
+
+      wrapper.vm.state = 'ready'
+
+      wrapper.vm.$nextTick(() => {
+        const linkButton = wrapper
+          .find('a.button')
+          .element
+
+        expect(linkButton.getAttribute('href')).toEqual('/mentor/certificates/')
```
joemsak: This href should be set by the result of the 3rd and final JSON GET response where the single property in the response object will be `fileUrl`

spec/javascript/components/CertificateButton.spec.js
```diff
@@ -0,0 +1,294 @@
...
+          },
+        })
+
+        axios.post.mockClear()
+
+        expect(axios.post).not.toHaveBeenCalled()
+
+        wrapper.vm.createJob()
+
+        expect(axios.post).toHaveBeenCalledWith('/mentor/certificates/')
```
joemsak: the passed in prop `teamId` must be appended to this url, if it's not 0 (and for mentors, it won't be 0)

spec/javascript/components/CertificateButton.spec.js
```diff
@@ -0,0 +1,294 @@
...
+          propsData: {
+            userScope: 'mentor',
+          },
+        })
+
+        expect(wrapper.vm.state).toEqual('generating')
+
+        wrapper.vm.handleGenerationRequest(response)
+
+        expect(wrapper.vm.state).toEqual('ready')
```
joemsak: it should also do the final GET request (the 3rd request and 2nd GET request) here, in order to get the certificate `fileUrl`

joemsak: HOLD on this until we pair

PR #1541, opened 2018-06-19
Axios mock refactor

adrichey:
Pulls the `axios` mock out of CertificateButton and puts it in `__mocks__` so it can be reused in other portions of the test suite. Renamed `axios` module to `axiosMock` in the spec so that we know it is being mocked and the implementation is replaced.

### DIFF
--- a/spec/javascript/components/CertificateButton.spec.js
+++ b/spec/javascript/components/CertificateButton.spec.js
@@ -376,21 +376,4 @@ describe('CertificateButton Vue component', () => {
       })
     })
   })
-
-  // Refactor this test once the logic for populating fileUrl is done
-  xit('contains the proper state once all AJAX requests and job have finished', (done) => {
-    const wrapper = shallow(CertificateButton, {
-      propsData: {
-        teamId: 3,
-        userScope: 'mentor',
-      },
-    })
-
-    setImmediate(() => {
-      expect(wrapper.vm.certificateUrl).toEqual('/mentor/certificates/3')
-      expect(wrapper.vm.jobId).toEqual(5)
-      expect(wrapper.vm.state).toEqual('ready')
-      done()
-    })
-  })
 })
--- a/spec/javascript/components/CertificateButton.spec.js
+++ b/spec/javascript/components/CertificateButton.spec.js
@@ -7,7 +7,7 @@ import CertificateButton from 'components/CertificateButton'
 describe('CertificateButton Vue component', () => {
   axiosMock.post.mockImplementation(() => {
     return Promise.resolve({
-      data: { jobId: 5, test: 'blah' },
+      data: { jobId: 5 },
     })
   })
   axiosMock.get.mockImplementation(() => {
--- a/spec/javascript/components/CertificateButton.spec.js
+++ b/spec/javascript/components/CertificateButton.spec.js
@@ -1,21 +1,16 @@
 import { shallow } from '@vue/test-utils'
 
-import axios from 'axios'
+import axiosMock from 'axios'
 import Icon from 'components/Icon'
 import CertificateButton from 'components/CertificateButton'
 
 describe('CertificateButton Vue component', () => {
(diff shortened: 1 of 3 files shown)


spec/javascript/components/CertificateButton.spec.js
```diff
@@ -1,21 +1,16 @@
 import { shallow } from '@vue/test-utils'
 
-import axios from 'axios'
+import axiosMock from 'axios'
```
joemsak: Is this still importing from the package, or from our __ mocks __/axios module?

If it's from our own file, can we make that clearer somehow?

spec/javascript/components/CertificateButton.spec.js
```diff
@@ -1,21 +1,16 @@
...
-  // We will want to move this into a proper mock later on
-  axios.get = jest.fn()
-  axios.post = jest.fn()
-
-  axios.post.mockImplementation(() => {
+  axiosMock.post.mockImplementation(() => {
     return Promise.resolve({
       data: { jobId: 5 },
     })
   })
```
joemsak: Do you think it might help reduce boiler plate if we could do something more like...

```js
mockAxios({ ...returnData }, options = {})
```
First argument is what is returned by promise resolve
Second arg is various options like which method to mock, and whether to mock once or not, and possibly to change to a promise reject

```js
{
  method: "post", // default "get"
  once: true,     // default false
  reject: true,   // default false
  // OR
  resolve: false, // default true
}
```

spec/javascript/components/CertificateButton.spec.js
```diff
@@ -25,8 +20,8 @@ describe('CertificateButton Vue component', () => {
...
 
   beforeEach(() => {
-    axios.post.mockClear()
-    axios.get.mockClear()
+    axiosMock.post.mockClear()
+    axiosMock.get.mockClear()
```
joemsak: My previous comment would need a little more thought for this, or else we simply provide another function to export from our __ mocks __ module

spec/javascript/components/CertificateButton.spec.js
```diff
@@ -271,15 +264,15 @@ describe('CertificateButton Vue component', () => {
           },
         })
 
-        axios.get.mockClear()
+        axiosMock.get.mockClear()
 
-        expect(axios.get).not.toHaveBeenCalled()
+        expect(axiosMock.get).not.toHaveBeenCalled()
```
joemsak: Okay yeah so if we had a "less boilerplate" function for mocking the implementation, I suppose we'd need separate exports, one for the implementation mocking and the other for normal expectation code

spec/javascript/components/CertificateButton.spec.js
joemsak: OR, the method is first arg, the return data is second arg, and options (removing the 'method' option) are last

joemsak:
```js
mockAxios('post', { ... }, opts = { })
```

spec/javascript/components/CertificateButton.spec.js
adrichey: I don't know how to make this more clear, honestly. The name has to be an exact match to pull from the `__mocks__` folder. That is a Jest hoisting and autoloading convention. Perhaps we can make a comment in the code in addition to the name change?

spec/javascript/components/CertificateButton.spec.js
adrichey: I am on the fence about this. I do agree that it is less boilerplate and might be easier to do basic spies/assertions, but it complicates things as you have to learn a new module for testing, and it requires more cognitive load when testing. It also complicates things in that I can't do things like chaining mock implementations. Finally, my reasoning is that we would need to test this new module and if it needed to change, then we have to go back and change all of our tests, rather than the spec using the simple 1:1 mock.

PR #1544, opened 2018-06-21
Screenshots refactor and initial tests

adrichey:
This is a pretty heavy copy and paste refactor of the screenshot uploading functionality to make it a little more testable. Before, the template was in the `erb` partial for the UI and the component code was in the screenshots.js file. This refactor consolidates the template and Vue instance code into a single component which allows us for simpler, more focused testing. Here is a rundown of all the changed that were implemented:

- Moved screenshots.en.html.erb template into ScreenshotUploader.vue
- Moved the Vue instance code in screenshots.js into ScreenshotUploader.vue
- Moved data attributes that populate `data-sort-url`, `data-url`, and `data-team-id` to the properties on ScreenshotUploader
- Moved the Vue.vueDragula event bus to the global scope so that it can be used in the component. This may need to be looked at for improvements as it is only used within the component, but requires setup outside of it to get the directive to work as far as I can see. I made a note to come back to this.
- Created tests to begin testing the functionality.

**Important Note:** I did not write the tests before the refactor. I took two full hours last night after work to manually test against QA. This isn't ideal, but I felt that testing with the template and instance separate comes with its own issues. We can discuss any of this if you find it problematic. I figured I would get this in front of a set of eyes quickly to see if this workflow is appropriate for these cases where components are spread out.

Thanks! :)

### DIFF
--- a/app/javascript/components/ScreenshotUploader.vue
+++ b/app/javascript/components/ScreenshotUploader.vue
@@ -1,0 +1,229 @@
+<template>
+  <div id="screenshot-uploader">
+    <template v-if="maxFiles > 0">
+      <label class="label--wraps-file-input">
+        <input type="file" multiple @change="handleFileInput" />
+
+        <span class="button button--small">
+          + Add {{ prefix }} {{ maxFiles }} {{ object }}
+        </span>
+      </label>
+    </template>
+
+    <template v-else>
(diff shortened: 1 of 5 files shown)


adrichey:
I also did some minor cleanup on the attendee spec in order to get it using vue-test-utils for consistency.

PR #1566, opened 2018-06-27
Pie Charts Vue Refactor: Part 1

adrichey:
This converts the charts.js functionality over to using the new PieChart Vue component for testability. I believe we can remove the `content_tag :canvas` from the system as well, but I am not aware of how to do that on the Ruby/Rails side of things, so I didn't do that during this go around.

The next step after this is merged is to create a more flexible admin interface that hooks into the Vue Router so that we can create endpoints that can dynamically populate the various chart interfaces we have.


app/javascript/components/PieChart.vue
```diff
@@ -0,0 +1,137 @@
+<template>
+  <div class="certificate-button">
```
joemsak: probably should rename this class

app/javascript/admin/index.js
```diff
@@ -0,0 +1,16 @@
+import Vue from 'vue/dist/vue.esm'
+import PieChart from '../components/PieChart'
+
+document.addEventListener('turbolinks:load', () => {
+  const btnEls = document.querySelectorAll('.vue-enable-pie-chart')
```
joemsak: probably should rename this variable

app/javascript/packs/admin.js
```diff
@@ -12,5 +12,6 @@ Vue.use(TurbolinksAdapter);
 Vue.use(VTooltip)
 Vue.use(Vue2Filters)
 
+import '../admin'
```
joemsak: maybe this should be admin/dashboard? or some other directory/name for a shared js module because RAs will import it, too?

app/javascript/packs/ra.js
```diff
@@ -12,5 +12,6 @@ Vue.use(TurbolinksAdapter);
 Vue.use(VTooltip)
 Vue.use(Vue2Filters)
 
+import '../ra'
```
joemsak: see note above for sharing the module, seems we should import the same file as they should do the same thing and be agnostic to the user scope

app/javascript/ra/index.js
```diff
@@ -0,0 +1,16 @@
...
+  btnEls.forEach((btnEl) => {
+    new Vue({
+      el: btnEl,
+
+      components: {
+        PieChart,
+      },
+    })
+  })
+})
```
joemsak: hopefully this file isn't necessary if admins and RAs can share an imported module file, as the data will come from the endpoint, and the scope can be passed in on the HTML template

app/views/admin/dashboards/show.html.erb
```diff
@@ -12,77 +12,73 @@
...
+          :chart-classes="{
+            'quickview-charts': true,
+            'quickview-charts--students-permitted': true,
+          }"
+          :chart-data="{
             labels: [
-              "With parental permission – #{show_percentage(@permitted_students, @students)}",
-              "Without parental permission – #{show_percentage(@unpermitted_students, @students)}",
+              'With parental permission – <%= show_percentage(@permitted_students, @students) %>',
+              'Without parental permission – <%= show_percentage(@unpermitted_students, @students) %>',
```
joemsak: I'm assuming this is the sort of stuff you meant for the next steps, to get this data from the endpoint that initializes the chart

app/views/admin/dashboards/show.html.erb
```diff
@@ -12,77 +12,73 @@
...
-                  ),
-                  admin_participants_path(
+                  ) %>',
+                  '<%= admin_participants_path(
                     accounts_grid: {
                       scope_names: ["student"],
                       parental_consent: "not_parental_consented",
                     }
-                  ),
+                  ) %>',
```
joemsak: same comment as above, I'm assuming the next step is to get data and URLs from the JSON endpoint we need to wire up

spec/javascript/components/PieChart.spec.js
```diff
@@ -0,0 +1,182 @@
...
+        },
+
+        colorRange: {
+          type: Object,
+          default: expect.any(Function)
+        },
+      })
+    })
+
+    it('chartClasses sets the quickview-charts class by default', () => {
```
joemsak: if the sentence can't read right with "it" should we not put that part in a surrounding describe?

```js
describe("chartClasses", () => { it("sets...", () => { }) })
```

spec/javascript/components/PieChart.spec.js
```diff
@@ -0,0 +1,182 @@
...
+      expect(PieChart.props.chartClasses.default())
+        .toEqual({
+          'quickview-charts': true,
+        })
+    })
+
+    it('colorRange sets a light blue to light pink range by default', () => {
```
joemsak: same comment as above with making the sentence read nice with 'it'

app/javascript/components/PieChart.vue
adrichey: This is why peer review is necessary. I am rolling my eyes at myself right now. This is what I get for copying and pasting and then purging content. I'll get that changed.

app/javascript/packs/ra.js
adrichey: I dig it. I will update that.

app/views/admin/dashboards/show.html.erb
adrichey: Correct. First, implement the current charts using a component. The next step will be creating a wrapper Vue that will make the calls to populate the components/props more dynamically.

spec/javascript/components/PieChart.spec.js
adrichey: Oooooh, I like that. That makes sense. It definitely reads better.

app/views/admin/dashboards/show.html.erb
joemsak: Alright, I was imagining the component itself would grab the data, and that the surrounding app is more for layout and composition purposes

joemsak: so the component accepts enough props to make the AJAX call and then expects a response that lets it populate the chart options (labels, data, and URLs)

spec/javascript/components/PieChart.spec.js
joemsak: well you did it that way in your other specs  😂

app/views/admin/dashboards/show.html.erb
adrichey: Okay. I will ping you on Slack to discuss this in a moment. I am fine with that. I just want to make sure I am visualizing how all of this fits in with the router and/or vuex, as it sounds like that might not factor in at that point.

adrichey:
I have addressed your issues with the pull request. Feel free to review again when you have time @joemsak.

PR #1575, opened 2018-07-03
Pie Charts Vue Refactor: Part 3

adrichey:
This is a pretty major overhaul on how we initialize charts for the Admin section dashboard. Please note that once this is merged, I will begin work on the Regional Ambassadors section. I wanted to get this in so that we could see what the section changes look like before I got in too far.

- Adds a tabbed interface to Admin dashboard for our charts using Vue Router
- Adds Vuex for state management: caching AJAX calls to get chart data
- Adds full testing suite for the changes, including routes and Vuex


app/views/admin/dashboards/show.html.erb
```diff
@@ -2,64 +2,17 @@
...
-        />
-      </div>
-    </div>
-  </div>
-</div>
+<div id="admin-dashboard">
+  <admin-dashboard
+    :totals="{
+      students: <%= number_with_delimiter @students.count %>,
+      mentors: <%= number_with_delimiter @mentors.count %>,
```
joemsak: we'll want the app to start getting these things from their endpoints, to avoid the double counting in the backend

adrichey: Noted. I will make a note of that. We should be able to calculate this on the fly with the data we get back. I can just add all the data together to get that number.

adrichey:
I have been investigating the issue where the Mentors tab shows the charts loading and then disappears in Chrome and have been able to replicate it locally. The issue isn't with the charts, but with the data. If the URL endpoint and/or state has an empty data set of zeroes in its `data` array, then the chart simply does not appear. This makes sense because there is no data to be displayed. I think we may want to display some kind of "No data found" message in these cases instead of loading the chart. What are your thoughts @joemsak?

adrichey:
Added a "No data to display" message. Please feel free to review again @joemsak.

joemsak:
tada emoji

PR #1578, opened 2018-07-05
Pie Charts Vue Refactor: RA Section

adrichey:
I will begin work on the bar chart component that will be used to display the participants data while this goes through the PR process. Here's a rundown of my changes within this section.

- Add props to DashboardSection to allow for flexible initialization of components throughout app to promote reuse
- Fixes a bug in data_analysis endpoint for mentors for the data labels
- Converts the Recruitment section in the RA dashboard over to Student and Mentors sections to match that of the Site-wide Admin dashboard


app/views/regional_ambassador/dashboards/_mentors.en.html.erb
```diff
@@ -0,0 +1,33 @@
...
+
+<hr />
+
+<div id="ra-admin-mentors-section">
+  <mentors-section
+    <% if current_ambassador.country == "US" %>
+      :international="false"
+    <% else %>
+      :international="true"
+    <% end %>
```
joemsak: alternatively, I believe this should work

```erb
:international="<%= current_ambassador.country != "US" %>"
```

app/views/regional_ambassador/dashboards/_mentors.en.html.erb
```diff
@@ -0,0 +1,33 @@
...
+    :totals="{
+      mentors: '<%= number_with_delimiter @mentors.count %>',
```
joemsak: really seems like this data could (and should) come from the initial data load, so that we don't do the SQL count both on the HTML page load and during the subsequent JSON data loads

app/views/regional_ambassador/dashboards/_mentors.en.html.erb
```diff
@@ -0,0 +1,33 @@
...
+    }"
+    :chart-endpoints="{
+      <% if current_ambassador.country == "US" %>
+        onboarding_mentors: '<%= regional_ambassador_data_analysis_path(:onboarding_mentors) %>',
+      <% else %>
+        onboarding_mentors: '<%= regional_ambassador_data_analysis_path(:onboarding_international_mentors) %>',
+      <% end %>
```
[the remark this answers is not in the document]
adrichey: I have consolidated the two endpoint urls into a single line. I will need your assistance in setting up the endpoint logic on the back-end before we merge this in @joemsak.

adrichey:
I have addressed all of the issues you had with the pull request. The totals are now pulled from the AJAX endpoints that populate the charts based on the data analysis serializer JSON responses. Additionally, I ended up moving around the totals storage from the Vuex store to the components that extend DashboardSection in order to alleviate a weird issue where the store wasn't pulling store states from getters properly. The only things that remains is to hook up the logic in the `regional_ambassador_data_analysis_path(:onboarding_mentors)` function call to return the correct data based on whether or not we are looking as US or international mentor data, which I will need your help with @joemsak.

joemsak:
emoji: Ta Da

PR #1581, opened 2018-07-10
Top Countries Bar Chart Implementation

adrichey:
- Adds a new bar chart component which builds out stacked bar charts.
- Adds a new section to the admin dashboard which features data on top countries.
  - Contains top ten countries with the highest aggregated student, mentor, and judge counts.
  - Can be sorted to only display students, mentors, or judges data.
  - Can click through to the filters by clicking on the bars in the chart.
- Creates a data analysis endpoint to populate the top countries bar chart.
- Creates a new isEmptyObject utility function for code reuse.
- Moves click handling on the PieChart component over to the chart.js plugins.
- Improves the cursor hover states on all charts.


app/javascript/admin/dashboard/components/AdminDashboard.vue
```diff
@@ -22,6 +22,17 @@
           Mentors
         </button>
       </router-link>
+
+      <router-link
+        tag="li"
+        class="tab-link"
```
joemsak: deprecated css class, consider using `tabs__menu-link`

app/javascript/admin/dashboard/components/TopCountriesSection.vue
```diff
@@ -0,0 +1,36 @@
+<template>
+  <div id="top-countries">
+    <h3>Top Countries<span v-if="getTotal('top_countries')"> ({{ getTotal('top_countries') }})</span></h3>
```
joemsak: consider breaking this line up for length

app/javascript/utilities/chartjs-plugins.js
```diff
@@ -22,4 +23,23 @@ Chart.plugins.register({
...
+
+      const { type } = chart.config
+      const { urls } = chart.data.datasets[datasetIndex]
+
+      if (typeof urls !== 'undefined' && urls.length > 0) {
+        window.location.href = urls[index]
+      }
+    }
+  },
+})
```
joemsak: 👍 very cool

app/models/data_analysis.rb
```diff
@@ -319,4 +327,117 @@ def data
...
+      .where("country IS NOT NULL")
+      .group(:country)
+      .limit(10)
+      .order('count_all desc')
+      .count
+
+    @top_students = Account.current
+      .left_outer_joins(:student_profile)
+      .where("student_profiles.id IS NOT NULL")
+      .where(:country => @top_countries.keys)
```
joemsak: nitpicky but modern ruby hash syntax is more like JSON

`country: @top_countryes.keys`

app/models/data_analysis.rb
```diff
@@ -319,4 +327,117 @@ def data
...
+      .where("judge_profiles.id IS NOT NULL")
+      .where(:country => @top_countries.keys)
+      .group(:country)
+      .order('country desc')
+      .count
+  end
+
+  def totals
+    {
+      top_countries: number_with_delimiter(@top_students.values.sum + @top_mentors.values.sum + @top_judges.values.sum),
```
joemsak: let's break this line due to length, starting at the opening paren, the `+` symbols can be left hanging to the next line

app/models/data_analysis.rb
```diff
@@ -319,4 +327,117 @@ def data
...
+    }
+  end
+
+  def labels
+    @top_students.keys.map do |country_code|
+      FriendlyCountry.(
+        OpenStruct.new(country: country_code),
+        {
+          prefix: false,
+        }
```
joemsak: you can omit the `{` & `}` when the last or only arg is a hash, ruby can handle it :)

app/models/data_analysis.rb
```diff
@@ -319,4 +327,117 @@ def data
...
+    @top_judges.keys.each do |country_code|
+      judges.push(
+        url_helper.public_send("#{user.scope_name}_participants_path",
+          accounts_grid: {
+            scope_names: ["judge"],
+            country: [country_code],
+          }
+        )
+      )
+    end
```
joemsak: instead of initializing an empty array variable, you can set that variable to the result of a `map` on the keys (instead of doing an `each`)

app/javascript/admin/dashboard/components/AdminDashboard.vue
adrichey: Ah, right. Good catch. I will get that updated shortly.

app/javascript/admin/dashboard/components/TopCountriesSection.vue
adrichey: I can do that. There are layout tests around this, so I will tweak them to be more flexible at the same time. They are what forced this to be on one line.

app/models/data_analysis.rb
adrichey: Cool! This is good to know. I had been looking at older docs. I will update that.

adrichey:
@joemsak Feel free to take another pass over this. I believe I have addressed all of the issues you raised. I think I was able to fix all of the deprecated classes.

app/models/data_analysis.rb
```diff
@@ -319,4 +327,103 @@ def data
...
+        )
+      end,
+      @top_judges.keys.map do |country_code|
+        url_helper.public_send("#{user.scope_name}_participants_path",
+          accounts_grid: {
+            scope_names: ["judge"],
+            country: [country_code],
+          }
+        )
+      end,
```
joemsak: oh if you're going to inline the maps (which is fine) I would opt for the `{...}` syntax over `do...end` because the commas are a bit awkward with the `end` keyword

and, nitpicky again, but ** in general ** break all the args if you break one

```ruby
@top_judges.keys.map { |country_code|
  url_helper.public_send(
    "#{user.scope_name}_participants_path",
    accounts_grid: {
      scope_names: ["judge"],
      country: [country_code],
    }
  )
},
```

joemsak:
tada emoji

PR #1582, opened 2018-07-10
Admin Dashboard: hideTotals property

adrichey:
Implements a hideTotals property on the DashboardSection Vue component so that we can hide the totals in our admin dashboard sections where it makes sense, regardless of whether or not the total is present.

Also removes the total count from the Top Countries section in the admin dashboard.


joemsak:
🎉

PR #1584, opened 2018-07-11
Datagrid Table Layout Improvements

adrichey:
- Datagrid tables that become larger horizontally than the viewing area get scrollbars on the top and bottom so that they don't bleed over.
- The Actions column in the datagrids are now sticky and float to the right when the scrollbars appear.
- Header columns now display without being wrapped to keep them smaller and more readable.

### DIFF
--- a/app/assets/stylesheets/_datagrid.scss
+++ b/app/assets/stylesheets/_datagrid.scss
@@ -1,5 +1,15 @@
 .datagrid {
   background: white;
+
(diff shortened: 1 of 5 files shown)


joemsak:
can you be sure to use browserstack.com and check IE 11, MS Edge, safari, and firefox?

joemsak:
oh I guess we'll have to have it on QA first, sorry

PR #1598, opened 2018-07-12
IE11 Fixes

adrichey:
Please see commits for details. There may be subsequent fixes that go in, but I want to see how many errors the two of these clear up.

- Global `jQuery` was not present due to scoping and initialization, so it has been converted to `$` which was available in doubleScroll.
- Converted ES6 fat arrow function over to standard function as it was throwing errors in IE11. https://caniuse.com/#feat=arrow-functions
- Fixed table styles for FF and Edge browsers for floating Actions column in datagrids.

We may want to go through our non-webpacker compiled JS and convert anything over from ES6 to something more backwards compatible friendly.


joemsak:
🥇

PR #1599, opened 2018-07-12
IE11 Fixes Round Two

adrichey:
- Convert ES6 functionality over to ES5 for IE11 compatibility in static JS asset files
- Make production webpacker JS uglifier use ES5 instead of ES8 for IE11 compatibility


joemsak:
> Make production webpacker JS uglifier use ES5 instead of ES8 for IE11 compatibility

Damn, didn't know about this or even to look something up like this! Nice dude

PR #1600, opened 2018-07-12
Location Form Improvements

joemsak:
Adds
`app/javascripts/location/components/LocationForm.vue`

To replace the HTML in `app/views/signups/_location_fields.en.html.erb`

`GET /student/current_location`
`PATCH /student/location`
`GET /mentor/current_location`
`PATCH /mentor/location`
`GET /judge/current_location`
`PATCH /judge/location`
`GET /admin/current_location`
`PATCH /admin/location`

if the Admin endpoints detect an `account_id` parameter, they will engage with that account, otherwise they will work for the currently signed in Admin (though we don't need location pages for admins, so that probably won't happen)

subsequently, the same is true for the Admin endpoint regarding a `team_id` parameter

if the Student endpoints detect a `team_id` parameter, they will use the currently logged in student's team as the location object, regardless of the actual ID passed with the parameter

if the Mentor endpoints detect a `team_id` parameter, they will use the currently logged in mentor's list of current teams to find the correct team by the passed in ID, meaning incorrect team IDs cannot be used and will raise an exception

the judge endpoints do not support or detect a `team_id` parameter, as judges are not on teams

Also adds
`app/technovation/handle_geocoder_search.rb`

```ruby
json_data_as_hash, http_status = HandleGeocoderSearch.(account_or_team, controller_params)
```


app/javascript/constants/HttpStatusCodes.js
```diff
@@ -0,0 +1,4 @@
+export default {
+  MULTIPLE_CHOICES: 300,
+  NOT_FOUND:        404,
+}
```
adrichey: I am a fan of this logic. Not having raw magic numbers for the win!

app/javascript/packs/application.js
```diff
@@ -9,21 +9,32 @@
...
-    for (let i = 0; i < btnEls.length; i += 1) {
-      new Vue({
-        el: btnEls[i],
-
-        components: {
-          CertificateButton,
-        },
-      })
-    }
-  }
```
adrichey: Did you test this change in IE11? That `forEach` was deliberately changed to a native `for` for that reason. `forEach` is not implemented on NodeLists in IE11.

app/javascript/constants/HttpStatusCodes.js
joemsak: Thanks, I figure we can add more as needed

app/javascript/packs/application.js
joemsak: no this is a bad merge conflict resolution, sorry

adrichey: Every time I look at the `btnEls` variable I think that it could use a better name. Not something you need to change, but something I might go back and clean up later on.

app/javascript/location/components/LocationForm.vue
```diff
@@ -0,0 +1,404 @@
...
+      this.status = err.response.status
+
+      if (this.status === HttpStatusCodes.MULTIPLE_CHOICES) {
+        this.suggestions = err.response.data.results.map(result => {
+          return new LocationResult(result)
+        })
+      }
+    },
+
+    handleCountryChange (value) {
```
adrichey: `handleCountryChange` doesn't appear to be used anywhere. I think we can get rid of it.

joemsak: oh yeah i was using vue-select (kind of a chosen clone) at first but the UX just sucked so i said forget it, let the user type whatever, the geocoder is actually quite good

app/javascript/location/components/LocationForm.vue
```diff
@@ -0,0 +1,404 @@
...
+      params[rootParamName] = {
+        city: this.city,
+        state_code: this.stateCode,
+        country_code: this.countryCode,
+      }
+
+      return params
+    },
+
+    countryOptions () {
```
adrichey: I am not seeing `countryOptions` used anywhere. Looks like it may not be needed.

app/javascript/location/components/LocationForm.vue
adrichey: Having used vue-select heavily on my last project, I can honestly say that it has many quirks. So I completely understand haha.

adrichey:
Looks great! Thanks for getting those suggestions addressed so quickly! I will merge it in now.

PR #1601, opened 2018-07-12
IE11 Fixes Round Three

adrichey:
- Fix a meta tag to not be a content_tag for IE11 compatibility so we don't get a bunch of console warnings.
- Convert `forEach` on an object to native `for` loop so that we don't throw IE11 errors.
- Move jQuery doubleScroll widget to the assets pipeline vendors directory and add it to the admin.js entry point since it is a static non-compiled asset.


app/views/application/_meta.en.html.erb
```diff
@@ -27,11 +27,10 @@
...
 
-<%= content_tag :meta, nil,
-  data: {
-    current_profile_id: current_profile.id,
-    current_profile_type: current_profile_type,
-  } %>
+<meta
+  data-current-profile-id="<%= current_profile.id %>"
+  data-current-profile-type="<%= current_profile_type %>"
+/>
```
joemsak: weird that this was a problem in IE11 ... can you elaborate? thanks!

adrichey: It wasn't throwing errors, it was just adding a few lines to the console as warning. In IE11, meta tags that are considered content (`<meta></meta>`) throw errors as meta tags aren't supposed to have matching end tags. IE is quirky haha. So converting it to a self closing tag removed the warnings. I was mainly converting it for readability as the console was pretty packed at times yesterday. :)

PR #1602, opened 2018-07-12
Convert forEach loop to native for to support IE11

adrichey:
IE11 doesn't like using `forEach` on NodeLists.

### DIFF
--- a/app/javascript/admin/dashboard/index.js
+++ b/app/javascript/admin/dashboard/index.js
@@ -14,15 +14,17 @@ import { router } from './routes'
 document.addEventListener('turbolinks:load', () => {
   const pieChartElements = document.querySelectorAll('.vue-enable-pie-chart')
 
-  pieChartElements.forEach((element) => {
-    new Vue({
-      el: element,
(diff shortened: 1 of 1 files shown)


app/javascript/admin/dashboard/index.js
```diff
@@ -14,15 +14,17 @@ import { router } from './routes'
...
-        PieChart,
-      },
-    })
-  })
+        components: {
+          PieChart,
+        },
+      })
+    }
+  }
```
joemsak: We use forEach on elements in a handful of spots. Is it possible to get a polyfill, or was it something you had already decided against?

adrichey: I actually did a grep for this in our application and only found two spots where this could be problematic. In most cases `.forEach` should be fine, so I would rather not polyfill. The only time to keep this in mind is when working with `NodeLists` which are returned by `querySelectorAll` as they aren't arrays. This was my problem as I didn't think about IE11 support at the time. So we should be okay going forward.

PR #1605, opened 2018-07-13
Remove deprecated tabs classes throughout the application

adrichey:
This converts all non-BEM tabs classes throughout the application over to their non-deprecated BEM versions. I thoroughly tested the UIs to make sure that the javascript edits I made didn't break anything, and to confirm that things were styled correctly. I know this is a large pull request and it grew very large after I started in on this, so I apologize in advance for the pot of coffee you might have to consume to make sure all of this works haha.


joemsak:
good enough for the summer, we can fix anything if something goes awry

Issue #1562, opened 2018-06-25
Create auto-fill for top company sign ups


Allicolyer:
Program team to give us a list of top companys that mentors and judges sign up from.

Allicolyer:
@madhavitechnovation could your provide us with a list of top companies that mentors and judges sign up from?

madhavitechnovation:
Salesforce, Adobe, Gogle, uber, Samsung, oracle, AECOM, HARMAN, Microsoft, Accenture, Bank of America.

I hope we can add more tot he list, if we have new partners, correct?

adrichey:
Work for this has been completed. We have implemented an input that allows you to either select from a list of top companies or add your own company that isn't in the list. @madhavitechnovation editing the list of top companies is very straightforward and won't break anything if we ever need to add or remove one from the list as it is done in one place.

Allicolyer:
@joemsak @adrichey will this auto fill based on IP addresses or is this just a dropdown list for signup?

Allicolyer:
@adrichey I noticed something unexpected about this text box for Company name. If you start typing something, you must also hit enter for it to save any characters you typed. If you click out of the box or click tab it will not save the characters you've typed.

adrichey:
@Allicolyer This will not fill based on IP addresses for this iteration. I would suggest that we save that piece of functionality for an improvement card later on. The component we implemented is called vue-select, and it requires that enter key press or a click on the drop down as it functions similarly to `<select>` element rather than a standard input.

Allicolyer:
okay great!

PR #1609, opened 2018-07-20
Top Companies Signup Autofill

adrichey:
This implements an `AutocompleteInput` Vue component that takes an AJAX `url` String or `options` Array and creates an input that has a typeahead autofill feature. It also creates a matching `AutocompleteInput` SimpleForm input that can be used to create the markup we need to initialize one of these fields on any given application page.

The autocomplete fields have been implemented for the following:
- /judge/signup - Company Name
- /mentor/signup - School or Company Name
Please type "of", "Goo", or "ub" into these fields to see various top company results.

This addresses issue #1562. This is also my second crack at some actual Ruby functionality, so please let me know if anything seems awry in that neck of the woods. I also didn't write any rspec around this code, so let me know if I need to do so and I will work on getting that in there ASAP.


adrichey:
We are failing rspec test: /app/spec/features/admin/invite_users_spec.rb

The reason being is that webpacker JavaScript is not able to initialize the Vue enabled autocomplete fields, and therefore the `fill_in` method isn't able to find an input to fill in. Which of course is problematic, because if remove the following lines, the tests fail:
```
fill_in "School or company name", with: "John Hughes High."
fill_in "Company name", with: "John Hughes High."
```

I am going to need your assistance to rectify this @joemsak as my Rails knowledge is very limited in this regard.

adrichey:
This is ready for review once again.

PR #1613, opened 2018-07-27
Improve autocomplete input with v-model support and custom vue-select for blocking chrome autofill

joemsak:
This allows a v-model from the parent Vue instance, and uses a custom vue-select (forked from 2.4.0)  where I add attributes to the Select component so that chrome will leave it alone for autofill


adrichey:
Checked out the branch and took a look around. Looks good! I am going to merge this in and delete the branch.

PR #1632, opened 2018-08-28
Reduce vendor js

joemsak:
@adrichey would be good if you load this up and make sure various views work for you

sign up
student dashboard
ra events app functionality
admin content & settings

i can't tell if i'm having env issues on my own machine with getting package dependency to work or if i screwed up the branch somehow. i intermittently lose modules being found by the webpack dev server compiler


joemsak:
by the way, using this tool suggests I removed a full MB from the production bundle

https://razorjack.net/visualizing-optimizing-javascript-bundle-size-rails-webpacker/

joemsak:
linked in the article:

https://www.npmjs.com/package/babel-plugin-lodash

PR #1643, opened 2018-08-30
Improve BasicProfile UX while it is embedded by showing the saved state

joemsak:
Got a couple things that might be anti patterns here, wanna run it by you

When the basic profile component is "embedded" there is no "next" button so instead I disabled it and show the "Saving..." and "Saved!" states

The method in which I get the embedded: true prop to BasicProfile concerns me, wanna let you see it first

### DIFF
--- a/app/javascript/registration/components/BasicProfile.vue
+++ b/app/javascript/registration/components/BasicProfile.vue
@@ -118,10 +118,10 @@
       </a>
       <button
         class="button"
-        :disabled="!nextStepEnabled"
+        :disabled="!nextStepEnabled || embedded"
         @click.prevent="handleSubmit"
       >
-        Next
+        {{ nextButtonLabel }}
       </button>
     </div>
   </form>
@@ -167,7 +167,17 @@ export default {
         'Past Technovation student',
       ],
       expertiseOptions: [],
+      saving: false,
+      saved: false,
     }
+  },
+
+  props: {
+    embedded: {
+      type: Boolean,
+      required: false,
+      default: false,
+    },
   },
 
   beforeRouteEnter (_to, from, next) {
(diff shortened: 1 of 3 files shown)


app/javascript/registration/components/BasicProfile.vue
```diff
@@ -215,7 +245,7 @@ export default {
       },
 
       set (value) {
-        this.$store.commit('registration/firstName', value)
+        this.handleDataChanges('registration/firstName', value)
```
joemsak: This happened because I forgot about the debouncedProfileUpdate and originally was gonna set saved/saving here

but then realized I could leave this as a discussion point about our common boilerplate

app/javascript/student/routes/index.js
```diff
@@ -15,6 +15,18 @@ import { routes as registrationRoutes } from 'registration/routes'
...
+  route => route.name === 'basic-profile'
+)
+let basicProfileRoute = modifiedRegistrationRoutes[basicProfileRouteIndex]
+
+basicProfileRoute.props = {
+  embedded: true,
+}
+
+modifiedRegistrationRoutes[basicProfileRouteIndex] = basicProfileRoute
+
```
joemsak: this is the code of my highest concern

joemsak: actually now this concerns me more because I realized I forgot to set the `children` of the registration app route to this new variable and it was working correctly in the browser

app/javascript/registration/components/BasicProfile.vue
adrichey: This doesn't bother me terribly since the saved and saving are only driving the label for the Next button and nothing else.

PR #1649, opened 2018-09-04
Mentor Onboarding 2019

joemsak:
Copying from student, want to identify reusable parts and extract them out, try to avoid doing too much extraction too early but prefer FE optimization


adrichey:
Didn't find anything glaring. I can assist with pulling out similar code and working with Vue extends next week to reduce the code duplication. Merging this in.

Issue #1650, opened 2018-09-04
Add resources to dashboard


Allicolyer:
Add in resources such as judging rubric, curriculum and submission guidelines.

QA -- make sure these links are correct and updated.

joemsak:
@Allicolyer for students we have a proof of concept for a new profile area where a menu of options may one day live

for now, it shows our first step of putting the resources there in the top right corner with the student's name (i hope in wishlist time we can make this a kind of profile menu with an avatar, etc)

thanks!

Allicolyer:
Thanks Joe! I think this looks good. I have a few questions

1. We are missing the curriculum link and the project workbook, is that because this is still a WIP?
2. Do you want to work with Ruby to make the resources less crowded?

joemsak:
@adrichey please see #1

@Allicolyer for #2 we only just released this to QA today so give me some time to work it into the UI more comfortably as we merge the mentor version over the next few days. thanks!

we do have a full working mentor version locally and we just need to eliminate some of our duplicated work to make it open to creating a judge dashboard app for judge related sprints and to reduce the front-end code size

adrichey:
#1 is addressed by the following PR: https://github.com/Iridescent-CM/technovation-app/pull/1671

joemsak:
should be updated in about 10 min

Allicolyer:
I see the missing resources. Thank you!

adrichey:
@Allicolyer Since you are able to see the resources, can we close this ticket?

Allicolyer:
Okay makes sense.

So I just noticed that the submission guidelines are repeated twice for students.

joemsak:
Perfect card example, thanks Alli!

Joe Sak
Technovation Web Development

SMS
+1 (312) 945-6317

Allicolyer:
okay

PR #1668, opened 2018-09-05
Add resources to dashboard

adrichey:
Addresses issue #1650 by adding missing resources links back into mentor and student sections.

They are now located under the right-hand header with the student/mentor notices under a link that is labeled `More Information`. When `More Information` is clicked, a menu is collapsed downward and the resource links are displayed.

This section has also been added to the judge dashboard as well in preparation for the changes that will come with implementing the wizard workflow in that section. If that is not ideal, commit `a399dfb` can be reverted.

This will most likely be moved once the top navigation is improved during a later "wishlist UI" sprint.


Allicolyer:
Thanks Alan!

PR #1671, opened 2018-09-06
Add missing curriculum and project workbook links to drop-down

adrichey:
Addresses @Allicolyer comment:
https://github.com/Iridescent-CM/technovation-app/issues/1650#issuecomment-419161422


joemsak:
fine for now but lets do these links as props then so the necessary HTML isn't a concern

Issue #1721, opened 2018-09-27
Code Checklist Point Counter for Pictures of Process


Allicolyer:
As a student
With a submission
That already has two uploaded screenshots
When I go to edit my code checklist
And I attach the two pictures of my process (before pressing save)
Then I should see the counter award 3 out of 3 points

As a mentor
With a submission
That already has two uploaded screenshots
When I go to edit my code checklist
And I attach the two pictures of my process (before pressing save)
Then I should see the counter award 3 out of 3 points

amyiridescent:
I am not sure if I am following instructions well here. So I am awarded the 3 out of 3 points (from the total of 10 points available under "code check section", when I add two picture to the "pictures of your process" category. I do see that. As long as two screen shoots were already uploaded. 

![uploads/b55bac11-6a89-494d-ae13-e5edb4bbf654/image.png](https://s3-us-west-2.amazonaws.com/prod.huboard.com/uploads%2Fb55bac11-6a89-494d-ae13-e5edb4bbf654%2Fimage.png)

If I upload the screenshoots AFTER I started the "code checklist", the score does not get updated on the dashboard. I need to click back into "code checklist" before it gets updated.  


![uploads/ff1ba4a2-b33b-48bb-a081-d51cacb3d730/image.png](https://s3-us-west-2.amazonaws.com/prod.huboard.com/uploads%2Fff1ba4a2-b33b-48bb-a081-d51cacb3d730%2Fimage.png)

(For both mentor and student)

joemsak:
good looking out @amyiridescent you were right to approach the test from another angle, and we can fix this too.

amyiridescent:
Well I didn't do it on purpose. I didn't read Alli's instructions careful and ended up having to do it from both angles. And that is how I found the difference. hehe :)

joemsak:
@amyiridescent I checked into this and it seems like your issue is that the points do not update dynamically on the left-side menu when you upload screenshots. We didn't design the code to ever do that, updating it dynamically has not been a requirement as of yet, and it does update the 3 points when you load any page that has that menu on it. We'll have to consider this card passing the original acceptance criteria, and I wouldn't recommend prioritizing making the left-side submission menu update dynamically at any time soon

amyiridescent:
Sounds perfectly reasonable to me. We can think about dynamic update (?) features next year. Closing the ticket now.

PR #1745, opened 2018-10-03
Update code checklist javascript to work with screenshots

adrichey:
Addresses #1721.

While this code fixes the issue, I think that we need a system test to load up two screenshots and then populate the two file fields at some point. I feel like doing the test on the JavaScript side doesn't give an actual feel of whether or not this works as it will need to mock the DOM, which is telling that the test is not adequate.

Unfortunately, my skills on the RSpec side are lacking, so I am unable to do that.

### DIFF
--- a/app/assets/javascripts/code-checklists.js
+++ b/app/assets/javascripts/code-checklists.js
@@ -4,18 +4,23 @@ $(document).on("ready turbolinks:load", function() {
   });
 });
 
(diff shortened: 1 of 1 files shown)


joemsak:
If it passes the AC, it's okay with me, it's a small display bug, the points being awarded in the ruby logic is unit tested

PR #1768, opened 2018-10-10
Top Countries Labels Regression Bugfix

adrichey:
Get country friendly country names from Carmen for the top countries bar chart table since FriendlyCountry was modified for the location form to use record objects. OpenStruct no longer works as the record parameter as it used to.


joemsak:
Hmm, maybe we just have to fix the open struct object? I want "friendly country" to be the centralized, standard module in the codebase for getting this info

joemsak:
Looks like maybe we could just pass an object that responds to 'address details' instead - have that return the country code

joemsak:
In other words, pass it ```ruby OpenStruct.new(address_details: country_code)```

PR #1795, opened 2018-11-02
Update bundle, including a security vulnerability

joemsak:


joemsak:
test passed locally, and are very slow, merging off-site

Issue #1804, opened 2018-11-05
RAs cannot select teams & judges for their event


joemsak:
**Describe the bug**
When I create an event as an RA, and I go to manage teams or judges, trying to pull up the list of teams / judges to choose just spins. It appears that the server is having a 404 error.

**To Reproduce**
1. Login as an RA
2. Create a new event
3. Go to Manage Teams or Judges
4. Click to add teams
5. See that it spins / web developer console says 404 not found error

**Expected behavior**
I expected a list of teams or judges to appear

**Additional context**
This needs to be fixed immediately to enable exploratory QA

joemsak:
@adrichey just a quick update for you, this is not a 500 server error, the javascript is not executing properly. Do you want to screenshare and look at it together? Let me know what you prefer. Thanks!

adrichey:
I will take another look at this as time allows today then. Can you please put the details of the error and what you think the issue is in this thread for tracking? Thanks!

joemsak:
Sure! Thank you, I really appreciate that. I'd hate to leave you on your own, mostly due to how poorly I wrote this code, so if you do feel like a screensharing session would be more productive please don't hesitate to ask.

Anyway, the main starting point of the bug seems to be with this bit:

https://github.com/Iridescent-CM/technovation-app/blob/qa/app/javascript/ra/events/EventsTable.vue#L219

It apparently isn't triggering when it should, so the event object doesn't get that fetchTeams/fetchJudgesUrlRoot property, and so when you go to manage teams/judges you'll see the rails server is being asked for a GET on /regional_ambassadors/undefined

I haven't been able to figure out more beyond that yet, hence why I was hoping to get your help.

Thanks again and please let me know if i can do more to be helpful to you.

adrichey:
@joemsak I have fixed the front-end issue of this ticket in branch `1804-error-when-managing-events`. I have created the following pull request for it: https://github.com/Iridescent-CM/technovation-app/pull/1827

The 500 error remains, even with the correct JSON endpoints in there due to the `AttendeesSerializer` not getting an ID attribute needed for the fastJSON API. Can you please take another crack at it in that `1804-error-when-managing-events` branch and then let me know once you have made progress? Thanks!

Attaching screenshots so you can see the results of the errors.
[...]

PR #1827, opened 2018-11-07
Fix 500 errors when managing teams and judges in RA events section

adrichey:
Addresses #1804.

This is not finished. I am passing this to @joemsak so he can fix the back-end serializer issues that are causing the actual 500 errors when the correct back-end endpoint is hit.

Please note that the fix to `app/javascript/application/sticky-header.js` is not needed for this, but it is a necessary addition I am squeezing in.

### DIFF
--- a/app/javascript/ra/events/EventsTable.vue
+++ b/app/javascript/ra/events/EventsTable.vue
@@ -200,6 +200,9 @@
           return e.id === event.id;
         });
 
+        event.fetchTeamsUrlRoot = this.teamsListUrl;
+        event.fetchJudgesUrlRoot = this.judgesListUrl;
+
         if (idx !== -1) {
           this.events.splice(idx, 1, event);
         } else {
--- a/app/serializers/attendees_serializer.rb
+++ b/app/serializers/attendees_serializer.rb
@@ -1,8 +1,7 @@
 class AttendeesSerializer
   include FastJsonapi::ObjectSerializer
 
-  attributes :id,
-    :name,
+  attributes :name,
     :scope,
     :status,
     :human_status,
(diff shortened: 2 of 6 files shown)


app/javascript/ra/events/EventsTable.vue
```diff
@@ -200,6 +200,9 @@
           return e.id === event.id;
         });
 
+        event.fetchTeamsUrlRoot = this.teamsListUrl;
+        event.fetchJudgesUrlRoot = this.judgesListUrl;
+
```
joemsak: Oh geez 😅 it was so simple. Thank you Alan

joemsak:
I'm going to check this branch out locally and investigate the remaining issues. Thanks again Alan!

joemsak:
@adrichey please see and confirm my changes, thank you!

app/javascript/ra/events/EventsTable.vue
adrichey: Welcome haha. I figured you would have that reaction. I know that feeling all too well. :)

app/serializers/attendees_serializer.rb
```diff
@@ -1,8 +1,7 @@
 class AttendeesSerializer
   include FastJsonapi::ObjectSerializer
 
-  attributes :id,
-    :name,
+  attributes :name,
```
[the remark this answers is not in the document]
adrichey: That makes sense!

adrichey:
This looks good @joemsak. I believe this is ready to be merged unless you think otherwise. Can I go ahead and do so?

joemsak:
Thank you Alan, yes please!

PR #1830, opened 2018-11-08
Apply reverse geocoding simply if coordinates are valid, not based on changes to data

joemsak:
This is still a work in progress to address #1824

### DIFF
--- a/lib/tasks/fix_empty_states.rake
+++ b/lib/tasks/fix_empty_states.rake
@@ -7,13 +7,38 @@ task fix_empty_states: :environment do
   puts "Reverse geocoding now..."
 
   accounts.find_each do |account|
+    puts "===================================="
+
+    city = account.city
+    state = account.state
+    country = account.country
+
     account.reverse_geocode
-    account.save
+
+    if state != account.state
+      puts "Account##{account.id} GEOCODING FIXED:"
+      puts "City from: `#{city}` to `#{account.city}`"
+      puts "State from: `#{state}` to `#{account.state}`"
+      puts "Country from: `#{country}` to `#{account.country}`"
+
+      account.geocoding_fixed_at = Time.current
+      account.geocoding_city_was = city
+      account.geocoding_state_was = state
+      account.geocoding_country_was = country
+      account.save
+    else
+      puts "*********** !!! ************"
+      puts "Account##{account.id} had NO CHANGES:"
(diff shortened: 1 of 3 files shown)


joemsak:
@adrichey @stenington please let me know if you can think of any other suggestions beyond what I've done here, as I am feeling a bit stumped at the moment. All the same, I'll try to spend some time in careful thought on the problem. Thank you so much for any effort you may be able to put into this

joemsak:
@adrichey what do you think about doing something like this

http://lab.iamrohit.in/php_ajax_country_state_city_dropdown/

Where maybe "City" lets you enter your own text? I worry about forcing a choice of cities in case data is incomplete or missing. For example in choosing Hong Kong as the country, I got no choices of cities

joemsak:
World city data...
Maybe using the Geonames API?

http://www.tweetegy.com/2011/02/how-to-get-a-semantic-list-of-all-countries-and-locations-in-the-world/

http://www.geonames.org/products/premium-data.html

lib/tasks/fix_empty_states.rake
```diff
@@ -0,0 +1,44 @@
...
+      account.geocoding_country_was = country
+      account.save
+    else
+      puts "*********** !!! ************"
+      puts "Account##{account.id} had NO CHANGES:"
+      puts "City from: `#{city}` to `#{account.city}`"
+      puts "State from: `#{state}` to `#{account.state}`"
+      puts "Country from: `#{country}` to `#{account.country}`"
+    end
+  end
```
joemsak: @adrichey Take a look at this rake task now - it logs changes to the accounts if any are detected (mainly if the state changes)

I would recommend we run this task like so:

```
rake fix_empty_states > fix_geocoding.log
```

and then we could keep that record in the log file, maybe consider saving it in source? Let me know what you think, and if you have any other suggestions. Thanks!

adrichey:
@joemsak I am cool with cherry-picking so that we don't have to wait for the next release, but I think I would want Alli to chime in as our product owner to let us know how quickly this needs to be on production. So I will respond to the card as such. Cherry-picking in a hotfix sounds like a great thing we should probably document soon.

joemsak:
Alright, and I'm not sure if you're aware, but I am not authorized to merge PRs, so make sure you do so when you're ready. Thanks!

adrichey:
Merged!

Issue #1832, opened 2018-11-08
As an RA, I cannot select more than one team to add to an event


Allicolyer:
**Describe the bug**

**To Reproduce**
1. Log in as an RA
2. Click on events
3. Click on 'manage teams'
4. Select a team to add to your event
5. Once one team is select the other options disappear and I cannot 'unselect' that team
6. I am are unable to select another team to add to the event or 'unselect' the team, even if I close and reopen the 'manage teams' menu 

**Expected behavior**
I would expect to be able to select more than one team at a time to add to my event.

**Screenshots**
![events-add-teams](https://user-images.githubusercontent.com/11083917/48219673-f81f9980-e341-11e8-92b8-e85079704bbf.gif)

**Desktop (please complete the following information):**
 - OS: MacOS Mojave
 - Browser: Chrome

adrichey:
I believe I have a fix for this. Assigning to myself and beginning work.

Allicolyer:
@adrichey this works now! Thank you!

Issue #1831, opened 2018-11-08
As an RA I cannot save teams or judges that I invite to my event


Allicolyer:
**Describe the bug**
As an RA I can select a team to add to my event, but I can't save them

**To Reproduce**
1. Log in as Ra@ra.com
2. Click on Events
3. Find an event and click on 'manage teams'
4. Invite a team and try to click 'save teams'
5. Nothing happens 
6. When the page is refreshed the teams have not been added

**Expected behavior**
I expect to be able to add the teams and judges.

**Screenshots**
![screen shot 2018-11-08 at 10 19 32 am](https://user-images.githubusercontent.com/11083917/48219259-e689c200-e340-11e8-9c72-5ea5c1ba0202.png)

**Desktop (please complete the following information):**
 - OS: MacOS Mojave
 - Browser: Chrome

adrichey:
I believe I have a fix for this. Assigning to myself and beginning work.

Allicolyer:
@adrichey this is working great now! I will test it out futher to find more bugs and then file more bug reports.

PR #1836, opened 2018-11-08
Events Attendee models missing ids

adrichey:
Addresses #1831 and #1832.

Basically the switch to fastJSON API and serializers caused the IDs to not be on the same level as the attendee attributes, so the Attendee JavaScript model objects weren't populating correctly (the had `NaN` as their ids). This in turn made the Vue loops crash and burn, and therefore everything blew up violently.

### DIFF
--- a/app/javascript/ra/events/AttendeeSearch.vue
+++ b/app/javascript/ra/events/AttendeeSearch.vue
@@ -175,6 +175,8 @@
 
           success: (json) => {
             Array.from(json.data || []).forEach(obj => {
+              // Need to massage this data since serializers modify the JSON structure
+              obj.attributes.id = obj.id
               const item = new Attendee(obj.attributes)
               let idx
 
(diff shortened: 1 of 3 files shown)


joemsak:
Ah, I see, so when I removed it from attributes, I should have checked if that broke anything.

Do you think it might be better if instead of patching the attributes of the response object, that the object just reads the response object correctly? Or do you anticipate that that will cause more unforeseen issues? Let me know your preference. Thanks Alan!

joemsak:
Looking a bit closer I can see that it is the attributes property being passed in the initializer in the first place. Let me know if you have anything to add to that, otherwise I will mash the approve button. Thanks again Alan

joemsak:
Thanks Alan. So as I mentioned in the work I did before, it is redundant to add ID in the attributes property, because it is there in the root data object. But if you strongly prefer we just go with that, I will gladly reintroduce that bit to the serializer. Mind if I do so on this branch?

Issue #1826, opened 2018-11-07
RA creating event: Time select is broken


joemsak:
**Describe the bug**
As an RA when I am creating a new event, and go to select start / end times, there is no menu to pick from times

**To Reproduce**
1. Login as an approved RA
2. Click on "Events"
3. Add a new event
4. Focus the mouse in the start / end time fields
5. See that no menu appears to pick times

**Expected behavior**
I expected a dropdown menu to select standardized times

adrichey:
@joemsak I can address this issue. What format should the time be in when it comes to the backend? 24-hour time (17:00), for example.

joemsak:
Hey Alan, thanks for taking an interest in this bug. For what it's worth, since it's brand new and wasn't discussed in sprint planning, I'm not sure what the priority would be for it just yet, but I will leave that decision up to everyone else. ANYWAY to answer your question regardless, the menu was previously working with 24 hour time. It's my belief that you ought to find existing code that simply isn't executing properly.

adrichey:
Noted! I will talk to Alli and see if we can make some time for this soon. I will take a look into the previous functionality soon. Thanks for making this card and putting visibility on this!

joemsak:
Sure thing, I discovered it while trying to fix the critical bug :)

adrichey:
After speaking with @Allicolyer we have decided to pull this into the sprint. Beginning work now. :)

Allicolyer:
@adrichey this is working great.

1. If I try to make my event start after it ends, the time picker will adjust itself. 
2. The times I select get converted to 24 hour time. 

@kmorton2017 Could you check this out? I know that most of our RAs probably don't use AM & PM, but I think this is a good solution.

adrichey:
@Allicolyer and @kmorton2017 I can also set the time pickers to 24-hour time if that works better. Let me know if that is desired and I can do that!

joemsak:
Just a note in support of 24 hour time, I won't say which RA but when we had AM/PM time, one of our RAs picked their times in the very early morning when they meant to pick them in the afternoon, so I would advocate for 24 hour time

Allicolyer:
Thanks @joemsak! I think @kmorton2017 and @ophelieh should make the final call!

ophelieh:
I would recommend the 24 hour time as only a few countries actually use the AM/PM time. 

Thank you @joemsak @adrichey @Allicolyer!

Allicolyer:
[...]

PR #1842, opened 2018-11-09
Update event time pickers to use flatpickr

adrichey:
Addresses #1826.

When `element-ui` was removed for pull request #1632, we didn’t update the time pickers to use a different library. I have implemented the time pickers using flatpickr. I have also added some functionality to prevent end-users from adding a start time after the end time to improve the UX a bit and prevent potential errors.

### DIFF
--- a/app/javascript/ra/events/EventForm.vue
+++ b/app/javascript/ra/events/EventForm.vue
@@ -80,7 +80,7 @@
             From
             <datetime-input
               v-model="eventStartTime"
-              :options="timeOpts"
+              :options="startDateInputOptions"
             />
           </label>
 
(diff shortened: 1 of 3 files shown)


app/javascript/ra/events/EventForm.vue
```diff
@@ -331,6 +353,18 @@
...
+
+      createTimeFromString(twentyFourHourTime) {
+        const newTime = new Date();
+        const timeParts = twentyFourHourTime.split(':');
+        newTime.setHours(
+          parseInt(timeParts[0], 10),
+          parseInt(timeParts[1], 10),
+          0
+        );
+        console.log('createTimeFromString called for', twentyFourHourTime, newTime);
```
joemsak: Just wanna check with you if this console.log is meant to stay here, otherwise, thanks Alan it all looks great to me

adrichey:
@joemsak I meant to get rid of that log. My bad. I will need you to approve this again with the new Github rules we have in place. I got rid of that log. :)

PR #1848, opened 2018-11-16
Registration UI label updates for Vue

adrichey:
Update label population code for registration Vue app to account for state and birthdate objects being returned in addition to strings in the UI.

### DIFF
--- a/app/javascript/registration/store/getters.js
+++ b/app/javascript/registration/store/getters.js
@@ -60,7 +60,15 @@ export default {
   },
 
   getBirthdate (state) {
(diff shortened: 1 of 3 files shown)


adrichey:
@joemsak I think this might be part of the issue that is breaking the location form. If a user comes back to the form right now, they get "[object Object]" as their state on any other router view within registration except the location form. This essentially safeguards against that. I think we will want to get this into production soon to potentially fix their state issues. Tagging @stenington for visibility to that effect.

adrichey:
The location functionality is the bane of our existences at this point haha. Thanks for the review Joe! 😄

PR #1851, opened 2018-11-19
Location form country responses bugfix

adrichey:
Country codes and countries were not populating properly in the current locations controllers after an update to the `FriendlyCountry` class. This addresses that issue.

### DIFF
--- a/app/controllers/admin/current_locations_controller.rb
+++ b/app/controllers/admin/current_locations_controller.rb
@@ -15,17 +15,14 @@ module Admin
         prefix: false,
       )
 
-      country = FriendlyCountry.(
-        OpenStruct.new(country: record.country),
-        prefix: false,
-      )
+      country = FriendlyCountry.new(record).country_name
 
       render json: {
         city: record.city,
         state: state,
         state_code: record.state_province,
         country: country,
-        country_code: record.country,
+        country_code: record.country_code,
       }
     end
   end
--- a/app/controllers/admin/current_locations_controller.rb
+++ b/app/controllers/admin/current_locations_controller.rb
@@ -7,22 +7,20 @@ module Admin
         record = Team.find(params.fetch(:team_id))
       end
 
-      state = FriendlySubregion.(
-        OpenStruct.new(
-          state_province: record.state_province,
-          country: record.country
-        ),
+      state = FriendlySubregion.(record, prefix: false)
+      state_code = FriendlySubregion.(record, {
         prefix: false,
-      )
+        short_code: true
+      })
 
-      country = FriendlyCountry.new(record).country_name
+      friendly_country = FriendlyCountry.new(record)
 
       render json: {
         city: record.city,
         state: state,
-        state_code: record.state_province,
-        country: country,
-        country_code: record.country_code,
+        state_code: state_code,
+        country: friendly_country.country_name,
+        country_code: friendly_country.as_short_code,
       }
     end
   end
--- a/app/controllers/judge/current_locations_controller.rb
+++ b/app/controllers/judge/current_locations_controller.rb
@@ -1,22 +1,20 @@
 module Judge
   class CurrentLocationsController < JudgeController
     def show
-      state = FriendlySubregion.(
-        OpenStruct.new(
-          state_province: current_account.state_province,
-          country: current_account.country
-        ),
+      state = FriendlySubregion.(current_account, prefix: false)
+      state_code = FriendlySubregion.(current_account, {
         prefix: false,
-      )
+        short_code: true
+      })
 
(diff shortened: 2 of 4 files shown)


app/controllers/admin/current_locations_controller.rb
```diff
@@ -15,17 +15,14 @@ def show
         prefix: false,
       )
 
-      country = FriendlyCountry.(
-        OpenStruct.new(country: record.country),
-        prefix: false,
-      )
+      country = FriendlyCountry.new(record).country_name
```
joemsak: So, I think what you would want here, in using this class the way it's been intended to... I can totally, totally understand it not being so obvious, but you would want maybe more like this:

```ruby
friendly_country = FriendlyCountry.new(record)

# ....
country: friendly_country.country_name,
country_code: friendly_country.as_short_code,
# ...
```

You may look at that and say, well 6 of one, half dozen of the other, and you may totally be justified in that. It's just that what I'm advocating for is, if we are using that FriendlyCountry class as our value helper object, then let's always use it as the source of truth. Please let me know if I need to explain that better.

app/controllers/judge/current_locations_controller.rb
```diff
@@ -9,17 +9,14 @@ def show
...
-      )
+      country = FriendlyCountry.new(current_account).country_name
 
       render json: {
         city: current_account.city,
         state: state,
         state_code: current_account.state_province,
         country: country,
-        country_code: current_account.country,
+        country_code: current_account.country_code,
```
joemsak: So, then, again, I would give the same advice here as above

app/controllers/mentor/current_locations_controller.rb
```diff
@@ -37,7 +37,7 @@ def show
           state: state,
           state_code: current_account.state_province,
           country: country,
-          country_code: current_account.country,
+          country_code: current_account.country_code,
         }
       end
     end
```
joemsak: And likely here, as well, though I can't see the full context in this diff but I assume FriendlyCountry was used above. In fact, I see we go to state_province on the account / team, when again it should come from FriendlySubregion...

Please let me know if what I am saying is bonkers or needs more explanation

app/controllers/student/current_locations_controller.rb
```diff
@@ -14,7 +14,7 @@ def show
         state: FriendlySubregion.(record, prefix: false),
         state_code: record.state_province,
         country: country,
-        country_code: record.country,
+        country_code: record.country_code,
       }
```
joemsak: Wash, rinse, repeat here.

Otherwise, Alan, I think you did a great job figuring out what did figure out and I am absolutely sure that what you did works and is fine, this is now a discussion of, what should be used consistently moving forward? Thanks!

app/controllers/admin/current_locations_controller.rb
adrichey: That totally makes sense to me! I will make that change as needed throughout the PR when I have time this afternoon. :)

adrichey:
@joemsak Your comments make total sense and make a more robust solution. Less room for failure since we are using the class as the source of truth. I will get that changed before too long and let you know when this is ready for review again. Thanks!

joemsak:
Oh that is great Alan thank you! And just a quick note that we see state_province is being referenced directly on the record in all those spots too (not just the one I noticed at first) - Would you be able to use FriendlySubregion for this? They should be saved as their short-codes in the database for consistent filtering / sorting / searching. Thanks!

joemsak:
And finally, when you want to *read* the values on the object (not when deciphering and saving them) you can turn your attention to this mixin which *should* be included in every model that has location data:

https://github.com/Iridescent-CM/technovation-app/blob/qa/app/models/concerns/active_geocoded.rb

See the instance methods that start at Line 80... let me know if you need assistance in grokking this... I admit that, especially around location stuff, my code has gotten more and more confused about itself... due to all these ongoing bugs... :/ I was trying to finally sort it all out in the following way:

When you want to get location information from a source of truth (usually for saving), use FriendlyCountry / FriendlySubregion to get the data from the 3rd party helpers

When you want to read what is already saved on the record, use the instance methods in this mixin

I'll be honest that I could be totally overthinking / overengineering this whole freakin thing

adrichey:
All right Joe, I have made changes to this according to your feedback earlier. I didn't rewrite `FriendlySubregion` to match the methods found on the class `FriendlySubregion` as I felt like that is more work than necessary right now and we have a lot going on, but I am now using that `FriendlySubregion` to populate the state and state codes. Hopefully that works!

joemsak:
Oh i TOTALLY get that, dude, thank you!!!

Issue #1854, opened 2018-11-19
RA: Link to the submission page does not work the team page when accessing the team page from an event


Allicolyer:
**Describe the bug**
Link to the submission page does not work the team page when accessing the team page from an event

**To Reproduce**
1. Log in as an RA
2. Click on one of the teams attending an event
3. On the team page, scroll down to the submission
4. Get an error when you click on the link to the submission

**Expected behavior**
I would expect to be linked to the submission page.

**Screenshots**

![view-submission](https://user-images.githubusercontent.com/11083917/48726039-faf07900-ebe2-11e8-9e55-2faeeed06614.gif)

**Desktop (please complete the following information):**
 - OS: MacOS Mojave
 - Browser: Chrome

**Additional Information**

The submission page works when it is accessed from the team page through the team menu.

Allicolyer:
@joemsak and @adrichey would this be easy to fix?

Allicolyer:
Hey @joemsak and @adrichey I tried this again today and it did not work, I was still redirected to the home page with a "you don't have permission to go there" message.

adrichey:
I am also seeing that error message. I will look into that today @Allicolyer.

adrichey:
This code works for me locally, so I am assuming that it didn't get deployed to QA properly. There are also tests written around this that passed, so my guess is that the code just hasn't been deployed properly, or that there is a caching issue. Continuing to investigate, but I don't think any new code needs to be written.

adrichey:
Turns out it didn't get deployed yesterday to QA. I have deployed that manually. This is once again ready for testing. Thanks!

Allicolyer:
Thanks Alan! It's working great now!

PR #1855, opened 2018-11-19
Fix the RA's link to the team's submission in the debugging area

joemsak:
This PR addresses #1854 by:

* Adding a system spec (regression test) for the RA clicking on the link from the team debugging area
* Using the rails convention of putting route names together with an array, and the custom helper method of `current_scope` which knows if you're logged in as an RA, admin, mentor, etc...

### DIFF
--- a/app/views/admin/teams/_debugging.html.erb
+++ b/app/views/admin/teams/_debugging.html.erb
@@ -7,8 +7,7 @@
     <p>
       View this team's submission:
 
-      <%= link_to team.submission.app_name,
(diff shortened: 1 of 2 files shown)


joemsak:
@adrichey for your reference of how this array thing works... https://makandracards.com/makandra/805-generate-a-path-or-url-string-from-an-array-of-route-components

adrichey:
Code look good! Thanks for the explanation on that `link_to` change. I appreciate it!

Issue #1846, opened 2018-11-12
Add fields to collect for survey


Allicolyer:
Hi @kmorton2017 and @ophelieh, can you tell me if this is all the fields that we need to collect for the pre-survey?

First and Last Name,
*School
City 
State / Providence
Country
Email Address
*Team Name

The ones that have a star* are the ones that Joe and Alan need to add functionality for. Are there any other new fields that you need this year?

ophelieh:
Looks right to me. We could possibly add age. This could tell if age makes a difference in filling out the survey or not.

kmorton2017:
@Allicolyer just noticed a typo it should be "State / Province"

For mentors it should gather: 
First and Last Name,
**Company**
City
State / Province
Country
Email Address
*Team Name

kmorton2017:
@alli just checking in to see if there is an update here! @joemsak

Allicolyer:
Hi @kmorton2017 and @ophelieh! We can move this into to-do. Do we want to collect age?

ophelieh:
Yes, please! @Allicolyer :) 

Thank you!!!!

Allicolyer:
Hey @joemsak can you work on this ASAP? The program team needs the following fields added to the student and mentor surveys:

Student Survey:
<ul>
<li> Team Name </li>
<li> Age </li>
<li> School </li>
</ul>

Mentor Survey:
<ul>
<li> Team Names </li>
<li> Company Name </li>
</ul>

adrichey:
@joemsak This would actually take priority over the location form. @Allicolyer and I spoke this morning and since this is a more straightforward fix and needs to be in our next production release -- which I have planned for this week once this is done -- it takes priority. Thanks in advance for being flexible on this. :slightly_smiling_face:

joemsak:
No prob, I will get on it once I return from the dentist. Got half my mouth
feelling like there’s a baseball in it right now haha


-- 
Joseph M. Sak
+1 312 945 6317
+52 1 33 23 10 69 05
work@joesak.com

joemsak:
@Allicolyer can you please tell me how each of those variables are named in survey monkey?

joemsak:
@Allicolyer Actually, let me amend my request: make sure all the variables are set as we walked through before, create the web collector URL and copy that into this card. That will be the most accurate representation of what the software should expect to replace with real data

adrichey:
[...]

PR #1865, opened 2018-11-27
1846 survey fields

joemsak:
This update adds the formatting on the survey links necessary for the student survey Web Collectors

TODO:
  * Pending the work needed for the Mentor survey web collector, which is
    * Pending staff to update the custom variables and provide the URL

Addresses #1846


adrichey:
Cool. All of this looks good so far. Hopefully we can get you the mentor URL early tomorrow so you can finish this up. Thanks Joe!

joemsak:
@adrichey after seeing the update mentor survey URL, this PR is already done, ready for QA

adrichey:
@joemsak I might be missing something here, but I am not seeing the `company=[company_value]` piece for mentors in their URL. We have `school=[school_value]` for students, but that isn't the same value for mentors. I believe we will need to add a check for whether not we are looking at a student or mentor and replace the `school=[school_value]` variable with `company=[company_value]` for mentors. Assuming I am not missing some sort of conversion piece somewhere.

joemsak:
ah yeah you're right, sorry, i overlooked it because in the database, school and company share the same field. will update ASAP

joemsak:
Also FWIW we don't have to check for the profile type because the URL will be formatted for just whatever variables are there (that the code is programmed to be aware of) - so either school or company will get filled in, and the code can actually just try to populate both. take note of `Account#profile_school_company_name`

joemsak:
Okay pending CI it should be good to go

PR #1868, opened 2018-11-28
Location form carmen rework fix specs

joemsak:
Still a WIP, @adrichey (having trouble getting "simple" specs to behave) but I figured you might want to see the diffs before this work goes into your branch (note that the PR is asking to merge into your branch, not `qa`)


spec/system/mentor/location_spec.rb
```diff
@@ -29,9 +29,11 @@
     click_button "Profile"
     click_button "Region"
 
+    # TODO - this cannot stand >:[
+    sleep 1
```
joemsak: @adrichey here's an example of me trying to deal with misbehaving specs - I don't know why the forms aren't working if I don't sleep :( - note this is not the only place this has happened

adrichey: Noted. I will see what I can come up with on that front. I agree, sleeping is not a sustainable approach as we add more tests. I will dig into that test this morning.

joemsak: And sometimes sleep 1 doesn't work :/ i'm also trying to dig into it

adrichey: I think it might have something to do with the vue_select_input_helper. The value isn't persisting properly as it is built right now. I think we need to find a better way to populate those fields, but I am not sure how right now. I will continue digging.

joemsak:
Okay, in the meantime, I'm revisiting this post

https://robots.thoughtbot.com/headless-feature-specs-with-chrome

And finding that I didn't implement everything in it... strange

joemsak:
Not sure why but I didn't do this part...

https://robots.thoughtbot.com/headless-feature-specs-with-chrome#capybara-configuration

I don't know if I found something out or just plain... missed it? Trying it now

spec/system/location/team_locations_spec.rb
```diff
@@ -12,10 +12,12 @@
       expect(page).to have_content("Chicago, Illinois, United States")
 
       click_link "Change this team's location"
-      fill_in "State", with: "California"
-      fill_in "City", with: "Los Angeles"
+
+      select_vue_select_option "#location_country", option: "United States"
+      select_vue_select_option "#location_state", option: "California"
```
joemsak: @adrichey I'm having an issue in this spec, it's not filling in the country option and therefore doesn't find the CSS for the state selector. Have I used this code correctly?

joemsak: hold on this for just a moment, other changes in my local WD might be making things work better.... will update you

joemsak: haha right as i type that, this test fails for the same reason again :)

```
Team locations changing an existing team location works fine
     Failure/Error: page.find(selector, wait: 10).click

     Capybara::ElementNotFound:
       Unable to find visible css "#location_state"

     [Screenshot]: tmp/screenshots/failures_r_spec_example_groups_team_locations_changing_an_existing_team_location_works_fine_375.png



     # ./spec/support/vue_select_input_helper.rb:8:in `select_vue_select_option'
     # ./spec/system/location/team_locations_spec.rb:17:in `block (3 levels) in <main>'

Finished in 1 minute 42.55 seconds (files took 3.5 seconds to load)
83 examples, 1 failure
```

I will push my latest work up now

joemsak: I *think* I have everything passing with this test skipped

adrichey: Cool. I am wrapping up lunch (I am migrating to a coffee shop for the afternoon) and then I will hop on this and see what I can come up with.

joemsak: oyyyyyy some of these tests are only failing in the full suite (and not in isolation)

...and then only sometimes :-/

adrichey: Got it! The form just needed to be cleared before you set the values. By setting the country field to United States when it was already set actually caused it to be de-selected, so putting the form clear in there fixed that. We should be good to go! See commit: 38e57f3211fece20b9fc70f09cb07d897aea1b69

adrichey: I am running the full suite now and will look into any failures as well.

joemsak:
@adrichey just a heads up i'm futzing with the various setup in terms of gems and rspec config.... seems like i may have mixed too many strategies, so i'm trying to pare it back down to the bare minimum needed

joemsak:
Alright @adrichey ... pending CI! 🤞

joemsak:
@adrichey hmmm all I can say is, the tests passed locally :( I'm going to have to give this one a rest for the day and try with a fresh brain tomorrow I suppose

adrichey:
That works @joemsak. I have been knocking out other things this afternoon and my brain is getting pretty frazzle-dazzled as well. I will take another sweep at it tomorrow too. Hopefully we can get it sorted then. Thanks for your work on this today.

joemsak:
current hypothesis - this seems dependent on the randomized order

How to replicate: check what 'seed' the tests are ran with in the CircleCI report and you can re-use it

```
[ CIRCLE CI SECTION: Parallel RSpec ]
#!/bin/bash -eo pipefail
bundle exec rspec --order rand

/home/circleci/circleci-demo-ruby-rails/vendor/bundle/ruby/2.5.0/gems/has_secure_token-1.0.0/lib/active_support/core_ext/securerandom.rb:4: warning: already initialized constant SecureRandom::BASE58_ALPHABET
/home/circleci/circleci-demo-ruby-rails/vendor/bundle/ruby/2.5.0/gems/activesupport-5.1.6/lib/active_support/core_ext/securerandom.rb:4: warning: previous definition of BASE58_ALPHABET was here

Randomized with seed 29738
```

Take that seed and confirm it fails locally:

```
bundle exec rspec --seed 29738
```

and then bisect it

```
bundle exec rspec --bisect --seed 29738
```

Which is what I am doing now...

joemsak:
okay, so for some reason... unrelated tests that play before these ones with the Login.vue component... cause the usual partial label feature not to work

meaning... usually if you have a label/input like this

```html
<label for="email">Email Address</label>
<input id="email" type="email" />
```

In capybara you can use

```ruby
fill_in "Email", with: "joe@example.org"
```

because "Email" is a partial match and it's fine

this also works for `click_button` and `click_link`

but for whatever reason, they now have to be a 100% match... sometimes.

so... okay. here comes commit-FINAL-FINAL-FINAL2.PSD

joemsak:
@adrichey ready!!!!!!!!!!!!!!!!!!!!!

spec/models/account_spec.rb
```diff
@@ -205,10 +205,38 @@
...
+    Timecop.freeze(Date.new(2016, 2, 29)) do
+      expect(account.age).to eq(14)
+    end
+
+    Timecop.freeze(Date.new(2016, 3, 1)) do
+      expect(account.age).to eq(15)
+    end
+  end
+
+  it "calculates age correctly for leap year birthdays" do
```
joemsak: @adrichey looks like you'll want to rebase your branch on qa to eliminate this diff

adrichey:
@joemsak I rebased `qa` into `location-form-carmen-rework` and then `location-form-carmen-rework` into this branch. That way it eliminated all the duplicate code commits that are in all three places. So all that remains in here is your work. Since this is ready, pending a green CI pass, I think we will be ready to merge this into the parent branch! Woo!

Issue #1837, opened 2018-11-08
On a team page,  make pending students clickable to view their profile pages


Allicolyer:
**Is your feature request related to a problem? Please describe.**
Admins cannot click on a pending student's or mentor's profile from a team page

**Describe the solution you'd like**
Admins would like to be able to click on a pending students and mentor from their team pages, if their profile exists.

![screen shot 2018-11-08 at 11 13 52 am](https://user-images.githubusercontent.com/11083917/48221631-716dbb00-e347-11e8-9cb6-ac79a2da5275.png)

**Acceptance Criteria**

As an admin
I can go to a team page
Which has a pending student who is not fully on-boarded
And click on the student's name to visit their profile page

As an admin
I can go to a team page
which has a pending mentor who is not fully on-boarded
and click on the mentor's name to visit their profile page

The only caveat here is when the student has not even registered their email yet. In that case no student profile exists, so their name should not be clickable.

Pending students are students who are on teams, but have not finished all of their onboarding steps.

Allicolyer:
@adrichey I updated the acceptance criteria with a caveat.

adrichey:
Thanks @Allicolyer! That is helpful.

adrichey:
This is now ready for testing!

Allicolyer:
Thank you @adrichey. Would you be able to add this for RAs too? Should I make a new card for that or should we keep it on this card?

adrichey:
This is also available for RAs, so you should be able to test that functionality now @Allicolyer. :slightly_smiling_face:

PR #1869, opened 2018-11-28
Make pending students on a team clickable within team page in admin UI

adrichey:
Addresses #1837.

### DIFF
--- a/app/views/teams/_onboarding_member.en.html.erb
+++ b/app/views/teams/_onboarding_member.en.html.erb
@@ -5,7 +5,22 @@
   </div>
 
   <div class="grid__col-10">
-    <h5 class="reset"><%= onboarding_member.full_name %></h5>
+    <h5 class="reset">
+      <% if onboarding_member.student_profile.present? %>
+        <%= link_to onboarding_member.full_name,
+          [
+            current_scope,
+            :participant,
+            {
+              id: onboarding_member.account_id,
+              allow_out_of_region: true,
+            }
+          ] %>
+      <% else %>
+        <%= onboarding_member.full_name %>
+      <% end %>
+    </h5>
+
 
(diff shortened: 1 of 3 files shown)


app/views/teams/_onboarding_member.en.html.erb
```diff
@@ -5,7 +5,22 @@
...
+            :participant,
+            {
+              id: onboarding_member.account_id,
+              allow_out_of_region: true,
+            }
+          ] %>
+      <% else %>
+        <%= onboarding_member.full_name %>
+      <% end %>
+    </h5>
```
joemsak: hmmm this seems like it would affect all users, not just admins, and so the link would break for students/mentors

also, shouldn't need to check if the student profile is present, as only students can be onboarding and on a team at the same time, plus why not link anyone regardless?

tip: pass in the partial local variable "ra_admin" (a boolean) to detect if you should create a link or not - i imagine it's already being passed in somewhere along the line, but if not, you can do it yourself - also brings up another point: why not enable this for RAs as well? Staff *often* asks for admin features without remembering that they will want it for RAs too. Always good to check and ask

joemsak: (hint search around the code for "ra_admin" to see how you might use it here ... usually i default my locals at the tops of my partials, so look for "ra_admin ||= false" to see how that might work)

adrichey: @joemsak Would you mind listing these profile variables in documentation somewhere, or at least point me to where I might find those more readily next time I need them? I would really appreciate that. I should have this PR ready for another look shortly. Thanks again for the feedback!

adrichey:
This is once again ready for your approval. Thanks for the guidance on this! 🙂

Issue #1813, opened 2018-11-05
Suspicious Scores has a 500 error and is not showing up 


Allicolyer:
Suspicious Scores has a 500 error and is not showing up 

1. As an admin, turn on Quarterfinals Judging
2. Log in as an on-boarded judge who is not a mentor and who has not already judged any teams
3. Judge and submission and give the lowest score possible
4. Log in as as admin and confirm that the score did not show up under "suspicious scores"

I would expect this score to show up under "suspicious scores" but it does not. Instead it appears that the "suspicious scores" are just loading forever.

Needed for Judging. Absolute deadline is in April.

![uploads/bf92aaa2-71e7-4a69-8994-c0e86d405c92/Screen Shot 2018-11-05 at 4.31.22 PM.png](https://s3-us-west-2.amazonaws.com/prod.huboard.com/uploads%2Fbf92aaa2-71e7-4a69-8994-c0e86d405c92%2FScreen+Shot+2018-11-05+at+4.31.22+PM.png)

**Desktop (please complete the following information):**
 - OS: MacOS Mojave
 - Browser Chrome

adrichey:
I have a fix for this. So after speaking with @Allicolyer we have decided to pull this into the To Do column. I should have a pull request in for this today.

adrichey:
This is now complete and ready for QA!

Allicolyer:
@nickiredescent thank you for the throughout QA! We should look into some of these discrepancies for the scoring sprint.

PR #1870, opened 2018-11-29
Fix incorrect serializer initialization for suspicious scores

adrichey:
Addresses #1813.

### DIFF
--- a/app/controllers/admin/suspicious_scores_controller.rb
+++ b/app/controllers/admin/suspicious_scores_controller.rb
@@ -2,7 +2,10 @@ module Admin
   class SuspiciousScoresController < AdminController
     def index
       suspicious_scores = SuspiciousSubmissionScores.new
-      render json: SuspiciousScoreSerializer.new(suspicious_scores.scores).serialized_json
+      render json: SuspiciousScoreSerializer.new(
+        suspicious_scores,
+        is_collection: true
+      ).serialized_json
     end
   end
 end
--- a/app/controllers/admin/suspicious_scores_controller.rb
+++ b/app/controllers/admin/suspicious_scores_controller.rb
@@ -2,7 +2,7 @@ module Admin
   class SuspiciousScoresController < AdminController
     def index
       suspicious_scores = SuspiciousSubmissionScores.new
-      render json: SuspiciousScoreSerializer.new(suspicious_scores).serialized_json
+      render json: SuspiciousScoreSerializer.new(suspicious_scores.scores).serialized_json
     end
   end
 end


joemsak:
And some study material on `Enumerable` FWIW

https://www.codementor.io/ruby-on-rails/tutorial/rubys-swiss-army-knife-the-enumerable-module

app/controllers/admin/suspicious_scores_controller.rb
```diff
@@ -2,7 +2,7 @@ module Admin
   class SuspiciousScoresController < AdminController
     def index
       suspicious_scores = SuspiciousSubmissionScores.new
-      render json: SuspiciousScoreSerializer.new(suspicious_scores).serialized_json
+      render json: SuspiciousScoreSerializer.new(suspicious_scores.scores).serialized_json
```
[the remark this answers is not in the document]
adrichey: This is a far better solution! Thank you for pointing this out @joemsak. Yeah, this is just my unfamiliarity with Ruby coming through. I need to look more into the serializers as they are one of my major grey areas.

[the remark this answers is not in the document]
joemsak: Oh it's totally all good, I would not have expected you to know about enumerable :)

adrichey:
@joemsak there is a test for the overall suspicious scores functionality, but not for the serializer response as far as I can tell.

joemsak:
Gotcha. I would say maybe to throw one in, since this was broken but no tests were failing. What do you think? If you agree, would you want to do that, or would you prefer that I do it? Just let me know, thanks dude!

spec/requests/admin/suspicious_scores_requests_spec.rb
```diff
@@ -0,0 +1,15 @@
+require "rails_helper"
+
+RSpec.describe "Admin / Suspicious Scores requests" do
+  describe "GET #index" do
+    it "returns OK" do
+      sign_in(:admin)
+
+      expect {
+        get admin_suspicious_scores_path
+      }.not_to raise_error
```
joemsak: @adrichey I did not start this test by having this wrapping expect block (although, I really like these kinds of expectations because they allow specificity of the action and result, which increases my confidence)

What happened was, I did not get a failure on the other expect below, instead the tests bombed out because of a fast_jsonapi exception (which is what was happening on production), and so I wrote this expect block in order to get a test failure instead of an exception booting the test out. I'm giving you this info just FWIW :)

And again, I removed the fix while running this test, so that I could be certain that the fix passes the test

joemsak:
@adrichey I went digging and found out why fast_jsonapi doesn't recognize enumerable mixed-in instances as collections:

https://github.com/Netflix/fast_jsonapi/blob/master/lib/fast_jsonapi/object_serializer.rb#L109

They check if it responds to size (an Array does, an Enumerable does not) and does not respond to each_pair (which is a Hash thing)

So FWIW another option is to add size methods to enumerable classes, or rather setup a delegator to the internal collection of the enumerable.

Just a thought, no actionable items here

spec/requests/admin/suspicious_scores_requests_spec.rb
adrichey: That is really good to know! Thanks for being so thorough with that description of your logic. That is helpful! 🙂

adrichey: Makes sense! This looks good. I think that should cover us for this particular issue. Thanks again Joe!

PR #1871, opened 2018-11-29
Fix age specs that fail on Nov 29 for upcoming leap years - make sure…

joemsak:
… leap year logic is working


joemsak:
@adrichey hope I didn't step on your toes, I went ahead and updated the spec and added a couple more around leap year logic

joemsak:
Thanks to you too man, sorry for the double effort, but glad to know we're on the same page at least!

PR #1877, opened 2018-12-03
Location form carmen rework follow up

joemsak:


joemsak:
@adrichey in the meantime of pondering what to do with existing data, this should clear up the remaining code issues and be ready for merge to your branch - worth pointing out that due to the current issue we're discussing that `FriendlyCountry` falls back to Geocoder searching for data, and I believe you'll want that removed before finalizing this.

Please do not hesitate to let me know of any other concerns you have, I will be happy to do my best

PR #1880, opened 2018-12-04
[WIP] Rework FriendlyCountry to work with Carmen instead of Geocoded

adrichey:


joemsak:
So just a quick distinction here -

Geocoded (with a d at the end) is different from Geocoder (with an r)

Geocoder is the gem we use which connects to mapping APIs

Geocoded is just a special object wrapping class I made to have a standard interface for dealing with the results from Geocoder

So the real trouble is the final fallback in FriendlyCountry to Geocoder after queries to Carmen fail

You'll note that Geocode**d** works with the countries gem (oh boy, I guess there are 3 freakin strategies at play here, and I am so sorry for that) - perhaps it would be better to change Geocoded to use Carmen? And also not fall back in Friendly Country to Geocoder?

The idea was to bury use of Carmen in a single class to expose methods to get the data as needed, so that way any trouble with Carmen can stay isolated there, and not across the codebase

Does that make sense? Let me know what I can clarify

joemsak:
So just a quick rundown of the pieces at play

- countries gem - I think was originally used to power the very first iteration of county / state dropdown but data was incomplete and UX was not great, so I scrapped it but I guess I kept it around for its usefulness of getting ISO codes ?? it can probably be scrapped

- carmen - better data, but very specific matches are necessary with varied data, hopefully the migration fixes this as using carmen to power the dropdowns and for massaging the data should make the problems go away

- geocoder - gem to connect to mapping APIs, now only to be used to grab Lat / Lng values from human friendly data (carmen formatted)

- geocoded - a custom wrapping object for exposing a consistent interface on geocoder results

- friendly-country - was originally used to get full country names out of country codes for Admin/RA screens, with the prefix option for quick sorting/scanning

- geocoding - a procedure module for running the geocoder methods on any supported model - with ability to shoehorn in more steps / callbacks to the procedure as needed


Ideally in my mind it should work like so...

Carmen powers the dropdowns
Values from the dropdowns go to the controller endpoints and save to the models directly

The controllers tell Geocoding to perform on the models
Geocoding leans on Geocoder for the API response
the API response is massaged by Geocoded for a consistent interface - this may not be so necessary anymore if geocoder is only used for lat/lng values

FriendlyCountry / Subregion leans on Carmen
Geocoded should also lean on Carmen

carmen stays isolated so that any changes to its interface or how the developers use it won't propagate across the codebase

When I write it all out it's very complex and I totally... TOTALLY admit that. It's the result of 2.5 years' worth of trial and error on what has been the single hardest feature to support for over 120 countries' worth of young age / ESL users

I would love a chance to throw it all out and start over but here we are :(

joemsak:
Annnnnd I missed ActiveGeocoded :)

ActiveGeocoded is rails concern for the models which need geocoding - this is to get a consistent interface that can be adjusted in one place, and a consistent behavior in terms of any callbacks or mixins included, again so as not to have to change any of those things across all supported models

adrichey:
@joemsak Thank you for the run down! This makes sense and I have looked through the code to verify things and it brings some clarity on pieces I was a little unsure about. I appreciate you explaining that.

I am going to work for this a bit more, and push things up before I clock out today. We want to have this location form stable by Friday, so I may need your help on this tomorrow in order to get this work out the door by then. 🙂

joemsak:
No problem Alan just let me know whatever you want my help on. I appreciate
you taking it upon yourself to learn how the ruby side works. You’re doing
a fantastic job. A lot of the trouble is how overly tangled my trial and
error patchwork has become and does not communicate itself very well. I
hope we can remove a lot of it and simplify it n
-- 
Joseph M. Sak
+1 312 945 6317
+52 1 33 23 10 69 05
work@joesak.com

stenington:
@joemsak's comments here make lots of sense to me. I think the only thing going in to `country` and `state_province` should be carmen country and subregion codes respectively, and then the various formatting helpers can use carmen reliably to turn a code back into a full name.

I'm not 100% sure what that's going to look like for getting lat/long though. Hopefully carmen data geocodes well?

joemsak:
By the way Alan - ultimately the design of how this works end to end (and back again) is completely up to you. I've merely laid out for you what my thought process was and a glimpse into how it got here over the years. It's been really eye opening for me to see my own work come back in such poor shape, and has given me pause for avoiding rushed solutions in my future projects.

PR #1886, opened 2018-12-11
#1824 slowly simplifying location stuff

stenington:
Detailed comments inline below.

### DIFF
--- a/app/controllers/registration/locations_controller.rb
+++ b/app/controllers/registration/locations_controller.rb
@@ -7,25 +7,6 @@ module Registration
         wizard_token: custom_location_params[:token]
       )
 
-      country_param = (custom_location_params[:country] || "").strip
-      country = Carmen::Country.named(country_param)
-      country_code = country && country.code
-
-      state_code = if country
-        state_param = (custom_location_params[:state] || "").strip
-        state = country.subregions.named(state_param)
-        state ||= country.subregions.coded(state_param)
-        state && state.code
-      else
-        nil
-      end
-
-      attempt.update!({
-        city:         custom_location_params[:city],
-        state_code:   state_code,
-        country_code: country_code,
-      })
-
       render json: SignupAttemptSerializer.new(attempt).serialized_json
     end
 
--- a/app/serializers/signup_attempt_serializer.rb
+++ b/app/serializers/signup_attempt_serializer.rb
@@ -3,7 +3,7 @@ class SignupAttemptSerializer
   set_key_transform :camel_lower
 
   attributes :email, :birth_year, :birth_month, :birth_day, :profile_choice,
-    :city, :state_code, :country_code, :latitude, :longitude, :first_name,
+    :city, :state, :state_code, :country, :country_code, :latitude, :longitude, :first_name,
     :last_name, :gender_identity, :school_company_name, :job_title,
     :referred_by, :referred_by_other, :wizard_token, :mentor_type, :expertise_ids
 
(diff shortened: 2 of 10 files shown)


app/controllers/registration/locations_controller.rb
```diff
@@ -7,25 +7,6 @@ def create
         wizard_token: custom_location_params[:token]
       )
 
-      country_param = (custom_location_params[:country] || "").strip
```
stenington: Removing all of this because it happens **after** we tell the user "Hey we saved your location as X, Y, Z" and it can change the X, Y, and Z. In particular, I believe it was responsible for nulling out the Spanish states, which we probably indicated to the user that we were actually saving.

app/serializers/signup_attempt_serializer.rb
```diff
@@ -3,7 +3,7 @@ class SignupAttemptSerializer
   set_key_transform :camel_lower
 
   attributes :email, :birth_year, :birth_month, :birth_day, :profile_choice,
-    :city, :state_code, :country_code, :latitude, :longitude, :first_name,
```
stenington: The normalized state and country attributes in this serializer were responsible for small bits of weirdness, like "Guadalajara, Jal., Mexico" changing to "Guadalajara, Jalisco, Mexico" depending on where you were in the wizard because not all pieces use the serializer.

app/technovation/geocoding.rb
```diff
@@ -34,7 +34,6 @@ def with_save
   module GeocodingUpdater
     def apply_geocoding_changes(controller)
       geocode if !valid_coordinates? || detect_location_changes?
-      reverse_geocode if valid_coordinates?
```
stenington: This was doing super weird stuff. It would take "A, Utah, USA" and turn it in to "Salt Lake City, Utah, USA". It took "Malmö, Sweden" and turned it in to Fosie, Sweden, which is like a suburb of Malmö maybe?

app/technovation/handle_geocoder_search.rb
```diff
@@ -81,7 +81,7 @@ def self.handle_one_result(db_record, geocoded, controller)
       status_code = :not_found
     end
 
-    return { results: [geocoded] }, status_code
+    return { results: [Geocoded.new(db_record)] }, status_code
```
stenington: `geocoded` here didn't necessarily agree with `db_record`, although I think that was largely because of the reverse geocoding, which has also been removed. But basically we were showing the user a different set of values than what we had in the DB and saying "Hey this is what we stored."

stenington:
@adrichey I believe this should be a good first step towards more predictable results coming out of the registration location selection screens.

stenington:
Looks like I have tests to fix!

stenington:
Okay, looks like I have more work to do because I removed the reverse geocoding. There are a couple of places that copy coordinates between models and rely on reverse geocoding to fill in the rest, but I think that's an unreliable strategy.

stenington:
@adrichey I see in your `location-form-carmen-rework` branch that the `it "reverse geocodes when coords change"` test was removed completely. Are you comfortable if I do that here?

adrichey:
@stenington Feel free to remove that test since your code also makes that change. Makes sense to me!

spec/requests/location_requests_spec.rb
```diff
@@ -69,15 +69,11 @@
     end
   end
 
-  %w{mentor student judge admin registration}.each do |scope|
```
stenington: This test previously passed for the registration scope because `HandleGeocoderSearch` wasn't returning data from the `db_record`. With that change, in this test `db_record` is actually a `NullSignupAttempt` and the test blows up.

app/controllers/registration/locations_controller.rb
adrichey: Yep, this appears to be the code that was knocking out the states. I tracked it down as well to this and walked Joe through that when this issue first started a couple of weeks ago, so we are on the same page here. Cool.

app/technovation/geocoding.rb
adrichey: Noted. Thanks for catching this!

stenington:
@adrichey As mentioned in Slack, since we're taking a strategy of many smaller changes I'm less concerned about QA being a clean slate. That said, since you're sort of the defacto release manager for now, it's up to you (and in general my personal preference is to release things in smaller batches).

Issue #1853, opened 2018-11-19
Confusing menu for selecting teams to RPEs


Allicolyer:
**Describe the bug**
The menu for selecting teams is a bit confusing, it does not update after clicking "save selected teams". 

**To Reproduce**
1. Log in as an RA
2. Create an event
3. Add some teams to your event
4. Click "Save Selected Teams"
5. Click "Add Teams"
6. Observe that this menu has not changed since you clicked "save selected teams"

**Expected behavior**
I would expect that clicking "save selected teams' would update the menu so that those teams no longer appear on the list and can't be "unselected". If you refresh the page this is what happens.

**Screenshots**
![selecting teams](https://user-images.githubusercontent.com/11083917/48725704-d2b44a80-ebe1-11e8-9bde-f1a07fcf381e.gif)

**Desktop (please complete the following information):**
 - OS: MacOS Mojave
 - Browser: Chrome

adrichey:
This is ready for testing. Clicking `Save selected teams` should save the teams you have selected and remove them from the search and select modal window for the next time you click `+ Add teams` for a more coherent experience.

Allicolyer:
Thanks! @adrichey this is working great.

PR #1887, opened 2018-12-12
Add event bus action for RPE save to alleviate confusing UX

adrichey:
Addresses #1853.


adrichey:
@stenington This is ready for review. Please note that this section of the code uses a great deal of Vue anti-patterns and some logic contained in models that really shouldn't be, so I did as little work as possible to create a solution and used a Vue Event Bus since it was already present in the application. I think at some point this code should be reworked to use Vuex properly, but that is a job for another day.

PR #1890, opened 2018-12-12
#1878 improve RA selection

stenington:
This changes RA selection logic to more often check for the closest RA where previously it would have selected at random.


app/models/account.rb
```diff
@@ -289,6 +289,20 @@ def assigned_teams
       Team.none
   end
 
+  def regional_ambassador
```
adrichey: Good call moving the RA logic out of the controller and on to the model!

PR #1891, opened 2018-12-13
#1878 QA fixes

stenington:
I compared a before/after of RA assignments and found some potential issues. More issues remain, hence this is WIP.

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -290,14 +290,14 @@ class Account < ActiveRecord::Base
   end
 
   def regional_ambassador
-    ra = find_regional_ambassador_by_state
+    ra = ::NullRegionalAmbassador.new
 
-    unless ra.present?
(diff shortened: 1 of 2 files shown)


stenington:
@adrichey This is a tiny bit more aggressive about not matching people with bad address details, which is ultimately a good thing. And I thought Katie's RA account matching was going to be a problem but she says it's appropriate.

Also, the rake task change here produces a slightly more useful report.

stenington:
Oh, let me update the rake task to include mentors.

stenington:
Okay, the task runs for mentors now as well. I also added two warn statements... I'm not sure if the location selection process is letting bad data through or if the bad locations I'm seeing are there for historical reasons, but I wanted to add some logging and watch Papertrail a bit to see how often these messages crop up to start to understand the problem a bit better.

Issue #1895, opened 2018-12-17
Change EventBrite URL label text


Allicolyer:
Change the label for the Eventbrite URL field to just say "Event URL (optional)"

adrichey:
I have put in a pull request for Mike to review. I went ahead and changed our database variables to reflect this change as well, so I want his eyes on my work before I merge that into QA since it is the first time I am doing this sort of thing. :slightly_smiling_face:

adrichey:
This is now ready for testing!

Allicolyer:
Looks great! Thanks @adrichey!

PR #1906, opened 2018-12-18
Add migration to change eventbrite_link to event_link

adrichey:
Addresses #1895. It is also my first migration in Rails, so I am adding @stenington to review this work.


stenington:
Yeah this looks appropriate to me. This probably doesn't need stating, but make sure you have some RPEs with links on QA before merging, and then check that the data carries through the migration after deployment. (I've forgotten to prepare data beforehand before. 😑 )

adrichey:
Thanks @stenington. I will make sure to have a couple of events set up before hand.

Issue #1892, opened 2018-12-13
Have RegisterToCurrentSeasonJob do location checks


stenington:
**Is your feature request related to a problem? Please describe.**
We have users with bad location data, and it's unclear where they're coming from. If they're returning users, it would be nice to have some notice of how bad the problem is.

**Describe the solution you'd like**
RegisterToCurrentSeasonJob could log if a user returns with location data we know is likely to lead to problems with RA assignment, RPEs, etc. I'd start with a simple `valid_address?` and maybe `valid_coordinates?` check. I haven't done any investigation into whether coordinates and text addresses are out of sync in the database yet, but maybe we should consider Geocoding returning users as well, and noting changes.

adrichey:
@stenington I am going to start in on this work if you are too busy to take it on. Do you want me to simply submit a `Rails.logger.warn` call if the location is not a `valid_address?`? Or should we store this in the database somewhere? I know that we have done a mix of both practices in the past, so I want to make sure that I am going about this the right way. I feel like the logger might be enough, but only if that gets reported to Airbrake of course.

stenington:
Yeah, I was thinking `Rails.logger.warn` would be good. I don't know if warning log messages go to Airbrake, but we can set up alerts in Papertrail to catch them there and put them in #dev-robots or something

adrichey:
Awesome. I think I will use Airbrake for now to track these. I will try to get a pull request in to you soon.

adrichey:
Since this is a developer testable ticket and I have tested the Airbrake notification, I am closing the ticket.

PR #1923, opened 2019-01-09
Have RegisterToCurrentSeasonJob log address/coordinates errors

adrichey:
Log error to Airbrake when RegisterToCurrentSeasonJob encounters an invalid address or coordinates.

Addresses #1892.


adrichey:
@stenington Can you review this since it is your reported issue? I just want to make sure this reporting is adequate.

Issue #1951, opened 2019-01-26
CSV Column for Scores that allows admins to see score discrepancies


Allicolyer:
**Description**
Problem: Every year during judging, staff spends time calculating the range of scores for all submissions in a separate spreadsheet. You can see that work [here](https://docs.google.com/spreadsheets/d/1-CYXJEJwwIEW6WwYC-VrL_e78WMVKGFhagLPKXrnibw/edit#gid=824758563)

The purpose is to find submissions that have a big difference (more than 35 points) between their highest and lowest score, in other words, have a range that is greater than 35 points. This helps us find outliers in the scoring process and judges who might be scoring unfairly.

**Solution**
It would save us a lot of time if we could add a column to the score CSV report that already does this work for us. We can call this column "range".

**Describe alternatives you've considered**
I think it makes sense to calculate the range, or the difference between the highest and the lowest. 

However, with large ranges you don't always find an outlier, sometimes you just see a large spread of scores. We can also consider specifically looking for outliers, by calculating if one score is more than 17 points away from the average score. However, I think it's simpler and just as effective to look at the range.

**Acceptance Criteria**

1. As an admin
2. I can add a column to the CSV report called range
3. This column lists the difference between the highest and lowest score for each submission

adrichey:
@ophelieh and @kmorton2017 I have a questions on this. I looked at the CSV export to see what the column data should look like. Can we elaborate a bit more on what the column, or potentially columns, should look like? Here are things I need insight on:
1. Should there be two columns for range, official and unofficial range? Or should official and unofficial be lumped into one range?
2. Additionally, we will need to display the range on multiple lines. Because we show each individual score for every team, we will be displaying the range on multiple lines. For example, if a team received four scores (73, 67, 9, and 7), the range would be 64 and each of those four scores in the CSV would show 64 as the range. I want to confirm that this is all right since we cannot reorganize this without changing the entire CSV structure or creating a new type of CSV export.

I am marking this as blocked until I get feedback on this.

Allicolyer:
Hey @adrichey!

1. Let's only count official scores in the range 
2. It's okay for it to appear on multiple lines 

As long as we all know how this is working I think it will be okay.

adrichey:
Sounds good! Thanks @Allicolyer. I will mark this as unblocked.

adrichey:
This is now ready for testing. Thanks @stenington for knocking this out!

ophelieh:
[...]

PR #1968, opened 2019-02-06
#1951 Adding score range to admin

stenington:
This adds score ranges for official quarterfinal and semifinal scores as optional columns in the admin scores page.

The ranges are pre-calculated and stored like averages are, on the assumption that calculating it on the fly would be too slow. `rake scores:set_score_summaries_of_completed` can be used to force a recalculation of averages and ranges if needed.


spec/models/team_submission_spec.rb
```diff
@@ -178,48 +178,229 @@
     end
   end
 
-  it "only averages scores that count" do
-    SeasonToggles.set_judging_round(:qf)
+  describe "quarterfinals scores" do
```
stenington: I intend to add tests for semifinals too.

Issue #1950, opened 2019-01-25
Change events so that it can only be one division 


Allicolyer:
**Is your feature request related to a problem? Please describe.**
Problem: You can't make only one division unofficial for an event. This means that an RA must create two events if they do not get at least 10 teams for each division. This leads to staff time asking to RAs to delete and remake events, the re-inviting the team to the events, ect.

**Describe the solution you'd like**
Change the checkboxes for divisions to radio buttons

**Acceptance Criteria**
1. Make sure events are turned on as an admin
2. Log in as an ra
3. Create an event
4. Confirm that the event can only be junior or senior, not both.

adrichey:
This is now ready for testing. Please note that this change does not convert RPEs already in the system to one division. The reason why this was done is so that live events already created in the system are still able to function as before. If you edit an event that has both junior and senior divisions it will show them as neither being selected when the edit form loads because of this compromise. The plus side to this bit of weirdness is that it prevents people from hitting save on that form so they don't accidentally convert the event to a single division if they are on auto-pilot. We may want to convert all RPEs over to being a single division on the back-end at a later date. Comments were left in the code to this effect.

ophelieh:
@nickiredescent Could you test this today?

ophelieh:
Question - what happens for events that have already been created and have both divisions? (I just checked on the platform and we currently have 2 with 2 divisions)

ophelieh:
Otherwise, it works as expected - I can only select one division. 

_Side note - I was able to create an event that starts at 14:00 (2pm) and ends at 7:00. Not really important but wanted to bring it up just in case_

adrichey:
@ophelieh they will still be set for two divisions. This only affects new events.

As for the other issue, we have noticed that and talked about validating the inputs of the start and end time in the past, but it has never been a high priority. We may want to make a note of that somewhere and maybe address it later if we have some quiet time.

ophelieh:
Ok. Good to know! 

And should I add it to the icebox maybe?

adrichey:
It might be a good one for the icebox. Sure!

nickiredescent:
Also confirming that RA event creation can only be done with a single division selected.

PR #1969, opened 2019-02-07
Convert checkboxes over to radio buttons for selecting a division for RPEs.

adrichey:
Addresses #1950.

I did a front-end only fix for this as a back-end fix would be much more involved. If this isn't thorough enough, which I would totally understand, we can talk about a more involved fix. I feel like this is the easiest way to implement this feature as it doesn't break events already in the system, and it allows us to have single division events going forward.

### DIFF
--- a/app/javascript/ra/events/EventForm.vue
+++ b/app/javascript/ra/events/EventForm.vue
@@ -27,8 +27,9 @@
 
           <label>
             <input
-              type="checkbox"
-              v-model="event.division_ids"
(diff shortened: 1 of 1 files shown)


adrichey:
@stenington I added a comment to the file with a TODO label for our reference on why this functionality is the way it is and to possibly have us remove the ability to add a second division between seasons. Since you already approved this previously and the new addition is just a comment, I will merge this into QA once it passes through CI.

Issue #1936, opened 2019-01-23
Judging Negative Sentiment Detector


Allicolyer:
**Is your feature request related to a problem? Please describe.**
The negative sentiment detector was very buggy and expensive last year. We introduced it to help with negative comments and bad words, but it caused a lot of problems. For instance, it was overall buggy and caused some confusion. It also cost us a lot in API costs.

**Some possible solutions**

1. Removing it completely: This will take some work. It also puts us back at square 1 for controlling bad comments. 
2. Only running the check with the judge presses the button. This will help remove the some of the costs and some of the bugs. There might be potential problems with confusion with the UI

**Questions**

1. Were curse words really ever a problem? I think the word counter really helps with comments and eliminating one word comments.

Alan is going to do some digging to figure out how much work it would be to remove it completely.  It sounds like program team wants to remove it.

adrichey:
@Allicolyer @ophelieh @kmorton2017 It is a pretty decent level of effort to get this done, but the code would be much more simple afterward if we were to remove the functionality, so I am okay with tackling this during this sprint. Additionally, tackling this **before** #1941 would make that development effort far easier.

ophelieh:
If you are comfortable with removing it, then I would say let's do it! 

I feel like it caused more issues than anything last year. It is true that we will have nothing to monitor negative feedback, but I don't know how much it actually did. 

If @kmorton2017 also agrees, I think we can move forward with this.

kmorton2017:
Yes move forward with removing the judge sentiment detector for detecting negative words.

adrichey:
Awesome. I will try to get to this once we work through what is currently in the To Do column. I think this is doable. @stenington this is a front-end ticket that I am rather familiar with, so I will tackle this one.

adrichey:
Something to note once this work has been completed is that we will no longer be using Algorithmia for TC after this code is removed. @stenington are you using that service for CM? If not, we may want to delete that account, or at the very least, regenerate the current API key.

stenington:
@adrichey Nope, CM doesn't use Algorithmia.

adrichey:
This is now ready for testing.

ophelieh:
@nickiredescent Could you test this today as well? Basically, as a judge, you should be able to enter any word and nothing happens. 

Thanks!

nickiredescent:
[...]

PR #1970, opened 2019-02-07
Remove sentiment detection in judge scores

adrichey:
Addresses #1936.

@stenington this also has a bug fix that was preventing the review step from recalculating if the user didn't refresh the page. You can see that in `ReviewScore.vue`.


adrichey:
@stenington Also note that I did not create a migration to remove the unused columns/data from the database as there might be a time when we put this back in and improve it. If you think I should write that migration, let me know and I will knock it out.

adrichey:
Thanks for the review @stenington! Really, I mainly tagged you to weigh in on the column removal question. I will go ahead and write migrations for that and let you know when this is ready for review again.

adrichey:
@stenington this is once again ready for review with the migration you had asked for. I put in column options details into the migration so we can revert this back if that time ever comes.

PR #1972, opened 2019-02-07
Move JS process.env variables to .env file

adrichey:
Thanks to a Google deep dive leading me back to the Webpacker documentation, I have extended our `.env` functionality for the front-end. This will allow us to pull any values we need for the front-end API calls to be stored in the `.env` file instead of plaintext.

Source: https://github.com/rails/webpacker/blob/master/docs/env.md


adrichey:
@stenington correct. These will only be injected at the time of compilation and should only pull in those variables explicitly called out in the code. You cannot access the variables using the console in any way.

Issue #1946, opened 2019-01-25
Virtual Judging will end before RPEs this year


Allicolyer:
**Is your feature request related to a problem? Please describe.**
Program staff wants the RPE timeline to be longer than the virtual judging timeline this year. the season toggles only allow quarterfinals judging to be on or off. Additionally program staff wants virtual judging to happen in the beginning of quarterfinals, not the end. 

Virtual judges will need to stop judging after 3 weeks, but nothing on the platform will stop them from continuing to judge anyway. This will cause major headaches because semifinalists will already be selected ect.

**Some solutions**

1. Having a separate season toggle for virtual judging versus RPEs (very high technical lift)
2. Putting messaging on the platform telling people that judging is over
3. Doing "database surgery" to remove scores that were submitted after virtual judging cut off. 
4. Make both judging periods 4 weeks long - April 27-May 27th 

Right now we are thinking of options 2 and 3 together. 

**Info we need*
@kmorton2017 and @ophelieh can you follow up with us and answer these two questions.

1. What will the messaging say. Please give us exactly what you want it to say and where to display it. 
2. What is the date for the cut off.

ophelieh:
Hi @Allicolyer @adrichey We actually discussed changing the dates so both open and close at the same time. I need to double check with RAs to make sure they haven't scheduled events after the 27th but otherwise we would have virtual and live judging open on April 27th and close on May 27th. 

I will keep you posted!

ophelieh:
Hi all! Based on my conversation with RAs, we will have to go with option 3 & 4. 

Draft Text: "Virtual judging is now closed! **Any score submitted after today (May 27th) will be removed.** Thank you for your help." --- this should be on the landing page when they log in and on the page for calling new submissions. 

What do you think @kmorton2017? Do we want to keep the virtual judging to May 27th, or bring it back to the 20th?

Allicolyer:
@ophelieh Monitoring the virtual judging is a lot of work -- My recommendation is to go with option 2 and keeping the original dates.

adrichey:
@kmorton2017 It sounds like we are ready to move forward with this. Can you take a look at the draft text @ophelieh put together and put your stamp on this so I can get started soon?

Allicolyer:
Since we're doing the database surgery option, it might make sense to make another card for that to remind ourselves when the time comes.

kmorton2017:
[...]

PR #1974, opened 2019-02-08
#1946 Ending virtual qfs before live qfs

stenington:
Opening this for discussion.

This PR introduces a `judging_enabled_for?` companion to `judging_enabled?` that takes into account both the quarterfinals setting in the admin **and** the current date compared to the appropriate quarterfinal ending date (virtual or live). 

I wouldn't call it a totally "clean" solution for the same reasons discussed elsewhere, but until we iron out an overall approach to `SeasonToggles` vs. dates it may be enough to get us through this season?

The way this is intended to work is that we'd turn on quarterfinals at the quarterfinal start date, then both virtual and live judging will be available until the virtual quarterfinals end date. At that point virtual judges will no longer see scoring controls on their dashboards, but live judges still will. Then when live judging ends, we'd turn off quarterfinals in the admin. Once we're ready for semifinals, we should be able to turn that on through the admin and all judges should see the same virtual judging controls. If we needed to extend virtual judging, we'd do that by changing the date.

One thing I could do is screenshot out that whole flow and share that with the program team for a 👍 if we're not ready for this to land on QA.


adrichey:
@stenington I think this sounds like a good approach and the logic makes sense to me. I think having QA buy-in might be a good way to go on this one since it directly impacts the way they handle their season. That way they can properly put out any fires that come up. We will also want to put the helper text somewhere that they had outlined on #1946 so that people know when virtual judging will end if it is not already out there somewhere. I apologize if it is and I didn't see it, I am kind of all over the place this afternoon so I only focused on the code in this PR.

stenington:
Okay, so let me put together a series of screenshots next week that show the basic flow through the quarterfinals for live and virtual judges, and I'll post those in #1946 for approval. If changes get requested (e.g. better messaging of the deadlines) I'll add that to this PR.

stenington:
Just noting here: I haven't put together screenshots because I think the ongoing exploratory QA will give us the feedback needed here. At least I hope so.

stenington:
Even though we're now talking about virtual and live QFs ending on the same date, I think we should consider accepting this PR. I think it helps clarify some internal logic around judging, and enforces the cutoff more thoroughly.

stenington:
Okay @adrichey this is ready for review

stenington:
I'm closing this PR out, and will issue a new one if needed.

Issue #1939, opened 2019-01-23
Students using Thunkable Classic can't upload source code


Allicolyer:
**Is your feature request related to a problem? Please describe.**
When you select Thunkable as your development platform, our site asks you to give your project URL. Thunkable classic does not provide URLs, only Thunkable X provides URLs

**Solutions**
- Ask them to select App Inventor for their development platform and provide help text (this will affect who is using App inventor versus Thunkable). How important is this data to the program team? Is it okay if a few numbers are off?
- Have development platform option for Thunkable Classic and have it behave the same way as the App Inventor option. 
- Give them the option to upload a file for Thunkable Option

adrichey:
@ophelieh and @kmorton2017 Alli outlined a couple of potential solutions for this ticket. My solution would be to approach this ticket from the second solution: "Have development platform option for Thunkable Classic and have it behave the same way as the App Inventor option." I think that makes the most sense from a data collection standpoint and isn't just a work around. If that works, I will get started on this ticket soon.

ophelieh:
Hi @adrichey i agree that this might be the best option. Thank you!

adrichey:
@ophelieh and @kmorton2017 I am going to get started on this now since this will block students from uploading their source code potentially.

ophelieh:
Totally fine!

adrichey:
This is now ready for testing. It functions similarly to the "Java or Android Studio" or "Swift or XCode" selection in that you can choose Thunkable Classic and then upload your code on the Source Code Upload step. Thunkable Classic uses .aia files, so our uploader is set up to handle them properly. Please reach out with any questions.

ophelieh:
Works great!

PR #1976, opened 2019-02-08
Add Thunkable Classic option to the team submission options for development platform

adrichey:
Addresses #1939.


adrichey:
@stenington This is ready for review. It is a one liner change, but I added tests and did thorough testing of the various places it would appear in the codebase for thoroughness sake.

PR #1979, opened 2019-02-11
#1951 QA fix: make scores admin use same datagrid for web and csv views

stenington:
This makes the scores CSV export look like columns you see on-screen.

### DIFF
--- a/app/controllers/admin/scores_controller.rb
+++ b/app/controllers/admin/scores_controller.rb
@@ -4,10 +4,7 @@ module Admin
 
     helper_method :grid_params
 
-    use_datagrid(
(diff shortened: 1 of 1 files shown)


stenington:
@adrichey Ophelie found in QA that the new column didn't show up on export. I looked into it and found that the web view vs. CSV export used completely different datagrids. This PR makes CSV export use the same datagrid as the web view, with the caveats that 

* I have no idea *why* the CSV export used a different datagrid and whether that contained crucial columns the web view doesn't
* and I don't know why it was narrowing down the params on an ajax request, but leaving it that way makes the export 500 and other admin controllers don't seem to do that

Issue #1956, opened 2019-01-30
Static Page that contains Thunkable's promo code


Allicolyer:
**Problem**

Thunkable wants a way to guarantee a person using their promo code has a Technovation account. Right now their promo code is in our curriculum, and our curriculum is an open web page. You can see that here: https://technovationchallenge.org/curriculum/coding-1/#act2

**Some Solutions**
1. Send the promo code in an email 
- this will create an insane headache when people cannot find our welcome email
2. Put the promo on a static page on requires a log in to our Rails app. Put a link to the static page in place of the image of the promo code is is now. 
- I think this is the best option.
3. Put the promo code under the "helpful links" menu on people's dashboard.
- I think this will also be really hard for people to find and also doesn't really make sense.

@adrichey, I can discuss this with you tomorrow. Thunkable's developer, Paul, also offered to code up the static page himself and create a pull request on our app. I know you have mentioned that you are open to people doing that, I wasn't sure how serious you were though :)

Allicolyer:
1. On the submission page, if you choose Thunkable it will display the promo code
2. Static Page - figure out where this would live in the app.

Alan needs: Exact text that should be on the static page and where we want to link it

kmorton2017:
Static Page Text will be the following: 

1. Go to https://x.thunkable.com/event/5bc11dbb1c9d440000725286
2.  Log in with your gmail account. If you don’t have one, you should make one now.
3. Type in the code in this picture: 

This code will give you a free trial of Thunkable PRO. Thunkable PRO allows you to create private projects, that way you can keep your apps secret during the Technovation Challenge. If you decide not to purchase Thunkable PRO, no worries, just please note that your apps will become public after the competition.

![uploads/99d218ec-1689-40e7-82d5-7fd60e711217/image.png](https://s3-us-west-2.amazonaws.com/prod.huboard.com/uploads%2F99d218ec-1689-40e7-82d5-7fd60e711217%2Fimage.png)

@adrichey let me know if you would like me to provide the word press text changes here

adrichey:
The proposed solution for this is to create a static information page within the platform with the copy Katie referenced previously. This page will only be accessible if the user is signed in. On the WordPress site, we will remove the promo code and add a link to the static information page within the platform. When a user clicks this link, they will be prompted to log in if they have not already done so, and will then be redirected to the promo code information page.

[...]

PR #1980, opened 2019-02-11
#1956 Add thunkable promo page and a link on the development platform page

adrichey:
Addresses #1956.

What this doesn't address is the link on the WordPress side of things. That will need to happen with coordination with the WordPress team after this code has gone live.


adrichey:
@stenington, I believe this is ready for your review. Please note that I didn't use the `Authenticated` concern for the controller because the messaging to the user would always come back as being a little odd since since the scope_name is "application" for the NullAuth object.

PR #1986, opened 2019-02-15
Implement a team cap for RPEs

adrichey:
Addresses #1948.

This allows us to set a team cap for RPEs so that teams cannot join once the cap threshold is met. RAs can set their own caps on their events, and they cannot exceed the cap in their UI either.

### DIFF
--- a/app/controllers/mentor/regional_pitch_event_selections_controller.rb
+++ b/app/controllers/mentor/regional_pitch_event_selections_controller.rb
@@ -8,14 +8,17 @@ module Mentor
     def create
       event = RegionalPitchEvent.find(params[:event_id])
 
-      if event.at_team_capacity?
-        redirect_back fallback_location: mentor_regional_pitch_events_url,
-          error: t("controllers.student.regional_pitch_event_selections.create.event_full")
-      else
-        AddTeamToRegionalEvent.(event, current_team)
+      begin
 
+        AddTeamToRegionalEvent.(event, current_team)
         redirect_to mentor_regional_pitch_events_team_list_path,
           success: "#{current_team.name} is now attending #{event.name}"
+
+      rescue AddTeamToRegionalEvent::EventAtTeamCapacityError => ex
+
+        redirect_back fallback_location: mentor_regional_pitch_events_url,
+          error: t("controllers.student.regional_pitch_event_selections.create.event_full")
+
       end
     end
   end
--- a/app/controllers/mentor/regional_pitch_event_selections_controller.rb
+++ b/app/controllers/mentor/regional_pitch_event_selections_controller.rb
@@ -8,10 +8,15 @@ module Mentor
     def create
       event = RegionalPitchEvent.find(params[:event_id])
 
-      AddTeamToRegionalEvent.(event, current_team)
+      if event.at_team_capacity?
+        redirect_back fallback_location: mentor_regional_pitch_events_url,
+          error: t("controllers.student.regional_pitch_event_selections.create.event_full")
+      else
+        AddTeamToRegionalEvent.(event, current_team)
 
-      redirect_to mentor_regional_pitch_events_team_list_path,
-        success: "#{current_team.name} is now attending #{event.name}"
+        redirect_to mentor_regional_pitch_events_team_list_path,
+          success: "#{current_team.name} is now attending #{event.name}"
+      end
     end
   end
 end
--- a/app/controllers/student/regional_pitch_event_selections_controller.rb
+++ b/app/controllers/student/regional_pitch_event_selections_controller.rb
@@ -5,10 +5,15 @@ module Student
     def create
       event = RegionalPitchEvent.find(params.fetch(:event_id))
 
-      AddTeamToRegionalEvent.(event, current_team)
+      if event.at_team_capacity?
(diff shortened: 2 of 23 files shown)


adrichey:
@stenington This is ready for review. I am tagging you since this is a pretty big feature. I wouldn't worry too much about the front-end Vue code, but if you could look over the back-end stuff and related rspec, that would be appreciated.

app/controllers/mentor/regional_pitch_event_selections_controller.rb
```diff
@@ -8,10 +8,15 @@ def new
     def create
       event = RegionalPitchEvent.find(params[:event_id])
 
-      AddTeamToRegionalEvent.(event, current_team)
+      if event.at_team_capacity?
```
stenington: This is somewhat of a redundant check since `AddTeamToRegionalEvent` has it's own check. I don't necessarily see that as a problem, but there are 2 failure paths based on capacity: this check and the subsequent redirect, or the less likely (but I think still possible) scenario where this check passes but adding to the team actually fails due to race conditions, which I assume would result in a 500 since `AddTeamToRegionalEvent` raises an error?

app/javascript/ra/events/EventForm.vue
```diff
@@ -343,6 +371,10 @@
         form.append("regional_pitch_event[event_link]",
           vm.event.event_link);
 
+        if (this.showCapacity && Boolean(vm.event.capacity)) {
```
stenington: Won't this prevent me from unsetting a capacity if I change my mind as an RA?

spec/models/regional_pitch_event_spec.rb
```diff
@@ -34,6 +34,17 @@
     expect(rpe.reload.teams).to be_empty
   end
 
+  it "removes teams when trying to add and event is at team capacity" do
```
stenington: Technically it prevents the addition of the new team, right? It doesn't remove the existing teams, I think.

app/controllers/mentor/regional_pitch_event_selections_controller.rb
adrichey: The reason why I wanted this one here is so that the user sees some sort of feedback instead of an uh-oh page. I suppose this check is kind of unnecessary. I am open to leaving this or removing it. I leave that to you. If you would rather us fail hard with the `AddTeamToRegionalEvent` check, that is fine by me. I suppose this is a philosophical question then. Do we want to just fail hard in our code, or is feedback important with this sort of thing?

app/javascript/ra/events/EventForm.vue
adrichey: Oh man, yep. It will. I thought I had an else in there to set it to null, but I don't. Thanks for catching that.

spec/models/regional_pitch_event_spec.rb
adrichey: I just copied the wording from another similar test spec. I will change that. I like that better. I will do that when I fix the capacity issue on the front-end.

app/controllers/mentor/regional_pitch_event_selections_controller.rb
stenington: Right, I agree feedback is a good thing. Maybe do a try/catch catching the error that comes out of `AddTeamToRegionalEvent` and do the redirect there?

adrichey: @stenington, after looking at this more, I am not sure how to redirect within that `AddTeamToRegionalEvent` class. I tried catching the error, but once I get into the rescue block, I don't have access to the route URL variables or any of the variables that would allow me to figure out what scope I am currently logged in as (student versus mentor). All of this makes routing at that level difficult. If you know of a straightforward way do that, perhaps we could look at this together, or you could take a crack at it and push to this PR. Otherwise, I might leave this functionality as is.

app/javascript/ra/events/EventForm.vue
adrichey: This has been fixed and is ready for review again.

spec/models/regional_pitch_event_spec.rb
adrichey: This has been updated and is ready for review again.

app/controllers/student/regional_pitch_event_selections_controller.rb
```diff
@@ -5,10 +5,18 @@ class RegionalPitchEventSelectionsController < StudentController
...
-      AddTeamToRegionalEvent.(event, current_team)
+      begin
 
-      redirect_to [:student, current_team.selected_regional_pitch_event],
-        success: t("controllers.student.regional_pitch_event_selections.create.success")
+        AddTeamToRegionalEvent.(event, current_team)
+        redirect_to [:student, current_team.selected_regional_pitch_event],
+          success: t("controllers.student.regional_pitch_event_selections.create.success")
+
+      rescue AddTeamToRegionalEvent::EventAtTeamCapacityError => ex
```
adrichey: Awesome. This makes sense. I approached my non-working solution by putting the rescue in the `AddTeamToRegionalEvent` class. This makes sense to me and I like this solution. Thanks @stenington!

Issue #1985, opened 2019-02-15
Admin Events Page - Seeing the Wrong Number of teams


ophelieh:
**Describe the bug**
As an admin, when I look at an event, I can see the numbers of teams. However, when I click on that event, sometimes the number of teams is different than on the main page. 

**To Reproduce**
1. Go to 'Events'
2. Look at team count
3. Click on 'View' under Action
3. Look at number of teams 
4. See actual number of teams 

**Expected behavior**

**Screenshots**

![image](https://user-images.githubusercontent.com/32206421/52869117-902fda80-310a-11e9-86da-a11c0902cf28.png)

![image](https://user-images.githubusercontent.com/32206421/52869144-9d4cc980-310a-11e9-9490-63f37220fd1a.png)

![image](https://user-images.githubusercontent.com/32206421/52869172-accc1280-310a-11e9-8de0-6fa23dd93fc7.png)

stenington:
Ok, this should be fixed. QA suggestions:

1. Go to 'Events'
2. Look at team count
3. Click on 'View' under Action
4. Look at number of teams
5. Confirm that the two counts match
6. Remove a team
7. Confirm that the team count updates correctly
8. Add a team back
9. Confirm that the team count updates correctly

kmorton2017:
i wasn't able to do steps 8 and 9 with exploratory QA settings enabled (cannot add teams back with these settings) , but i was able to do 1-8 and confirm that the count updated

stenington:
RAs can add teams even when judging is enabled, can't they? I know teams or mentors can't add themselves, but I believe RAs can change their event teams and judges at any time.

kmorton2017:
hmmm it was telling me there were . no more teams left for me to add even tho there were definitely LA teams i could add to my event

stenington:
Oh, okay, let me look in to that @kmorton2017. I was working locally, not on staging, so let me see if there's an explanation for that.

stenington:
@kmorton2017 I don't see anywhere obvious that the season toggles would be preventing you from adding teams back. Which event and which team was it?

ophelieh:
This works great! Thank you! :)

PR #1991, opened 2019-02-22
#1985 try another approach to caching team counts on events

stenington:
Rails seems to have the concept of a counter cache, but it doesn't work with `has_and_belongs_to_many` so I think incrementing/decrementing the count through callbacks here was essentially trying to roll our own counter cache. 

I believe it was getting out of sync because it was only set up to fire from one side of the equation. If you did something like `event.teams.delete(team)` it would have worked (I think) but the app is coming at it the other way [on this line](https://github.com/Iridescent-CM/technovation-app/blob/qa/app/technovation/invalidate_existing_judge_data.rb#L20), where `participant` is sometimes a team. 

I thought about trying to turn that statement around, but the problem is that `participant` is also sometimes a judge, so that makes it more complicated.

Incidentally, the datagrid also includes a judge count [that's calculated, not cached](https://github.com/Iridescent-CM/technovation-app/blob/qa/app/data_grids/events_grid.rb#L42) but it's an optional field, so I'm less worried about what happens if all the extra queries become to slow.

So my proposed solution in this PR does two things:
* it adds a recalculation of `teams_count` from the `Team` side of things so that previously referenced line should cause a recalculation
* it also switches from using `increment_counter` and `decrement_counter` to `update_attributes` and querying for the count, which I think may be slightly slower since it involves an extra query, but slightly more robust

### DIFF
--- a/app/models/regional_pitch_event.rb
+++ b/app/models/regional_pitch_event.rb
@@ -190,17 +190,8 @@ class RegionalPitchEvent < ActiveRecord::Base
     "#{FriendlyCountry.(ambassador.account)} - #{name}"
   end
 
-  def update_teams_count
+  def update_teams_count(team = nil)
     update_attributes(teams_count: teams.count)
-  end
-
-  private
-  def inc_teams_count(team)
-    RegionalPitchEvent.increment_counter('teams_count', id)
-  end
-
-  def dec_teams_count(team)
-    RegionalPitchEvent.decrement_counter('teams_count', id)
   end
 end
 
--- a/app/models/regional_pitch_event.rb
+++ b/app/models/regional_pitch_event.rb
@@ -48,8 +48,8 @@ class RegionalPitchEvent < ActiveRecord::Base
 
   has_and_belongs_to_many :teams,
     -> { joins(:team_submissions) },
-    before_add: :inc_teams_count,
-    before_remove: :dec_teams_count
+    before_add: :update_teams_count,
+    before_remove: :update_teams_count
 
   has_many :team_submissions, through: :teams
 
@@ -188,6 +188,10 @@ class RegionalPitchEvent < ActiveRecord::Base
 
   def name_with_friendly_country_prefix
     "#{FriendlyCountry.(ambassador.account)} - #{name}"
+  end
+
+  def update_teams_count
+    update_attributes(teams_count: teams.count)
   end
 
   private
--- a/app/models/regional_pitch_event.rb
+++ b/app/models/regional_pitch_event.rb
@@ -48,8 +48,8 @@ class RegionalPitchEvent < ActiveRecord::Base
 
   has_and_belongs_to_many :teams,
     -> { joins(:team_submissions) },
-    before_add: :update_teams_count,
-    before_remove: :update_teams_count
+    after_add: :update_teams_count,
+    after_remove: :update_teams_count
 
   has_many :team_submissions, through: :teams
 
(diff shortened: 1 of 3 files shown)


adrichey:
This work makes sense to me, but it is throwing errors in the spec:
```
Failures:

  1) Student team submissions See live pitch event submission pieces on TOC
     Failure/Error:
       def update_teams_count
         update_attributes(teams_count: teams.count)
       end
     
     ArgumentError:
       wrong number of arguments (given 1, expected 0)
     # ./app/models/regional_pitch_event.rb:193:in `update_teams_count'
     # ./spec/features/student/submission_spec.rb:185:in `block (2 levels) in <main>'

Finished in 4 minutes 24 seconds (files took 2.84 seconds to load)
267 examples, 1 failure, 3 pending

Failed examples:

rspec ./spec/features/student/submission_spec.rb:149 # Student team submissions See live pitch event submission pieces on TOC
```
Not really sure why it is seeing a different number of arguments there, to be honest. So I think we will want to dig into that a bit. I will look around. If you know what might be the cause, I would really appreciate that knowledge transfer. 🙂

stenington:
Yep, just saw that failure. Turns out the callback for `before_add` and I assume `before_remove` send in the item being added/removed, so I updated the method signature to take an optional parameter.

adrichey:
Ah, it looks like `def update_teams_count` needs to become `def update_teams_count(team)` as it is passed the team that is being added or removed. So I think that should get the tests passing again.

stenington:
I should wait for tests to pass before I tag you for review, sorry about that @adrichey

app/models/regional_pitch_event.rb
```diff
@@ -190,18 +190,9 @@ def name_with_friendly_country_prefix
     "#{FriendlyCountry.(ambassador.account)} - #{name}"
   end
 
-  def update_teams_count
+  def update_teams_count(team = nil)
     update_attributes(teams_count: teams.count)
```
adrichey: Cool, this is what I found too. I will approve it as soon as CircleCI is green. Thanks Mike!

stenington:
Failed again, this time the tests caught a logical bug. Querying for the count before adding or removing results in the wrong count. 😬

app/technovation/add_team_to_regional_event.rb
```diff
@@ -1,23 +1,23 @@
...
-      else
-        event.teams << team
-
-        SendPitchEventRSVPNotifications.perform_later(
-          team.id,
-          joining_event_id: event.id
-        )
-      end
-    else
+    if event.divisions.exclude?(team.division)
```
stenington: Changed the check here from `include?` to `exclude?` and brought the `raise` up so this check can be at the same indentation level as the team capacity check and reduce the nesting overall.

app/technovation/add_team_to_regional_event.rb
```diff
@@ -1,23 +1,23 @@
...
       raise IncompatibleDivisionError,
         "This team is not in the correct division for this event"
     end
+
+    if event.at_team_capacity?
+      raise EventAtTeamCapacityError,
+        "This team cannot attend the event as it is currently full"
+    end
+
+    InvalidateExistingJudgeData.(team)
```
stenington: I moved this after the capacity check so that if the team is denied entry because of capacity, they don't have their scores emptied out as well.

adrichey: Good call.

PR #1992, opened 2019-02-22
Fix one type of sporadic test failure

stenington:
Sometimes admin tests fail because they assume a user has shown up on the first page of results when they haven't. This should fix that problem.


stenington:
It looks like other tests have the same problem, so I'm going to extract an admin helper of some sort that they can all use, I think.

Issue #1944, opened 2019-01-25
RAs should not see scores from virtual judging until scores are released 


Allicolyer:
**Is your feature request related to a problem? Please describe.**
RAs can see scores from virtual judging in real time, before we have formally released scores to the public. This caused many customer service problems last year. RAs either complain to us about what they are seeing or tell the girls what their scores are.

- RAs should see scores from events
- RAs should *not* see virtual scores
- RAs should see virtual scores once scores are released

**Possible Solutions**

- Alan can link this to the season toggle. RAs will only see virtual scores once this season toggle is on.
- Link it to the scores and certificates toggle
- RAs never see virtual scores -- why would they ever need this anyway? (not recommended)
- Leave it how it is and improve messaging when RAs complain. Tell them we are working on score issues and will get around to it soon. 

**Other things to consider**
- Filters on RA score page might need to change
- Seeing scores from past seasons.

kmorton2017:
update from meeting 02/08: do not prioritize this

stenington:
I know this was de-prioritized, but I'm going to see if I can do it while waiting for the results of judging exploratory QA.

stenington:
I believe these are the two places, along with the CSV download, that RAs would see scores: the scores page and score details page.

![technovation-qa herokuapp com_regional_ambassador_scores](https://user-images.githubusercontent.com/55319/53362687-b7f02100-3908-11e9-9b4c-4fd3b1d268a1.png)

![technovation-qa herokuapp com_regional_ambassador_score_details_59](https://user-images.githubusercontent.com/55319/53362686-b7f02100-3908-11e9-80a9-05ae35c15079.png)

What I can look in to doing is for RAs only, filtering down the teams shown and hiding various parts of pages or columns until scores are made accessible by changing the admin setting. While scores are not generally accessible, the scores page would:
* only show teams attending official RPEs
* not show any semifinals columns

and the score details page would:
* not show sections related to semi-finals

After scores are made available, the RAs would essentially see what admins see (but limited to teams in their region of course).

stenington:
@adrichey can you give me a gut-check on whether the proposed solution above would fix this issue?

adrichey:
I believe that sounds like a good solution, Mike. That was essentially what I had proposed with Alli when this ticket was created. That we basically lock things down for RAs based on whether the Score & Certs Available toggle was turned on.

I am going to tag @ophelieh and @kmorton2017 for one of them to look at the solution as well.

[...]

PR #1998, opened 2019-02-26
#1944 Only show RAs virtual qf and sf scores after score display toggle is set

stenington:
I think this will prevent RAs from seeing quarterfinals virtual scores and any semifinals scores until scores are displayed by setting the admin setting.

### DIFF
--- a/app/controllers/regional_ambassador/scores_controller.rb
+++ b/app/controllers/regional_ambassador/scores_controller.rb
@@ -9,14 +9,17 @@ module RegionalAmbassador
 
         if not scored_submissions_grid[:by_event].blank?
           scope.page(params[:page])
-        else
+        elsif SeasonToggles.display_scores?
           scope.in_region(user).page(params[:page])
+        else
+          scope.live.in_region(user).page(params[:page])
         end
       },
 
       csv_scope: "->(scope, user, params) { " +
           "if not params[:by_event].blank?; scope; " +
-          "else; scope.in_region(user); end " +
+          "elsif SeasonToggles.display_scores?; scope.in_region(user); " +
+          "else; scope.live.in_region(user); end " +
         "}"
 
     def show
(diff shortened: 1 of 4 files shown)


app/views/regional_ambassador/score_details/show.en.html.erb
```diff
@@ -8,9 +8,11 @@
     <dt>Event:</dt>
     <dd><%= @submission.team.event.name %></dd>
 
-    <% if @submission.semifinalist? %>
-      <dt>Semifinals average:</dt>
-      <dd><%= @submission.semifinals_average_score %></dd>
+    <% if current_account.is_admin? or SeasonToggles.display_scores? %>
```
stenington: Admins and RAs see the same view, so I just check if the user is an admin here. I'm not sure if it's better to split out the two templates though, I haven't seen a lot of other screens that share the template between the two.

app/controllers/regional_ambassador/scores_controller.rb
```diff
@@ -44,7 +47,7 @@ def grid_params
           end
         ),
         current_account: current_account,
-        round: params.fetch(:round) { round },
+        round: round,
```
stenington: There was actually a bug for RAs where the round parameter was never carried through no matter how they set it. It always defaulted to the current round.

This solution presents them with both quarterfinals and semifinals as options, but ignores their choice if scores aren't set to display in the season toggles. It's a crappy user experience, but slightly better than the broken experience they would have had before. That said, I can look into maybe not even showing the options that they won't be able to pick if you think it's worth it.

stenington:
I need to fix tests and try to write some new ones, maybe.

stenington:
@adrichey This is ready for review, I think. There are a couple places where I took the easy way out to see if it would pass code review, but I'm fine with you calling me on that. 😁

app/views/regional_ambassador/score_details/show.en.html.erb
adrichey: I am okay with this. I think the admin check is fine. I think the alternative doesn't really give us much here.

app/controllers/regional_ambassador/scores_controller.rb
adrichey: For now, I think this is good. Thanks for catching this.

PR #2002, opened 2019-02-28
Use Timecop to set appropriate date for tests failing between rounds

stenington:
@adrichey I think this fixes all the tests that would otherwise fail between rounds. I didn't do anything particularly clever, just used Timecop to specify the day before QFs starts.


stenington:
I'm running the test suite locally one last time with my dates set so it's between judging rounds since on CI 5/20 - 6/1 would be between rounds.

stenington:
Ok, they pass locally, now waiting for them to pass on CI.

Issue #1995, opened 2019-02-25
Improve links for scored submissions in various UI spots


adrichey:
**Is your feature request related to a problem? Please describe.**
Right now, users can have trouble navigating between team, score, and submission in the UI when trying to look at scores data. We would like to have links to help navigate back and forth whenever a user visits any of those pages.

Basically for all scenarios below, we need links to the team when team name is displayed, and submission links when the app name is present.

**Screenshots where we need links added**
Section 1
> Log in as a mentor and visit Scores & Feedback on your dashboard.
> ![screen shot 2019-02-25 at 12 03 50 pm 2](https://user-images.githubusercontent.com/1534554/53358749-c46b7e00-38f6-11e9-88a3-19d9b98bbd1b.png)
> The submission name should be clickable and link to the non-editable submission app page.

Section 2
> Log in as a mentor and visit Scores & Feedback on your dashboard. When viewing a score in the Scores & Feedback section, click View Details.
> ![screen shot 2019-02-25 at 12 18 39 pm 2](https://user-images.githubusercontent.com/1534554/53359184-d0a40b00-38f7-11e9-9138-c5d70f3bb54a.png)
> On this page, only the submission link is clickable. The team name should link to the team.

stenington:
The proposed PR also links the team name when viewing a score as an admin:

<img width="958" alt="screen shot 2019-03-01 at 2 56 52 pm" src="https://user-images.githubusercontent.com/55319/53663057-a1143c00-3c32-11e9-9320-419e2bed8b98.png">

stenington:
Looks like judge and RA pages now have a link too.

stenington:
Deleted my previous comments about other user types seeing this link, now it should only be mentors.

stenington:
Acceptance criteria:

1. With **Scores & Certificates > Scores & Certificates accessible** turned on in the admin
2. Log in as a mentor
3. Visit Scores & Feedback on your dashboard
4. Confirm that the submission name (*Go Run* in the screenshot in the initial description) links to the non-editable submission page.
5. Visit Scores & Feedback on your dashboard again
6. Click View Details
7. Confirm that the team name (*We <3 Coding* in the second screenshot above) links to the team page

ophelieh:
Hi @nickiredescent  and @KaiLokken could you both test this today or tomorrow? 

Thank you!

nickiredescent:
Tested. Can confirm that both scenarios have appropriate links.

[...]

PR #2013, opened 2019-03-01
#1995 Adding links from score views

stenington:
This adds the links requested in #1995. Incidentally, it adds a link for admins too, because one of those mentor views uses the admin template, but I think that's okay.

screen shot 2019-03-01 at 2 56 52 pm


stenington:
Just found a bug, judges don't have a path to view teams so this breaks.

stenington:
Okay, now the link is only there for mentors.

Issue #2009, opened 2019-03-01
Dashboard for admin/RAs does not work as expected


ophelieh:
**Describe the bug**
When clicking on the student or mentor pie, admin/RA is brought to an error page. 

**To Reproduce**
1. As an admin go to Dashboard
2. Click on any of the information in the student or mentor pies

**Expected behavior**
When clicking on either of the pies, I should see the appropriate information. For example students with parental consent on file. 

Note - this error might be caused by the filters we changed on the participants' page. 

**Screenshots**
![image](https://user-images.githubusercontent.com/32206421/53651961-66e47380-3c0d-11e9-98a3-439cfb585e91.png)

![image](https://user-images.githubusercontent.com/32206421/53651840-24bb3200-3c0d-11e9-9270-0cd6c33dd0ce.png)

ophelieh:
Hi @stenington! Could you look into this when you get a chance?

stenington:
Sure @ophelieh, I'll start on it today.

stenington:
@ophelieh It looks like ticket #1915 asked for the parental consent filter to be removed, which is part of the reason why the graph links are now broken. So I have some options on how to solve this:

* bring back the parental consent filter, including adding it back to the participants page
* bring back the parental consent filter, but try not to show it on the participants page (it only gets used when linking from the graph)
* update the graph to show onboarded vs. not onboarded, since that's what #1915 was asking to replace those filters with

What would be your preference for me to look in to?

stenington:
Oh, since mentors are broken too I guess the options are basically the same as above, but with the background check and consent waiver filters as well as parental consent.

ophelieh:
Hi @stenington I think option 3 (_update the graph to show onboarded vs. not onboarded, since that's what #1915 was asking to replace those filters with_) might make the most sense, especially for consistency. That might also make the mentor graph a little cleaner and if people want more detail they can use the "more columns" on the participants' page.

stenington:
@ophelieh or @kmorton2017 As I'm doing this ticket and working on the mentor part, I'm finding some conflicting definitions of what it means for a mentor to be onboarded. Can you help clarify? 

One source says they have to have:
* email confirmed
* training complete or not required
* consent signed
* background check complete, if US
* their bio isn't blank
* their type isn't blank (Industry professional, Educator, Parent, Past Technovation student)

Another source just says:
* email confirmed
* training complete or not required
* background check complete, if US

And a third says:
* email confirmed
[...]

Issue #1915, opened 2019-01-02
Replace mentor filters with on-boarded yes or no filter


Allicolyer:
The current filters on the participant page are as follows:
- PARENTAL CONSENT (STUDENTS ONLY)
- CONSENT WAIVER (MENTORS ONLY)
- BACKGROUND CHECK (MENTORS ONLY)
- MENTOR TRAINING (MENTORS ONLY)

Replace these filters with a more universal one:
- ONBOARDED (MENTORS AND STUDENTS ONLY)
This filter should include "yes and no" as the dropdown options. This filter should also work well with the filter that selects for different participant types. This will allow admins to more easily search for onboarded students and onboarded mentors.

To replace the other filters, we add more granular CSV columns. This is outlined in another card.

adrichey:
I ended up having to create two separate filters for this. Because we are using a general query for the various accounts in our system, coding one catch-all filter gets pretty convoluted in the codebase. And since we already had a separate "Onboarded Judges" I felt the two "Onboarded Students" and "Onboarded Mentors" filters were the way to go. Additionally, we have a couple of student accounts with funky data (they show up as onboarded, but aren't finished onboarding), so keep that in mind while testing. You can view a list of those accounts here:

https://drive.google.com/drive/folders/1fReh6xoIwEjnR1bfKkAN7bH62Vxmq4Jg?usp=sharing

adrichey:
This is now ready for testing!

adrichey:
I just saw those errors come through on the server @kmorton2017. Looks like there is an issue when applying the new filter. I will need to work on this. I am wrapping up for the day, so I will pull this back into the To Do column. Thanks!

adrichey:
@kmorton2017 Can you please test using the following filters:

Onboarded Mentors: No, still onboarding
More Columns: Mentor training (and whatever other columns you want)
Profile Type: Mentors

Onboarded Students: No, still onboarding
Profile Type: Students

I think the problem is that you aren't selecting a profile type first before applying the filter and it is bombing as a result. I was able to get the filters working locally by adding the proper profile type depending on which onboarding filter I was using. Thanks!

PR #2014, opened 2019-03-04
#2009 Change admin/RA dashboard graphs to show onboarded or not

stenington:
Fix for #2009, follows in the footsteps of #1915 to simplify the student/mentor onboarding pie charts.


app/models/data_analyses/onboarding_mentors_data_analysis.rb
```diff
@@ -12,10 +12,27 @@ def init_data
           .in_region(user)
       end
 
-      @cleared_mentors = @mentors.bg_check_clear.consent_signed
-      @signed_consent_mentors = @mentors.bg_check_submitted.consent_signed
-      @cleared_bg_check_mentors = @mentors.bg_check_clear.consent_not_signed
-      @neither_mentors = @mentors.bg_check_unsubmitted.consent_not_signed
+      #FIXME: these queries were cribbed from the datagrid, but should be shared somehow
```
stenington: This is the reason this is marked as **WIP**.

@adrichey My inclination here is to move the queries you added in #1915 to the accounts datagrid into scopes on Account. I know there will be duplication between those scopes and the scopes in e.g. MentorProfile, but I'm thinking of maybe trying to address that with tests? I'm not sure if there's a better way to share scopes between the two models, but I can test that both scopes give me the same users in a variety of scenarios.

So basically there would be `Account.onboarded_mentors` and `MentorProfile.onboarded`, and tests would help ensure those queries match, and then whichever is most appropriate can be used in the accounts datagrid and in these data analyses classes and we can expect the same results from both. (Repeat for students.)

What do you think about that?

adrichey: We already have quite a bit of duplication on the Account models that also falls on the mentor and student profile models, so I am not particularly turned off by the duplication, provided the tests are in place. I think that sounds reasonable.

stenington:
Some notes for review:

This PR tries to standardize the definition of `onboarded` for a mentor a bit. It does away with the `MentorProfile.onboarded` scope, which as best I can tell was used only in tests. Then it adds `Account.onboarded_mentors` and `Account.onboarding_mentors` scopes that try to match the logic in `MentorProfile.onboarded?` in terms of requirements.

For the dashboard graphs, it does away with the `international` concept which was used to display graphs including or excluding background checks, depending on location. That's now buried inside the concept of "onboarded" so both international and US RAs can see the same mentor graph.

For students, we're switching from the notion of "permitted" to the notion of "onboarded", so there's a whole bunch of renaming of things to follow suit.

PR #2019, opened 2019-03-04
Remove skipped test that we don't intend to fix

stenington:
🗑


stenington:
I searched for some of the strings the tests check for, and they only show up in the test. I think we're in the clear to remove these.

Issue #1987, opened 2019-02-20
Force users to enter a location and data agreement


adrichey:
Right now, users who have been invited to the platform without going through the registration form do not have to enter a location or adhere to the data agreement when they log in. After speaking with Katie and Ophelie, it is desirable to have all users fill out the data agreement and enter a location for data collection and event/workshop matching purposes.

**Proposed Solution**

We create a route guard on the platform when the user logs in and visits any page.
- If they don’t have a data agreement on file, we reroute them to the data agreement form so that they are forced to agree before they can continue.
  - We display some informational text on the data agreement form that tells them that they need to have the agreement on file before they can continue using the application as normal.
- Once they have a data agreement on file, if they don’t have a city/country on file, we reroute them to the location form so that they are forced to enter a location.
  - We display some informational text on the location form that tells them that they need to have a location on file before they can continue using the application as normal.
- Once they fill out the data agreement enter a location, they can route to other places on the platform as usual.

**Notes**
- The preferred deadline for this is March 20th, with April 1st at latest.
- @ophelieh and @kmorton2017 will provide copy for the message on the location form to the end-users.

**Acceptance Criteria For Testing**

@adrichey will provide testing accounts for these scenarios. Reach out to him before testing.

Scenario 1 - Data agreement missing, location present
1. Log in as a user without a data agreement, but location present (Alan to provide account)
2. Attempt to click links or navigate to different sections of the platform
3. You should be rerouted to the data agreement form and receive a message about why you must agree to the terms
4. Read the data agreement terms and check the agreement checkbox; submit the form
5. Confirm that you can now use the application as normal now that you have a data agreement on file and a location is present

Scenario 2 - Data agreement present, location missing
1. Log in as a user without a location (city and country present), that has agreed to the terms (Alan to provide account)
2. Attempt to click links or navigate to different sections of the platform
3. You should be rerouted to the location form and receive a message about why you must enter a location
4. Enter a location
5. Confirm that you can now use the application as normal now that your location is present

Scenario 3 - Data agreement missing, location missing
1. Log in as a user without a data agreement or location data (Alan to provide account)
2. Attempt to click links or navigate to different sections of the platform
3. You should be rerouted to the data agreement form and receive a message about why you must agree to the terms
4. Read the data agreement terms and check the agreement checkbox; submit the form
5. Attempt to click links or navigate to different sections of the platform
6. You should be rerouted to the location form and receive a message about why you must enter a location
7. Enter a location
[...]

PR #2021, opened 2019-03-04
#1987 Force all student/mentor/judge accounts to have a data agreement and location

adrichey:
Addresses #1987.


app/controllers/data_use_terms_controller.rb
```diff
@@ -0,0 +1,18 @@
+class DataUseTermsController < ApplicationController
```
adrichey: I created this controller to avoid having a hacky work around get introduced into the registration endpoint.

app/javascript/application/index.js
```diff
@@ -0,0 +1,63 @@
+import Vue from 'vue'
```
adrichey: This was moved out of the Webpacker `packs/application.js` file to keep that entry point file cleaner.

app/javascript/data_use_terms/store/actions.js
```diff
@@ -0,0 +1,15 @@
+import actions from 'registration/store/actions';
```
adrichey: Basically just overwriting a single function to keep existing functionality in place that loads in all of our account data in the `registration/initAccount` action in our Vuex store. This hits our new endpoint to update the `terms_agreed_at` status in the database.

app/models/account.rb
```diff
@@ -245,6 +245,8 @@ class Account < ActiveRecord::Base
     end
   }
 
+  after_initialize :populate_terms_agreed_from_signup_attempt
```
adrichey: I chose to implement this `after_initialize` callback instead of a migration as a migration would take around 4 hours to touch every account with missing terms. 😬 I had originally done this work as a migration and then changed my mind once I started running it locally.

adrichey:
@stenington this is _finally_ ready for your review. I know there is a lot going on here. Feel free to reach out if you have any questions. If anything concerns you and you want to do some of it differently, I am open to suggestion. 🙂

config/routes.rb
```diff
@@ -322,6 +322,9 @@
     resources :expertises, only: :index
   end
 
+  get "/account_data/data_use_terms/edit", to: "data_use_terms#edit"
```
stenington: I prefer resources notation for configuring routes because it forces sticking to the Rails conventions, although I recognize that doing that here will take some moving things around. But for example, in Rails you wouldn't POST to `/whatever/edit`, you would PUT or PATCH to `/whatever` to do an update. I'd rather stick to conventions than have to scrutinize code to know which way we're doing it per endpoint.

Do you have a strong preference for doing it this way? Let's discuss and make a decision on how to approach this one.

spec/features/student/create_a_team_spec.rb
```diff
@@ -16,28 +16,31 @@
...
-
-    within(".navigation") { click_link "Create your team" }
-    fill_in "Name", with: "Awesomest Saucesests"
-    click_button I18n.t("views.application.create",
-                        thing: I18n.t("models.team.class_name"))
-
-    team = Team.last
-    expect(current_path).to eq(edit_student_team_location_path(team))
-  end
+  ###
```
stenington: I think probably just delete this test if it's no longer relevant? Otherwise there's an rspec `skip` directive if you want to skip a test until some future date or whatever, but it seems like here we can just remove this one.

app/controllers/concerns/force_data_use_terms_agreement.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def no_data_agreement
+    !current_account.terms_agreed?
+  end
+
+  def not_on_data_agreement_form
+    return false if request.xhr?
+
+    begin
```
stenington: What's the scenario where Rails wouldn't be able to recognize the request path or the data use terms edit path? Is it if the user puts in a garbage url or something?

If Rails can't recognize `account_data_data_use_terms_edit_path` then something is very wrong, so maybe that should be outside of the begin block to fail fast.

config/routes.rb
stenington: This page has some useful tables for the routing conventions `resources` follows: https://guides.rubyonrails.org/routing.html#singular-resources

app/controllers/concerns/force_location.rb
```diff
@@ -0,0 +1,45 @@
...
+  def invalid_location
+    !current_account.valid_address? || !current_account.valid_coordinates?
+  end
+
+  def not_on_location_form
+    return false if request.xhr?
+
+    begin
+      original_request_path = Rails.application.routes.recognize_path(request.original_fullpath)
+      location_form_path = Rails.application.routes.recognize_path(
```
stenington: I'm not sure this is right. If all scopes have a location details path, then the same as the comment above applies here and maybe this shouldn't be in the begin/rescue. If not all scopes have a location details path, then I would think by definition they can't be on the location form and this should return true, not false.

app/controllers/data_use_terms_controller.rb
stenington: Makes sense to me!

app/controllers/concerns/force_data_use_terms_agreement.rb
```diff
@@ -0,0 +1,42 @@
...
+    current_account.authenticated? &&
+      current_account.respond_to?("#{current_scope}_profile") &&
+        !!current_account.send("#{current_scope}_profile")
+  end
+
+  def no_data_agreement
+    !current_account.terms_agreed?
+  end
+
+  def not_on_data_agreement_form
```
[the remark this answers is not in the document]
adrichey: Makes sense! I can totally see that. I will get that changed as I am all for reducing cognitive load. 🙂

app/controllers/concerns/force_data_use_terms_agreement.rb
```diff
@@ -0,0 +1,42 @@
...
+    return false if request.xhr?
```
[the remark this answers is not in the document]
adrichey: I like that. Good call.

app/controllers/concerns/force_data_use_terms_agreement.rb
adrichey: It failed for me during a redirect. For some reason it was trying to visit the "/judge/location" path without passing the right parameters, so I think it might need reassessment. I will dive into this a bit more.

app/javascript/data_use_terms/store/actions.js
```diff
@@ -0,0 +1,15 @@
...
+
+export default Object.assign({}, actions, {
+  updateTermsAgreed ({ commit, state }, { termsAgreed }) {
+    commit('termsAgreed', termsAgreed)
+
+    axios.post('/account_data/data_use_terms/edit', {
+      termsAgreed,
+      email: state.email,
+    }).then(({ data: { data: { attributes } } }) => {
+      commit('termsAgreed', attributes.termsAgreed);
```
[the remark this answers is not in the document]
adrichey: Correct. I want to take another pass at this today. I think we may be able to remove the first or second commit here. I will look into this.

config/routes.rb
adrichey: That makes total sense and I am in no way married to the way I am doing things. Convention is better. I can move things around. Not a problem.

spec/features/student/create_a_team_spec.rb
adrichey: I actually meant to remove this after I had updated the other tests and forgot. I will purge it. This was an oversight.

adrichey:
@stenington I have addressed the changes you wanted me to make but hit a wall where I have test failures locally. Basically, anything that uses the location form isn't recognizing the request as an XHR request, so I am going to have to take another crack at the failing tests tomorrow. Everything works without issue in the browser, so I am not sure what is going on here. I will need to look at this again tomorrow with fresh eyes. Maybe I am missing something straightforward here. I pushed up my code in any case.

adrichey:
@stenington This should once again be ready for review. The issue with the tests was that we weren't initializing axios properly. It ended up being a front-end issue so things that should have been sending as XHR requests weren't. We should be good to go now.

app/javascript/config/axios.js
```diff
@@ -9,6 +9,9 @@ window.axios = applyConverters(axios.create())
 document.addEventListener('turbolinks:load', () => {
   const csrfTokenMetaTag = document.querySelector('meta[name="csrf-token"]')
 
+  // Set all axios request to go through as XHR
```
stenington: This isn't really true, is it? Axios requests are XHR without this header, but this header is what Rails checks with `request.xhr?` I think, so this config is more about giving Rails what it wants rather than changing axios (unless I'm understanding axios wrong).

app/controllers/concerns/force_location.rb
```diff
@@ -0,0 +1,49 @@
...
+
+    location_form_path = Rails.application.routes.recognize_path(
+      send("#{current_scope}_location_details_path")
+    )
+
+    original_request_path == location_form_path
+  end
+
+  def ajax_request
+    request.xhr? || request.xml_http_request?
```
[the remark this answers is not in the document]
adrichey: It totally is. I will omit the latter.

app/javascript/config/axios.js
adrichey: You are correct. This is mainly to make Rails happy. Because when this code runs during rspec tests, essentially the the block with the csrf-token is skipped. I am not sure what is going on in the internals of Rspec, but this bit is needed to fix those tests and get axios sending things as XHR and not text/html. So the comment is misleading. I will just remove the comment. I don't want to remove the line, as I think it is a good practice to enforce that header regardless.

adrichey:
@stenington I have tweaked the code a bit for a final pass-through. If you could take another look at it I would appreciate it. :)

app/javascript/config/axios.js
stenington: Oh yeah, 100% agree the header is needed, just thought the comment was a bit off and I can't resist the urge to be pedantic all the time.

Issue #1921, opened 2019-01-09
Add helper text to company drop-down for registration


adrichey:
**Is your feature request related to a problem? Please describe.**
It is not necessarily clear that someone should hit the enter key when using the Company selection drop-down during registration for mentors.

**Describe the solution you'd like**
Add some sort of disclaimer under Company when mentors register so they know to press 'Enter' to validate their company if it isn't in the drop-down list.

Something like: If your company is not part of the options, type the name of your company then press 'Enter'"

**Developer Notes**
Update `vue-select` at the same time while addressing this ticket so that selecting a value on tab works for consistency throughout the application.

ophelieh:
Yay! Thank you for creating a ticket for this @adrichey . :)

Allicolyer:
Hey @adrichey I pulled this into to do.  What do you think?

adrichey:
Thanks for pulling this in. I will get started on it soon.

adrichey:
This is now ready for testing!

KaiLokken:
The new message about adding your company is up and running!

jessicabryan:
I tested it and it works. I'm wondering if we could add the sentence directly below school or company name heading. When you are typing the company name, where it currently is, gets blocked. You can see it from the beginning, but then disappears.

![screen shot 2019-03-07 at 1 55 02 pm](https://user-images.githubusercontent.com/47198665/53992005-a8c75b00-40e0-11e9-8525-349efb43d00a.png)

ophelieh:
Oh that's a good point, Jess. @stenington Do you think you would have time to look at this today? Otherwise, we can wait for Alan to get back. 

Thank you!

PR #2022, opened 2019-03-05
#1999 School/Company Name drop-down improvements

adrichey:
Addresses #1999.

Adding helper text we created in #1921, to the other sign-up areas for judges, mentors, and RAs.


adrichey:
@stenington This is ready for review. It is mostly copy, but I snuck in a test fix for that sporadic Israel test failure. Hopefully this will make it more consistent. I got it to fail twice (in ten runs locally) before adding that `within` block, and it didn't fail on the next ten runs after that fix was put into place locally. It may not solve every failure, but I feel like it strengthened the test a bit in any case.

Issue #1984, opened 2019-02-14
Code Checklist-Paper Prototype and Pseudocode 


kmorton2017:
**Describe the bug**
I tried to upload a file called "paper-prototype" into the prototype section of the code checklist. I also tried to upload a file called "pseudocode" into the pseudocode section of the checklist. The first time it happened, it selected a random screenshot on my computer for both (see thunkable logo). I tried again to choose the correct file for both the prototype and pseudocode sections. This time, the same file (pseudocode) appeared for both. 

**To Reproduce**
1. Start a submission.
2. Click on code checklist.
3. Scroll down to paper prototype and select a file. 
4. Scroll down to flowchart or pseudocode and select a file 
5. See the same image appear for both.

**Expected behavior**
I would expect these to be different images for the prototype and flowchart or pseudocode sections. 

**Screenshots**
![flowchart or pseudocode](https://user-images.githubusercontent.com/32206427/52810655-e556e800-3047-11e9-8dab-0b3e9dd4f88c.gif)

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**
 - Device: Macbook Air 
 - OS: OS Mojave
 - Browser Chrome
 - Version 10.14.2

**Additional context**

**Label**
You MUST use the label menu to add 'bug' and '0 - To discuss' as a label
(do *not* choose a bug priority label, devs will do that according to the Defect Priority matrix)

kmorton2017:
@ophelieh  @adrichey @stenington this same issue was brought up by a user today on zendesk.(Text below)  Since we are encouraging students to start submitting, can we move this into the to do column? 

We seem to be having problems with uploading our flowchart and paper prototype. We originally had a paper prototype and flowchart uploaded. We then uploaded a better version we had created of our paper prototype which had for some reason uploaded in the flowchart area as well. When I went into the code checklist today it says for both the flowchart and the paper prototype 'Unprocessed uploads'. When I tried to re-upload our flowchart and paper prototype it wouldn't work. Your help would be greatly appreciated.

ophelieh:
@stenington or @adrichey Would you be able to look into this? 

Thank you!

stenington:
I can start looking at this one.

adrichey:
Thanks for getting to this @stenington. Sorry, email got away from me today.

adrichey:
There have been some Airbrake JS errors today around this area. I am not sure if they are directly related, but it might be a good idea to check there when you start this @stenington.

[...]

PR #2026, opened 2019-03-06
#1984 give unprocessed uploads a unique directory

stenington:
Fix for #1984, copied from https://github.com/carrierwaveuploader/carrierwave/blob/f15f4ba71dd2c78e8df494bf65938bbaf13960f9/lib/generators/templates/uploader.rb#L12

I think unprocessed uploads were all overwriting each other on S3 😱

### DIFF
--- a/app/uploaders/image_uploader.rb
+++ b/app/uploaders/image_uploader.rb
@@ -13,6 +13,10 @@ class ImageUploader < CarrierWave::Uploader::Base
   # Override the directory where uploaded files will be stored.
   # CarrierWaveDirect::Uploader puts raw uploaded files in this directory on S3 as a first step
   def store_dir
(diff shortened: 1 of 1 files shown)


adrichey:
@stenington I began looking at this code locally and began by creating a new team for a new student and when I land on the team page, I get the following error: `undefined method `id' for nil:NilClass`. It looks like the CI tests caught this as well. It looks like the model is nil here, so I think we need to initialize that somehow. It also looks as though mounted_as is nil here, so maybe we are missing initialization somewhere?

stenington:
Yeah it seems like uploaders behave differently when mounted or not, and this one is used in both ways. I have kind of a quick-fix running in CI now, although we can talk about maybe not using uploaders in both scenarios?

stenington:
It passed on CI so I'm requesting a review, but I fully recognize this was a lazy fix.

PR #2028, opened 2019-03-08
#1999 QA fix: Move school/company field hints above inputs

stenington:
I think this does it. More involved than I expected!


app/assets/stylesheets/_base.scss
```diff
@@ -89,6 +89,10 @@ table {
   font-size: 0.9rem;
 }
 
+.hint_above {
+  margin-bottom: 0.25rem;
```
adrichey: The only thing I would suggest here would be to maybe change this class to `.hint--above` to be in line with BEM standards. It isn't a deal breaker for me though as Ruby and I have plans to eventually do an audit of the CSS anyway so that she can be more involved with the UI work for TC. I am going to mark it as approved since the solution works well. This really did get more involved haha!

stenington: I'll make the change, that makes sense to me.

Issue #1949, opened 2019-01-25
Include incomplete scores on printable PDF for events


Allicolyer:
**Description**
RAs should be able to download a PDF of scores for RPEs, including incomplete scores. Right now only complete scores show up. RAs use this printable PDF to print out before their event starts and hand to the judges there. The judges should in theory be able to finish a half complete score. 

**Acceptance Criteria**

1. Turn on events as an admin
2. Log in as a RA (ra@ra.com)
3. Create an event
4. Invite judges and two teams to the event
5. Turn on judging as an admin
6. Log in as the judge. Complete a score for the first team.
7. Start, but don't finish a score for the second team.
8. Log back in as an RA. Confirm that both the complete and incomplete scores appear on the printable PDF. You can find this printable PDF under the events menu.

kmorton2017:
update from meeting on 02/08 do not prioritize this

stenington:
Grabbing this one because it was at the top of the backlog (despite the do not prioritize note)

stenington:
Use the Acceptance Criteria in the initial description as QA steps (although feel free to find an existing event where there are both complete and incomplete scores).

ophelieh:
Hi @kmorton2017 could you test this today or tomorrow? 

Thank you!

kmorton2017:
This is a scenario that will likely not happen often in the real world but thought I would share. 

As an RA (katie+ra7@iridescentlearning.org) I created an event in China and added 2 judges. I logged in as Alicia Keys and completed score for the first team and left incomplete scores for the other team. 

I logged in as the other judge (Dolores Umbridge) who unbeknownst to the RA was already judging at another event.  I logged in as the judge Dolores and saw only the RPE in LA not the 2nd event in China I was also scheduled for.

When I logged in as the RA to see the PDF, i saw both Alicia Keys scores for the Chinese teams, and oddly enough was able to see Dolores' scores for the RPE in LA with those teams listed on the PDF.

I don't think this is a usual case (since I unknowingly added a judge who was scheduled at another event, and also one in another country). 

Let me know if my feedback needs any more explaining!
Katie

stenington:
Thanks @kmorton2017, that sounds to me like another bug with judges in multiple events. 😢 

Let me see if I can address it easily for these printable PDFs though.

stenington:
Ok @kmorton2017 I was actually able to account for this scenario, thanks for finding it! These printable pages should only show scores for submissions in the event now.

[...]

PR #2030, opened 2019-03-12
#1949 QA fix: don't assume judges are only in one event in printable page

stenington:
The printable events page had a bug where it assumed all of a judge's scores would be for that one event in the case of assigned judges. Now it filters to only show scores on submissions that are part of that event.


spec/features/regional_ambassador/print_scores_spec.rb
```diff
@@ -75,4 +75,44 @@
...
+
+  scenario "other event scores are filtered out when assigned judges are in multiple events" do
+    ra = FactoryBot.create(:ra, :approved)
+    event_a = ra.regional_pitch_events.create!(FactoryBot.attributes_for(:event, :senior))
+    event_b = ra.regional_pitch_events.create!(FactoryBot.attributes_for(:event, :senior))
+
+    judge_a = FactoryBot.create(:judge, first_name: "Alvin")
+
+    team_a = FactoryBot.create(:team, :live_event_eligible, :senior, name: "Ducktales")
+    team_b = FactoryBot.create(:team, :live_event_eligible, :senior, name: "Darkwing")
```
adrichey: Great references haha.

Issue #2025, opened 2019-03-06
Make judging rounds settings-based only


stenington:
**Is your feature request related to a problem? Please describe.**
Judging rounds are currently configured through a blend of dates and toggles in a way that's hard to explain and understand. We thought we needed that when live and virtual QFs were going to end at different times, but that created difficulties at a program level so we backed away from it.

**Describe the solution you'd like**
It would be nice now for the judging round changes to be **entirely based on admin settings**. Date configuration should continue to exist for display only.

The judging round settings should be expanded to include:

* Off
* Quarterfinals
* Between rounds
* Semifinals
* Finished

Staff will change the configuration at the right time manually, judges should never be prevented from doing something by date alone.

stenington:
@kmorton2017 @ophelieh @adrichey This is my proposal for changing how the judging round setting works so that admins would have full control over the various judging phases. Off, QFs, and SFs would basically be the same as we already have. Between rounds would show that special between rounds page, and Finished would basically be the same as Off but could be used in the future to show a thank you page or something like that.

Does this make sense? Does it seem like a good thing to have? I've done some exploration and I'm pretty confident I can build it in a reasonable timeframe for this season.

adrichey:
I think it makes sense. If we ever want to go down the road of having separate end dates for virtual and live judging, that should be a bigger conversation that will definitely require a more major rework; so I like this solution now that we aren't going down that road for now. I think these states all make sense to me based on what I have seen for our various states and screens that our end users see. And it will be nice knowing there is one source of truth for the season settings.

ophelieh:
Makes perfect sense. 

To have it documented - 
* In between round is when QFs is still turned on and we are waiting on semifinals to start - admins do not have to change any settings. 
* When semifinals are ready to start, an admin will change toggle from QFs to semifinals.

stenington:
@ophelieh Well with the change I'm suggesting, an admin **would** change from *Quarterfinals* to *Between rounds* when you wanted QF judging to end and the between rounds page to display. That way we'll have 100% control over it, and the platform won't surprise us by turning off quarterfinals early because of some mistake we made with date configuration or whatever.

ophelieh:
Oh so we would add another toggle? Am I understanding that right? 

Thank you!

stenington:
Yeah, it would probably look like this: 

[...]

PR #2031, opened 2019-03-13
#2025 more judging round settings

stenington:
This PR is for #2025, which we're talking about tomorrow. If there are no objections during that convo, I think this will be good to go after review. I'll hold off requesting review for now though.


stenington:
@adrichey This one is ready for review when you have a chance.

Issue #2037, opened 2019-03-14
Test that a submission with 10 pending scores can still be assigned to a judge


stenington:
Add a test or tests to `spec/technovation/find_eligible_submission_id_spec.rb` that tests that a submission with `SCORE_COUNT_LIMIT` pending scores can still be assigned to judges if all other submissions have `SCORE_COUNT_LIMIT` completed scores.

Issue #2036, opened 2019-03-14
Test that a submission with no scores will be assigned before one with pending scores


stenington:
Add a test or tests to `spec/technovation/find_eligible_submission_id_spec.rb` that tests that an otherwise valid submission with no scores, pending or complete, will be assigned before a valid submission with a pending score.

PR #2044, opened 2019-03-15
#2036 #2037 test more submission selection scenarios for scoring

stenington:
Testing some scenarios I've heard from several people were problematic in the past. Looks like they've been addressed.


stenington:
@adrichey Amy, Ophelie, and Katie all described problem scenarios from last year that were testable, so this just adds a couple of test scenarios. They pass, so I think the issues had already been fixed as we thought.

Issue #2054, opened 2019-03-22
Admin scores page round selection doesn't handle new rounds


stenington:
**Describe the bug**
If you put the platform into "between rounds" or "finished" for the judging round and try to view the scores page, it errors out. 

**To Reproduce**
1. As an admin, go to Content & Settings > Judging
2. Select Finished
3. Review and save the setting
4. Click on Scores in the nav menu
5. See the error

**Expected behavior**
You should see the scores page with quarterfinals selected as the round

stenington:
[This block of code](https://github.com/Iridescent-CM/technovation-app/blob/qa/app/controllers/admin/scores_controller.rb#L34) is the culprit, I think.

stenington:
QA steps:

1. log in as admin
2. set Judging Round to "Finished"
3. click Scores in nav menu
4. confirm the scores page loads without error

ophelieh:
@nickiredescent and @kmorton2017 Could you both test this please?

nickiredescent:
Tested. Can confirm that the score page is loading correctly when either the "Between Rounds" or "Finished" status is turned on.

stenington:
I just noticed this bug exists on the scores export page too. 😭 

I'm knocking this back into "In progress" to fix that page as well.

stenington:
QA steps for second fix (same thing but with Scores export page):

1. log in as admin
2. set Judging Round to "Finished"
3. click "Scores export" in nav menu
4. confirm the scores export page loads without error

nickiredescent:
Tested. Can confirm that the "Score export" page loads without error. Could be just me, the page loads a bit slowly, especially when I switch to quarterfinal scores.

ophelieh:
@kmorton2017 Could you test both of those steps? Merci! :)

kmorton2017:
Confirm that the scores page loads without error.
Confirm that the scores export page loads without error. I agree with Nick though, the scores export page was slower to load though.

stenington:
Ok, it looks like it might be slow because the country column hits the Google API, and while our API requests are being rejected that slows things down pretty badly.

PR #2058, opened 2019-03-26
#2054 default scores round to something sensible in all cases

stenington:
This makes the admin scores page try to pick a sensible value for the round based on the current round, if the user hasn't otherwise specified.


adrichey:
@stenington Feel free to add me as a reviewer when this is good to go. Can this wait until the next release, or should we try to get this into the release going out this week?

stenington:
It can wait

Issue #2060, opened 2019-03-26
Certificates: Add certificate section to admin UI participant pages


adrichey:
**Is your feature request related to a problem? Please describe.**
We need to add a certificates section when viewing a participant in the admin UI.

**Describe the solution you'd like**
This section should appear when you click on a student, mentor, or judge in the Participants section of the admin UI. This section should display what certificate(s) the user has and give the admin user the ability to download the generated certificate.

**Additional context**
This does not give the admin user the ability to modify the certificate(s) in any way; it is purely informational so that they can access certificates that have been generated. The ability to generate a new certificate will come in a future ticket to be linked later.

adrichey:
Ability to generate a certificate in this view will come in the work for #2061.

adrichey:
This is ready for testing.

ophelieh:
@nickiredescent and @kmorton2017 Could you both test this? Thanks!

kmorton2017:
I confirm that I can see the Certificate Section on the Participant's profile for student, mentors, and judges.

Issue #2056, opened 2019-03-26
Remove on-demand generation of season certificates


stenington:
This ticket gets us started on a transition from on-demand certificate generation to batch generation, which we're expecting to be easier to write and less buggy.

This ticket's work includes:
* pulling out FillPdfs calls from controllers or other places it's called on-demand right now
* stop using the CertificateButton Vue app in favor of querying in the controller and rendering what's found
* ~add logging where appropriate for unexpected data scenarios, e.g. pulling up two certificates where you would only expect one~ 
    * not sure this one is appropriate, it may make more sense as part of a defensive certificate generation strategy

stenington:
@adrichey I don't feel like this ticket needs QA by itself. It's the groundwork for a bunch of other tickets that will all get QA'd so I think we're better off moving on to those tickets. What do you think?

adrichey:
I think you are correct. My assumption on this work would be that it would be a developer ticket that would be closed on pull request. I feel like the new UI and functionality will cover the actual QA testing.

kmorton2017:
are there any action items for the program team on this ticket @adrichey and @stenington

adrichey:
@kmorton2017 No, this is basically all "under the hood" work. I will go ahead and close it since this work is now on QA and other work we are doing is dependent on it. Thanks!

PR #2065, opened 2019-03-27
#2056 removing on-demand certificate creation

stenington:
This PR gets things to a state where certificates are shown to users through simpler controller queries without attempting to generate certificates on demand, and it temporarily removes certificates from the admin (#2060 will put them back a little differently).


stenington:
@adrichey Do you feel like I should remove the CertificateButton Vue app in this PR?

stenington:
@adrichey I think my remaining work here is to remove the CertificateButton app, and to remove the certificate override model field and endpoint. No current judge has an override set, so we're not losing data. There's a `drop_lowest_scores.rb` rake task we may need to think of a way to account for, but I'd rather do something like marking scores as dropped and then including them in the judge certificate level calculation than doing the override.

Just keeping you up to date.

adrichey:
I think removing the CertificateButton and the endpoint makes sense at this point. Good to know that no one has an override right now. I think that marking the scores as dropped and then including them in the judge level calculation sounds like the right approach as well. Thanks for working on this Mike!

adrichey:
@stenington Can you please rebase QA into your branch by running `git rebase origin/qa`? I tried to run it locally to get everything current with the badge removal changes that are in QA, but I am getting conflicts and I think you will be better suited to solving those at this point. Thanks!

stenington:
@adrichey I've been merging in QA, so i just merged again. The changes look correct for review.

stenington:
And I just remembered to remove the WIP label.

adrichey:
Reviewing now! Thanks Mike.

spec/features/mentor/certificates_spec.rb
```diff
@@ -39,4 +39,24 @@
       href: ENV.fetch("FINALISTS_URL")
     )
   end
+
+  scenario "appreciation certificate per team" do
```
adrichey: Nice additional coverage!

Issue #2053, opened 2019-03-21
No thank you page for judges in Finished Stage


ophelieh:
**Describe the bug**
Once we close judging, there isn't any thank you page. It's just blank.

![image](https://user-images.githubusercontent.com/32206421/54790106-fb396900-4c02-11e9-8af6-d25dd25b4833.png)

**Expected behavior**
It would be nice to have a sort of thank you page after judging is in the "finished" phase. 

Something along the lines of "Thank you for your help scoring and giving valuable feedback to Technovation teams. Your judging certificate will be available June 24th. Thank you again for helping girls lead the change. See you next season!"

adrichey:
@ophelieh Should we pull this in now since it will pop up for our end users this season and might be jarring? I can just add that text to their dashboard for now. It won't take long to knock that out.

ophelieh:
Works for me! Thank you!

adrichey:
This is now ready for testing.

Testing Scenario 1:
1. As an admin, turn Judging round to "Finished", make sure that "Scores & Certificates Accessible" is left unchecked and save these settings.
2. Log in as a Judge account.
3. Confirm that a message is displayed informing the user that certificates will be available on June 24th.

Testing Scenario 2:
1. As an admin, turn Judging round to "Finished", make sure that "Scores & Certificates Accessible" is checked. Save these settings.
2. Log in as a Judge account.
3. Confirm that the blank box does not appear and that there is either a certificate available for the user, or that a message is displayed in the absence of a certificate.

ophelieh:
Hi @KaiLokken and @nickiredescent Could you test this? 

Merci! :)

nickiredescent:
Tested the scenarios. 
Can confirm that Scenario 1 provides a message saying that certificates will be available June 24th.
Can confirm that Scenario 2 has either a certificate available or if no certificate is available, a message is displayed.

PR #2069, opened 2019-03-28
#2053 Implement thank you page for finished judges

adrichey:
Addresses #2053.

Added a thank you page that appears when the judge rounds are set to finished, but scores and certificates are not yet available. Once scores and certificates are available, this disappears as we will either display the certificates to the user or a "no certificates available" panel.

### DIFF
--- a/app/views/judge/dashboards/show.en.html.erb
+++ b/app/views/judge/dashboards/show.en.html.erb
@@ -72,7 +72,9 @@
         <%= render 'judge/dashboards/onboarded/certificates' %>
       <% end %>
 
-      <%= render 'judge/dashboards/onboarded/events' %>
+      <% if !SeasonToggles.display_scores? %>
+        <%= render 'judge/dashboards/onboarded/events' %>
+      <% end %>
     </div>
   </div>
 </div>
--- a/app/assets/stylesheets/_utilities.scss
+++ b/app/assets/stylesheets/_utilities.scss
@@ -50,6 +50,10 @@
   cursor: pointer;
 }
 
+.display--block {
+  display: block;
+}
+
 .font-size--small {
   font-size: 0.8rem;
 }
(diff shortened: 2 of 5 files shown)


app/views/judge/dashboards/show.en.html.erb
```diff
@@ -72,7 +72,9 @@
         <%= render 'judge/dashboards/onboarded/certificates' %>
       <% end %>
 
-      <%= render 'judge/dashboards/onboarded/events' %>
+      <% if !SeasonToggles.display_scores? %>
```
stenington: `SeasonToggles.quarterfinals_or_earlier?` might be better here?

adrichey: That events view also contains logic for the between rounds functionality and also the finished state. So I think this should probably stay as is, or we should split the logic in the `judge/dashboards/onboarded/events` out into multiple files and put the conditionals in this `show` file. What do you think?

stenington: Oh I see! Yeah, that's a bit confusing. So `_events.en.html.erb` isn't really events, it's judging controls (whether that's an event or virtual judging or the between rounds message or the final thank you).

I believe there will only be a brief window where the judging round is set to "finished" and displaying scores is off. Should the thank you only be shown in that window, or should it still be visible alongside scores and certificates? I guess that's the real question I have.

adrichey: Yeah, it's a little weird. There is yet another thank you screen that is displayed in the form of `judge/dashboards/onboarded/certificates` > `dashboards/no_certificates`. So essentially, the screen in this PR will be replaced by the certificates or no certificates screens as soon as score become available. So to answer your question, I think it should only appear during that window.

stenington: Oh I see, ok! 👍

Issue #2061, opened 2019-03-26
Certificates: Add ability to generate certificates for a participant


adrichey:
**Is your feature request related to a problem? Please describe.**
Add the ability for an admin user to generate a different certificate for an end user on a Participant's page.

**Describe the solution you'd like**
After clicking on an account in the Participants page datagrid in the admin UI, you should see a certificates section (created in #2060) where you can both view currently generated certificates for a user, and have the ability to generate a new certificate. An admin user should get drop-down selectors in order choose the type of account they are generating a certificate for (student, mentor, or judge), what team the certificate should be generated for in the cases of mentor and student, and what type of certificate they should be receiving. There will also be a button to generate the certificate using the drop-down input. Here is a list of combinations that should be available (please note that these may change before QA ready):

- Student
  - No profile type selector
  - Team drop-down should only have one option, the team they are on for the season
  - Should have a drop-down to select a certificate type: Participation, Completion, Semifinalist
- Mentor
  - Should have a team selector to pick from a list of all teams they were on this season
  - Certificate type selector should be available but only have an option for Mentor Appreciation
  - Profile type selector should be present and able to flip between mentor and judge if they have both profiles
- Judge
  - No team selector
  - Certificate type selector should be available with the following options: Certified Judge, Head Judge, and Judge Advisor
  - Profile type selector should be present and able to flip between mentor and judge if they have both profiles

**Additional context**
This work should happen after the work for #2060 has been completed.

stenington:
QA steps:

For each of the following types of users (**Student**, **Mentor**, **Judge**, **Judge & Mentor**) do the following:

1. log in as an admin
2. look up the user in the **Participants** page
3. scroll down to the **2019 Certificates** section
4. confirm the appropriate controls are available for the profile type in the **Manually award certificate** section, e.g. student has student certificate, mentor has mentor certificate, etc.
5. use the controls to create a certificate
6. confirm the certificate appears in the **2019 Certificates** section (note you may have to wait and refresh the page if the certificate doesn't immediately appear)

ophelieh:
@KaiLokken & @nickiredescent Could you test this for student, judge, and judge & mentors

nickiredescent:
Tested. Can confirm that certificate generation works as described for all participants listed.

adrichey:
Thank you @nickiredescent! I will mark this as closed in that case.

PR #2071, opened 2019-03-29
Consolidate certificates views for admin participants UI into one place

adrichey:
Addresses #2060.

### DIFF
--- a/app/views/admin/participants/_certificates.en.html.erb
+++ b/app/views/admin/participants/_certificates.en.html.erb
@@ -7,7 +7,7 @@
       <ul>
         <% @certificates.each do |certificate| %>
           <li>
-            <%= link_to "#{certificate.cert_type.humanize.titleize} Certificate",
+            <%= link_to certificate.description,
               certificate.file_url,
               :target => "_blank" %>
           </li>
--- a/app/views/admin/participants/_certificates.en.html.erb
+++ b/app/views/admin/participants/_certificates.en.html.erb
@@ -1,0 +1,20 @@
+<h4 class="reset"><%= Season.current.year %> Certificates</h4>
+
+<div class="panel">
+  <% if @certificates.any? %>
+    <p>Certificates that are currently generated for this user:</p>
+    <p>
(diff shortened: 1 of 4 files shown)


adrichey:
@stenington I am not sure how large the scope of this should be. I have gotten this to the point where it displays certificates in the system. But I am not sure what else should be done to get this ready for the next steps in our overall solution. Can you take a look at this and let me know what else needs to be done here? I can fix the failing test(s) in the mean time.

stenington:
I think my PR might remove that test completely. You can write new ones if you want, but you might want to check ahead for merge trouble.

adrichey:
Yeah, I just looked at your PR. I think it might cause some merge drama, but I can cross that bridge when it comes to it. As soon as your PR gets merged in, I will rebase `qa` into this branch and go from there.

stenington:
I think this is perfect, scope-wise. I guess it may get tweaked as we figure out #2061, but I'd say let's start with this as-is.

adrichey:
I am going to go ahead and mark this as ready for review based on your other comments. I have rebased `qa` into this branch, so I am going to let CI catch any potential broken tests, but based on the changes we brought in from `qa`, I think we should be good to go.

app/views/admin/participants/_certificates.en.html.erb
```diff
@@ -0,0 +1,20 @@
+<h4 class="reset"><%= Season.current.year %> Certificates</h4>
+
+<div class="panel">
+  <% if @certificates.any? %>
+    <p>Certificates that are currently generated for this user:</p>
+    <p>
+      <ul>
+        <% @certificates.each do |certificate| %>
```
stenington: Maybe we should be including the team name somehow, if the certificate has a team attached? I'm picturing what this will look like for a mentor on several teams, and it'll be a bit hard to tell which is which without opening them all, I think.

adrichey: Good call. I will work on that.

Issue #2059, opened 2019-03-26
Certificates: Create a rake task to generate all certificates for the season


adrichey:
**Is your feature request related to a problem? Please describe.**
We want to generate ALL of the appropriate certificates for the season after semifinals ends but before we turn on certificates and scores availability for our end users.

**Describe the solution you'd like**
This will be a rake task that generates a batch of `Jobs` that will handle `Certificate` generation using `FillPdfs`. The rake task should only generate certificates for the 2019 season, but should be implemented in such a way that we can pass flags to the rake task so it can be run for different seasons, and have a flag so that we can limit the amount of jobs that are generated for testing purposes. For example, we can limit the task to only kick off jobs for the first ten records found.

stenington:
@adrichey Can you help me think through QA on this one? It will take some collaboration with the dev team because we'll need to run the rake task by hand. I was thinking something like the following:

1. Developer deletes all of this year's certificates
2. Developer runs one of the rake tasks to generate this year's certificates
3. QA team use Certificates admin page to look through and try to verify the results

Point 3 is deliberately vague; I don't think we need to look at every user in QA, but it would be nice to cover one user per type of certificate at least. What do you think?

adrichey:
@stenington I think that makes sense. I think we will want to go through and set up a scenario for each certificate type. Including General Judge, even though that functionality will be removed soon. Actually, I am almost wondering if we should wait to run this until after we remove General Judge? That might give us a more realistic testing example.

So the test cases in Step 3 become:

- General Judge: Judge account with <= 4 scores (optional, dependent on the assumption that this will be tested before #2001 is in place).
- Certified Judge: Judge account with 5 scores.
- Head Judge: Judge account with between 6 and 10 scores.
- Judge Advisor: Judge account with 11 scores.
- Completion: Student account on a team with a complete submission and that was a quarterfinalist.
- Mentor Appreciation: Mentor account that is not a judge on a team.
- Participation: Student account on a team with an incomplete submission (between 50 and 100% complete).
- Semifinalist: Student account on a team with a submission that made it to the semifinals.

stenington:
I agree about waiting until General Judge is gone. I would also add the test case of

* Judge + Mentor Appreciation: a Judge who qualified for a judge certificate and qualified for 1+ mentor certificates (make sure they received all)

stenington:
[...]

PR #2078, opened 2019-04-01
rework FillPdfs & CertificateRecipient

stenington:
This began as #2059, but I think has grown big enough that I should issue a PR now and finish the actual rake task for that ticket separately.

This reworks `FillPdfs` and `CertificateRecipient`, and introduces `DetermineCertificates` as the "brains" that determines the certificates a user needs. 

`CertificateRecipient` essentially becomes like a complex data type that associates a certificate type with an account and maybe a team. It provides helper methods needed in the PDF generation process as well, but no longer contains logic about which certificates an account actually needs.

`FillPdfs` uses `DetermineCertificates` to get a list of `CertificateRecipients` back for which to generate a pdf certificate. The nomenclature is a little confusing, because `FillPdfs` is called on a single account, and so the list of `CertificateRecipients` will all be for the same account, but it may represent multiple certificates for different teams for mentor, for example, or both a judge and a mentor certificate for a mentor/judge.


app/data_grids/judges_grid.rb
```diff
@@ -38,7 +38,22 @@ def self.current_judging_round
   end
 
   column :judge_rank do |account, grid|
-    CertificateRecipient.new(account).judge_certificate_level
```
stenington: I feel like this should reuse the logic in `DetermineCertificates`, so I'm working on that.

adrichey: Makes sense to me!

stenington:
If tests fail I may have to revisit, but I think you could start reviewing @adrichey

Issue #2063, opened 2019-03-26
Implement better logging and error handling for Jobs


adrichey:
**Is your feature request related to a problem? Please describe.**
Right now we don't have adequate logging or status updates when jobs fail. For example, if an export Job fails, we are not able to set its status to "failed." This has UI implications as failed jobs just stay in progress forever and to the end user it looks like the export is still running, even though it has failed. Another place where this is particularly useful is when Certificate jobs generate multiple certificates for an account/profile/team combo when it shouldn't.

**Describe the solution you'd like**
We should look into Sidekiq and our Job model and see if we can store an error status on the Job if the Sidekiq job fails in any way. We should also add better logging at this point if failures occur.

**Additional context**
See this documentation for more details: https://github.com/mperham/sidekiq/wiki/Error-Handling

** This is a developer only ticket which can be closed upon pull request approval and deployment to QA/staging. **

stenington:
@adrichey Can you give me more details on the export job failures you saw recently? I want to replicate that failure locally as a starting point for acquainting myself with job error handling options.

adrichey:
Sure. First I log in as an admin user and browse to "Background Jobs" in the dashboard. From here, under Dead Jobs, I basically look up an "ExportJob" in the job column for another user in the system. This will give me feedback similar to the following, where __ACCOUNT_ID__ is the actual account id of the person who ran the export:
```
__ACCOUNT_ID__, "AdminProfile", "AccountsGrid", {"season"=>"2018", "season_and_or"=>"match_any", "scope_names"=>"mentor", "column_names"=>["", "returning", "city", "state_province", "country"], "name_email"=>" ", "admin"=>true, "country"=>[], "state_province"=>[], "_aj_hash_with_indifferent_access"=>true}, "Admin::ParticipantsController", "->(scope, user, params) { scope }", "technovation-accounts-2019-03-27 13:51:30 -0700", "csv"
```
This basically shows me that I need to run a Participants export with the following settings:
- Columns: Returning?, City, State/Province, Country
- Profile Type: Mentors
- Season: 2018 (match any)
Then I pull up my server tail logs in my terminal and run the export locally using the production server data that caused the failure. From here, I wait until it bombs out and then I look into the failure. Hopefully that is thorough and helpful.

stenington:
Ok, I've learned a few things about Sidekiq, jobs, and our admin export notifications.

[...]

PR #2079, opened 2019-04-02
#2059 add rake tasks for directly creating certificates

stenington:
The rough plan is to add the following tasks:

* `certificates:award_direct` to call FillPdfs on one current account without a Sidekiq job
* `certificates:award_batch_direct` to call FillPdfs on a batch of current accounts without Sidekiq jobs
* `certificates:award_all_direct` to call FillPdfs on all current accounts without Sidekiq jobs
* `certificates:award_job` to call FillPdfs on one current account through a Sidekiq job
* `certificates:award_batch_jobs` to call FillPdfs on a batch of current accounts through Sidekiq jobs
* `certificates:award_all_jobs` to call FillPdfs on all current accounts through Sidekiq jobs

### DIFF
--- a/app/jobs/certificate_job.rb
+++ b/app/jobs/certificate_job.rb
@@ -16,16 +16,7 @@ class CertificateJob < ActiveJob::Base
 
   after_perform do |job|
     if db_job = Job.find_by(job_id: job.job_id)
-      certificate_type = job.arguments[0]
-      account = Account.find(job.arguments[1])
-
-      team = if job.arguments[2]
-               Team.find(job.arguments[2])
-             else
-               nil
-             end
-
-      recipient = CertificateRecipient.new(certificate_type, account, team: team)
+      recipient = CertificateRecipient.from_state(job.arguments.first)
 
       if recipient.certificate_issued?
         #FIXME: using .last here is still a bit risky
@@ -44,15 +35,7 @@ class CertificateJob < ActiveJob::Base
     end
   end
 
-  def perform(certificate_type, account_id, team_id = nil, **options)
-    account = Account.find(account_id)
-
-    team = if team_id
-             Team.find(team_id)
-           else
-             nil
-           end
-
-    FillPdfs.fill(CertificateRecipient.new(certificate_type, account, team: team))
+  def perform(certificate_recipient_state)
+    FillPdfs.fill(CertificateRecipient.from_state(certificate_recipient_state))
   end
 end
--- a/app/jobs/certificate_job.rb
+++ b/app/jobs/certificate_job.rb
@@ -4,7 +4,7 @@ class CertificateJob < ActiveJob::Base
   queue_as :default
 
   before_enqueue do |job|
-    account_id = job.arguments.first
+    account_id = job.arguments.second
     owner = Account.find(account_id)
 
     Job.create!(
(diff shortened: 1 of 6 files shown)


stenington:
@adrichey I'm noticing that `app/jobs/certificate_job.rb` and `FillPdfs` have a problem in that `FillPdfs` can attempt to generate multiple certificates, but `certificate_job.rb` is attempting to record the completion status of only a single certificate. 

One option is to skip using the `Job` model in the job and relying more on Sidekiq and/or Airbrake for figuring out the pass/fail status of certificates, maybe relying a bit on #2063. I guess another would be to build `FillPdfs` out more to allow specifying all the way down to a single certificate you want made and then kicking off a `CertificateJob` per certificate, which might incidentally be helpful for #2061. A third option is to just skip being able to create jobs for now, since these rake tasks would run in temporary dynos that theoretically can stay running as long as you want them to (hopefully I'm right about that), although I think #2061 would still want to kick off a Sidekiq job.

Any thoughts here on what makes sense to you?

adrichey:
Hmmmm. This is tough. Part of me thinks that being pressed for time makes the first or third solutions the most practical, but I feel like the second one makes the most sense in my opinion. Being able to kick off a job per single certificate might be the best option in terms of overall design. But it also sounds like a larger chunk of work. I am okay with any of these, but I think the second option sounds the most logical from a design standpoint.

stenington:
Yeah, I kind of think if option 2 doesn't happen now, it'll happen in #2061 so may as well do the heavy lifting up front. I'll see what I can figure out.

stenington:
Ok, so `CertificateJob` now takes a `certificate_type` argument first, and uses a new `FillPdfs.fill()` method that takes a single `CertificateRecipient` and generates the specified certificate.

Recording job errors still isn't perfect. I tried creating a `CertificateRecipient` job with an invalid certificate type, and the related `Job` just sits in the db as "queued" because the `after_perform` block errors out as well. But I kind of want to leave that as a problem to solve in #2063 when I have a better handle on Sidekiq errors and error monitoring in general so I know what the options are.

stenington:
One thing that's a bit clunky right now is the translation back and forth between `CertificateRecipient` and its serializable components that happens in the rake task: the task has a `CertificateRecipient`, creates a job with pieces of it, and then the job reconstructs the `CertificateRecipient` once in `perform` and once in `after_perform`. 

I feel like there's room to make that better somehow.

stenington:
Ok, since `CertificateRecipient.state` is serializable, I'm using that as the argument to the `CertificateJob` with a `.from_state` to recreate it. Possibly a bit silly but I can't quite wrap my head around a better solution right now. (I'm starting to wonder if `Certificate` and `CertificateRecipient` could be merged, but I'm not sure.)

adrichey:
> Ok, since `CertificateRecipient.state` is serializable, I'm using that as the argument to the `CertificateJob` with a `.from_state` to recreate it. Possibly a bit silly but I can't quite wrap my head around a better solution right now. (I'm starting to wonder if `Certificate` and `CertificateRecipient` could be merged, but I'm not sure.)

Looking over the code, it makes sense to me. I think in an ideal world the `Certificate` and `CertificateRecipient` would probably be merged, but I don't think it is necessary right now unless you want to take a crack at it. And agreed on the error handling. I think we can leave that for #2063.

stenington:
Yeah, the thing I like about them not being merged right now is that a `Certificate` should only exist if the actual pdf file exists, I think. Whereas `CertificateRecipient`s can exist before the file has been generated. Merging those will make it harder to understand and track the state of things, I think. Maybe not.

stenington:
I'm happy with the PR as is, so tagging you for review @adrichey

adrichey:
> Yeah, the thing I like about them not being merged right now is that a `Certificate` should only exist if the actual pdf file exists, I think. Whereas `CertificateRecipient`s can exist before the file has been generated. Merging those will make it harder to understand and track the state of things, I think. Maybe not.

Totally makes sense. Reviewing now.

Issue #2076, opened 2019-04-01
Rework drop lowest scores task


stenington:
During semifinals we plan to drop the lowest score for each team, but the dropped score should not count against that judges judged submission total when it comes to certificate levels. Previously we would try to detect that scenario in the rake task and set a certificate override. I think we should move to a solution where we add a `dropped` field or something to that effect that the rake task can set (while also soft deleting the score), and then the judge certificate level can query for scores including deleted and dropped scores.

adrichey:
I think this sounds like a sound approach. Tracking dropped scores will be really useful for data collection, so I like this approach.

adrichey:
@stenington What should the QA steps look like for this? I think it is a little hard to test this one until we are ready to actually drop the scores on QA. Unless we want to backup the database, run this and test, and then revert back. But that takes a good amount of coordination. I did thorough testing on my end and things seemed to look okay. Do you maybe want to test this and assert that lowest scores for semifinals get marked appropriately, but still appear in the certificate calculations? I think it might be easier for you to confirm this one, but I don't know how biased that makes the testing since you have seen the code as well.

stenington:
Yeah, I'll test locally and approve.

adrichey:
This is now ready for testing. Please note that this will be tested by Mike as we need a developer to run the task and approve/deny it works since it runs in the background.

stenington:
This passes QA. Before using in production, I'll do a trial run with a copy of the db and do some manual sanity checking with the program team to be safe.

PR #2083, opened 2019-04-03
#2076 Rework drop lowest scores rake task

adrichey:
`JudgeProfile.complete_scores` now takes dropped scores into account when calculating the total. I wasn't sure if I would need to address `JudgeProfile.current_completed_scores` during this work. I am also a little unsure if we have enough coverage around this functionality, but I don't know where I should add it if we don't. So if that is necessary, a little guidance would be appreciated. :)

Addresses #2076.

### DIFF
--- a/app/models/judge_profile.rb
+++ b/app/models/judge_profile.rb
@@ -82,7 +82,7 @@ class JudgeProfile < ActiveRecord::Base
   has_many :current_complete_scores, -> { current.complete },
    class_name: "SubmissionScore"
 
-  has_many :completed_scores, -> { complete },
+  has_many :completed_scores, -> { complete_with_dropped },
    class_name: "SubmissionScore"
 
   has_many :current_completed_scores, -> { current.complete },
--- a/app/models/judge_profile.rb
+++ b/app/models/judge_profile.rb
@@ -85,7 +85,7 @@ class JudgeProfile < ActiveRecord::Base
   has_many :completed_scores, -> { complete_with_dropped },
    class_name: "SubmissionScore"
 
-  has_many :current_completed_scores, -> { current.complete },
+  has_many :current_completed_scores, -> { current.complete_with_dropped },
    class_name: "SubmissionScore"
 
   has_many :current_quarterfinals_complete_scores,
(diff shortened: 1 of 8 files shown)


adrichey:
@stenington I think this about covers what was listed in the ticket, but I cannot help but feel like I am missing something here.

app/models/judge_profile.rb
```diff
@@ -82,7 +82,7 @@ class JudgeProfile < ActiveRecord::Base
   has_many :current_complete_scores, -> { current.complete },
    class_name: "SubmissionScore"
 
-  has_many :completed_scores, -> { complete },
+  has_many :completed_scores, -> { complete_with_dropped },
    class_name: "SubmissionScore"
 
   has_many :current_completed_scores, -> { current.complete },
```
stenington: `:current_completed_scores` should probably also use `complete_with_dropped`

lib/drop_lowest_scores.rb
```diff
@@ -25,6 +28,8 @@ def self.call(submission, logger_opt = nil)
       logger.info minimum_score.total
 
       submission.lowest_score_dropped!
+      minimum_score.drop_score!
+      minimum_score.save
```
stenington: This `save` should be unnecessary, the `update` in `drop_score!` should already handle that. I would also probably call `destroy` inside `drop_score!` and not call it here. To me `drop_score!` means "delete while also marking as dropped".

spec/lib/drop_lowest_scores_spec.rb
```diff
@@ -4,7 +4,14 @@
 
 RSpec.describe DropLowestScores do
   it "deletes the lowest score " do
-    lowest_score = double(:Score, total: 45, destroy: true, judge_profile: NullProfile.new)
+    lowest_score = double(
+      :Score,
+      total: 45,
+      destroy: true,
+      save: true,
+      :drop_score! => true,
```
stenington: `double` is new to me, but I think this test should check that `lowest_score` recieves `drop_score!` down on line 33 if you make the other suggested changes.

adrichey:
All right, @stenington, I made the changes you asked for, so I think this is ready for review again. :)

PR #2084, opened 2019-04-04
#2063 add a sidekiq default_retries_exhausted handler to update Job status

stenington:


config/initializers/sidekiq.rb
```diff
@@ -17,4 +17,19 @@
       chain.add Sidekiq::Middleware::Server::Profiler
     end
   end
+
+  config.default_retries_exhausted = -> (msg, ex) do
+    job_id = msg.dig('args', 0, 'job_id')
+
+    unless job_id
+      Sidekiq.logger.warn "Unable to get job_id in default_retries_exhausted for #{msg}: #{ex}"
```
stenington: Not really sure if the `msg` structure is always the same, so just being defensive.

config/initializers/sidekiq.rb
```diff
@@ -17,4 +17,19 @@
...
+      return
+    end
+
+    if db_job = Job.find_by(job_id: job_id)
+      db_job.update_column(:status, "dead")
+      Sidekiq.logger.warn "Job id=#{db_job.id} job_id=#{job_id} for " +
+                          "#{db_job.owner_type} id=#{db_job.owner_id} is dead"
```
stenington: Maybe logging a warning is useful? If this shows up in Papertrail we could conceivably set up alerts and have some sense of whose jobs have failed.

stenington:
This one is a little hard to test, but locally I raise an Error in ExportJob and then run the app with Sidekiq and it seems to work.

adrichey:
> This one is a little hard to test, but locally I raise an Error in ExportJob and then run the app with Sidekiq and it seems to work.

Sounds like this might be a little difficult to test for our QA folks. Should this ticket be "close on approval of pull request" then? I can test it by raising an error as you did while I review the code in any case.

stenington:
Yeah that sounds good to me.

The setup I had to change was:
* `ACTIVE_JOB_BACKEND=sidekiq` in `.env`
* run `bundle exec sidekiq -v -e development -q default` alongside `bundle exec rails s` in another terminal tab
* `raise Exception.new("kaboom")` on line 63 of `app/jobs/export_job.rb`

Then in the terminal tab running Sidekiq you can see the retries and the ultimate logging for the dead queue.

Issue #2050, opened 2019-03-21
Confirm/create rake task for setting semifinalists


stenington:
The idea here is that the program team would provide us a spreadsheet listing all of the teams/submissions that are moving on to the finals, and we'd have a rake task to read that file and update each submission's contest rank in the database.

Step 1 is probably figuring out what that spreadsheet might look like, and whether it has team IDs or submission IDs. 

The task should **complain very loudly** if it encounters any problem along the way, e.g. can't find an id. It might be nice for it to generate a human-readable list of the teams/submissions it updated as well for double-checking, although I believe we also plan to have people double-check through the admin.

adrichey:
I went ahead and tested this ticket since it does not have an associated UI element. It is working as intended. Closing the ticket.

PR #2086, opened 2019-04-05
#2050 add task for setting semifinalists

stenington:
New rake tasks:
* `rake judging:check_semifinalists[path/to/file.csv]`
* `rake judging:set_semifinalists[path/to/file.csv]`

Both expect a csv file with a header row, and require the following two columns:
* `team_id`: id of quarterfinalist team that will be moved to semifinals
* `team_submission_id`: the id of their submission

`check_semifinalists` will only check the file for validity, while `set_semifinalists` will first check the file for validity, and then run again making the actual changes in the database.

The tasks assume the csv includes **all semifinalists**, so if semifinalists are found in the database that aren't included in the csv, the task will fail.

### DIFF
--- a/lib/set_semifinalists.rb
+++ b/lib/set_semifinalists.rb
@@ -1,0 +1,98 @@
+class SetSemifinalists
+  private
+  attr_reader :filename
+
+  public
+  attr_reader :logger
+
+  def initialize(filename)
+    @filename = filename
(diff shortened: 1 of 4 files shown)


adrichey:
@stenington I am going to go ahead and merge this into QA now.

PR #2090, opened 2019-04-09
Convert CERTIFICATE_TYPES constant over to hashes

adrichey:
After this gets merged in, we need to remove `Certificates` with IDs greater than 240 from QA as they contain incorrect indexes.

### DIFF
--- a/app/technovation/determine_certificates.rb
+++ b/app/technovation/determine_certificates.rb
@@ -18,7 +18,7 @@ class DetermineCertificates
   def needed
     needed = []
 
-    eligible_types.each do |certificate_type|
+    eligible_types.select do |certificate_type|
       needed += needed_recipients(certificate_type)
     end
 
--- a/app/technovation/determine_certificates.rb
+++ b/app/technovation/determine_certificates.rb
@@ -10,9 +10,11 @@ class DetermineCertificates
   end
 
   def eligible_types
-    CERTIFICATE_TYPES.select do |certificate_type|
+    types = CERTIFICATE_TYPES.select do |certificate_type|
       gets_certificate?(certificate_type)
     end
+
+    types.keys.map { |type| type.to_s }
   end
 
   def needed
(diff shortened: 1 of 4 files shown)


adrichey:
This is a WIP. Will tag when complete.

adrichey:
@stenington I think this is ready for review. I tried to do thorough testing with both our automated tests and confirmation using database data on QA. I made sure that the eligible certificates made sense and that I was able to generate, view, and delete certificates as well to confirm that our new code isn't breaking with these changes. I think we should be good to go. Hopefully I didn't miss anything. Please let me know if anything needs tweaking and I will try to fix it ASAP so we can get tickets retested and a release cut to production soon.

app/technovation/determine_certificates.rb
```diff
@@ -10,15 +10,17 @@ def initialize(account, options = {})
   end
 
   def eligible_types
-    CERTIFICATE_TYPES.select do |certificate_type|
+    types = CERTIFICATE_TYPES.select do |certificate_type|
```
stenington: I would consider rewriting this as either 

```
CERTIFICATE_TYPES.keys.select { |certificate_type|
    gets_certificate?(certificate_type)
}.map { |certificate_type|
    certificate_type.to_s
}
```

or

```
types = CERTIFICATE_TYPES.keys.select do |certificate_type|
    gets_certificate?(certificate_type)
end

types.map(&:to_s)
```

I like `.keys` being attached to `CERTIFICATE_TYPES` so we know up front we only care about the keys, not the values here.

app/technovation/determine_certificates.rb
```diff
@@ -31,9 +33,11 @@ def season
   end
 
   def eligible_certificate_types
```
stenington: I think this method is actually unused, I think it was leftover from when I was working on this whole class. I would remove it.

app/technovation/determine_certificates.rb
adrichey: Makes sense. Thanks for making that change!

app/technovation/determine_certificates.rb
adrichey: I didn't even think to check that. Thanks!

adrichey:
@stenington Thanks for making the tweaks to this. As soon as this passes CI, I will merge that into QA and delete the certificates in the system that have the incorrect `cert_type` attached to them.

PR #2092, opened 2019-04-10
#2059 make sure not to award current mentors who were past students student…

stenington:
… certs any more


adrichey:
@stenington It looks like #2059 got tested and approved, so now that this is in place, should we have them retest that before deploying?

stenington:
Nah, I'll just confirm that one myself. I'll note it in https://github.com/Iridescent-CM/technovation-app/issues/2059

Issue #2043, opened 2019-03-15
Printable event page can show judges no longer in the event


stenington:
**Describe the bug**
The RA events view and printable view can show a different set of judges for an event.

**To Reproduce**
1. Login as Neil Degrasse Tyson
2. Go to events
3. Look at Judge Test Cap, note the 3 teams and 1 judge
4. Click the "Print for live event" icon
5. See only 2 teams and 4 judges

**Expected behavior**
@ophelieh @kmorton2017 Can you help fill in the expected behavior here? I assume I should see all partial and complete scores only related to the teams and judges currently assigned to the event, right?

**Additional context**
I assume a big factor here is that our staging data has gotten very messy. Hopefully in production we won't have quite the craziness with teams and judges being assigned, unassigned, scored, assigned again, etc. But nevertheless I think the printable view should be able to handle some of these potential scenarios.

ophelieh:
Expected behavior - As an RA, I only see judges currently assigned to my event. I should also see their partial or completed scores.

adrichey:
@ophelieh and @kmorton2017 I have a question about this. For the particular event that is listed in the description, we have various other judges that have judged the teams at some point due to the ever-shifting way we do our QA, so if you look at the Scores section for that RA, we get results that don't match the printable view if I put a fix in place.

Print out:
[printable_scores.pdf](https://github.com/Iridescent-CM/technovation-app/files/2997193/printable_scores.pdf)

Actual scores:
https://technovation-qa.herokuapp.com/regional_ambassador/score_details/83#/
https://technovation-qa.herokuapp.com/regional_ambassador/score_details/88#/

What do we want to do about this? Is there anything to be done? Can you give me more information on how we handled this in the past? I have a fix for this ticket, but I want to make sure I have a full picture of this before I submit my work.

ophelieh:
Should we create a new event to get some better data? @kmorton2017 Could you create an event, add some teams and have Nick and Kai score some submissions? At least 1 partial and 2 complete would be good. 

Also, I noticed that there are a lot of blank pages on the PDF, is that normal?

adrichey:
The blank pages is not normal, I planned on tackling that, I just wanted to get this in front of you both so you could provide feedback on the scenarios I mentioned previously. Thanks! 🙂

ophelieh:
Perfect. Thank you!

adrichey:
[...]

PR #2095, opened 2019-04-11
#2043 Refactor RPE print roster into something more in line with desired output

adrichey:


stenington:
@adrichey It seems to me like we can collapse the logic in `printable_scores_controller.rb` down quite a bit, I think it could be reduced to

```
def show
      @event = RegionalPitchEvent.includes(
        judges: :submission_scores,
        team_submissions: :submission_scores
      ).find(params[:id])

      judges = @event.judges

      judge_ids = judges.map { |judge| judge.id }
      @scores = @event.team_submissions
                .flat_map(&:scores)
                .reject { |score| !judge_ids.include?(score.judge_profile_id) }

      @groups = []
      judges.each.with_index do |judge, idx|
        teams = (judge.assigned_teams_for_event(@event) || @event.teams).to_a

        if index = @groups.find_index { |group| group[:teams] == teams }
          @groups[index][:judges].push(judge)
        else
          @groups.push({ judges: [judge], teams: teams })
        end
      end
    end
```

`assigned_teams_for_event` could even return `event.teams` if no assignments are otherwise found, since that seems to be the default.

adrichey:
> @adrichey It seems to me like we can collapse the logic in `printable_scores_controller.rb` down quite a bit, I think it could be reduced to
> 
> ```
> def show
>       @event = RegionalPitchEvent.includes(
>         judges: :submission_scores,
>         team_submissions: :submission_scores
>       ).find(params[:id])
> 
>       judges = @event.judges
> 
>       judge_ids = judges.map { |judge| judge.id }
>       @scores = @event.team_submissions
>                 .flat_map(&:scores)
>                 .reject { |score| !judge_ids.include?(score.judge_profile_id) }
> 
>       @groups = []
>       judges.each.with_index do |judge, idx|
>         teams = (judge.assigned_teams_for_event(@event) || @event.teams).to_a
> 
>         if index = @groups.find_index { |group| group[:teams] == teams }
>           @groups[index][:judges].push(judge)
>         else
>           @groups.push({ judges: [judge], teams: teams })
>         end
>       end
>     end
> ```
> 
> `assigned_teams_for_event` could even return `event.teams` if no assignments are otherwise found, since that seems to be the default.

This is really clean. I believe you are right. I will make the change and run a test and then push that change up.

adrichey:
@stenington This is once again ready for review.

app/controllers/regional_ambassador/printable_scores_controller.rb
```diff
@@ -9,34 +9,26 @@ def show
...
-            j.submission_scores.where(team_submission_id: event_submission_ids)
-          }
+      judge_ids = judges.map { |judge| judge.id }
+      @scores = @event.team_submissions
+                .flat_map(&:scores)
+                .reject { |score| !judge_ids.include?(score.judge_profile_id) }
+
+      @groups = []
+      judges.each.with_index do |judge, idx|
+        if judge.assigned_teams_for_event(@event).any?
```
stenington: Ah, good catch. I guess this is another argument for `assigned_teams_for_event` defaulting to all the event's teams, but I'm happy with this as is.

Issue #2239, opened 2019-10-03
Admin can not change email address for mentors [5?,?,?]


rsgonzal:
As an admin, staging isn't letting us change the email address for mentors. We're clicking save changes but it doesn't record. Stays on the same screen. When changing student addresses it goes back to the previous page and is recorded.

hellafitz:
next step: @rsgonzal confirm if this true on production as well. If yes, check w/ Program team on how painful this bug is (if they've even noticed).

hellafitz:
@rsgonzal please update

rsgonzal:
This is happening on production as well. Checking in with program team to check severity.

Get this screen when I try to edit:
![Screen Shot 2019-10-23 at 3 28 44 PM](https://user-images.githubusercontent.com/22801026/67439081-21185900-f5aa-11e9-9810-308f2d05d9fc.png)

rsgonzal:
Team does not see this as an issue. They usually ask mentors to reset their own passwords

sha11n:
I managed to recreate this bug locally. I think this is what's going on: the "Mentor Type" field is silently failing a validation check; so the page is being re-rendered. The error isn't being displayed b/c the "Mentor Type" field isn't setup correctly. Also, because it's not setup correctly, the drop-down isn't pre-selecting the value that is already set. :confused:

As a work-around, I found that you can select the value for "Mentor Type" that was already selected/saved (this value can be viewed on the previous page), which will pass the validation check, which will then allow the new email address to be saved. Let me know if this makes sense, and if it works for you, @rsgonzal. :smiley:

Lastly, the PR to address this is here: https://github.com/Iridescent-CM/technovation-app/pull/2289. Let me know if this fixes the issue. :smile:

sha11n:
While looking at this bug and the code more (w/ about 0.4% more understanding of how everything works :blush: ), from what I'm experiencing locally, none of the form fields for the mentor's account (first name, last name, email, dob, gender) will be updated while "Mentor Type" is left as "Choose one...". Also, when I select a new 'Mentor Type', the new value isn't being saved. I wonder if it's b/c I don't have my local environment setup completely, or it could be b/c I'm doing it wrong! :joy: :rofl: Anyhow, fwiw, the PR i opened #2289 fixed both of these issues locally for me. :man_shrugging: :smiley:

stenington:
[...]

PR #2289, opened 2019-11-09
Fix "Mentor Type" select field

sha11n:
This should default/pre-populate the "Mentor Type" field to the value that has already been set. It should display any error messages for "Mentor Type" as well.

Refs: #2239

### DIFF
--- a/app/views/admin/participants/edit.html.erb
+++ b/app/views/admin/participants/edit.html.erb
@@ -42,9 +42,16 @@
         <p>
           <%= m.label :mentor_type %>
 
+          <% if f.object.mentor_profile.errors[:mentor_type].present? %>
+            <div class="field_with_errors">
+              <span class="error"><%= f.object.mentor_profile.errors[:mentor_type].to_sentence %></span>
+            </div>
+          <% end %>
+
           <%= m.select :mentor_type,
             MentorProfile.mentor_types.keys,
-            prompt: "Choose one..." %>
+            prompt: "Choose one...",
+            selected: f.object.mentor_profile.mentor_type %>
         </p>
       <% end %>
     <% end %>


app/models/account.rb
```diff
@@ -118,7 +118,7 @@ class Account < ActiveRecord::Base
     dependent: :destroy
 
   has_one :background_check, dependent: :destroy
-  accepts_nested_attributes_for :background_check
+  accepts_nested_attributes_for :background_check, :mentor_profile
```
sha11n: This change could have also been added below/around line 37, where the `mentor_profile` relation is setup. Let me know if I should move it there, or somewhere else. :smiley: 

Also, it took an embarrassingly long time for me to find/fix this, and to make this simple change. :blush: :grimacing:  :sweat_smile:

app/controllers/admin/participants_controller.rb
```diff
@@ -31,6 +31,10 @@ def update
         id: params.fetch(:id),
       }.merge(account_params)
 
+      if @account.mentor_profile.present? && params.dig(:account, :mentor_profile_attributes).present?
+        profile_params[:mentor_type] = params.dig(:account, :mentor_profile_attributes, :mentor_type)
+      end
```
sha11n: Maybe there's a better way to do this logic/part? The `@account.mentor_profile.present?` check/part is pretty much a mirror of what's in the view (`views/admin/participants/edit.html.erb`), and the  `params` check seems necessary (maybe it's not?) because only the mentor profile has the `mentor_type` field. :thinking: Also, it doesn't look like `profile_params` is setup as/to use "strong parameters", maybe because there was only one field (now two)? :man_shrugging:

stenington:
Interesting, thanks @sha11n! Let me find some time to review this this week.

app/models/account.rb
sha11n: Thanks for the feedback, @stenington! I didn't know what I was getting myself into! :blush: :grinning: 

Sounds like this change isn't ideal, I'm about to push a commit to remove it. :+1: 

Some random things/findings:

I originally added the `accepts_nested_attribute_for :mentor_type` to use the `f.fields_for :mentor_profile` in `views/admin/participants/edit.html.erb`. Adding `accepts_nested_attribute_for :mentor_type` is what led me to update`participants_controller.rb`, I'll be removing those changes as well.

After I backed out all my changes locally I looked again at the "Mentor Type" select field. Without any of these changes, and when a value is selected, it seems to save/update successfully. The root issue seems to be that it's not defaulting/pre-populating (to whats already been set) correctly. I fixed this locally, I'll push it up for you to look at. :sweat_smile:

app/views/admin/participants/edit.html.erb
```diff
@@ -44,7 +44,7 @@
 
           <%= m.select :mentor_type,
             MentorProfile.mentor_types.keys,
-            prompt: "Choose one..." %>
+            selected: f.object.mentor_profile.mentor_type %>
```
sha11n: This will address the "Mentor Type" field not defaulting to what has already been set. This seems to be the root issue. 

I removed the "Choose one..." prompt because this is a required field, and the form can't be saved without selecting a value. This may be a conundrum though, if there are any mentors that don't have a "Mentor Type" set, and an admin updates their account, it will pick/save the first one in the list for them. Someone (not me b/c I don't have access) can run a query to see if there are any mentors who don't have a "Mentor Type" set; if there are, and we leave the "Choose one..." option there, and an admin tries to update a mentor **without** a "Mentor Type", the form won't be able to save b/c "Mentor Type" is a required field, which means they wouldn't be able to update the mentor w/o picking a "Mentor Type". So, it seems like it *could* be an issue either way? :face_with_head_bandage:  :dizzy_face: :dizzy: :joy: :smiley:

app/views/admin/participants/edit.html.erb
```diff
@@ -42,9 +42,16 @@
...
+          <% if f.object.mentor_profile.errors[:mentor_type].present? %>
+            <div class="field_with_errors">
+              <span class="error"><%= f.object.mentor_profile.errors[:mentor_type].to_sentence %></span>
+            </div>
+          <% end %>
+
           <%= m.select :mentor_type,
             MentorProfile.mentor_types.keys,
-            prompt: "Choose one..." %>
+            prompt: 'Choose one..',
```
sha11n: By the way, just FYI (I got stumped by this) the "Choose one.." prompt will only display if there isn't a mentor type set. If we want the "Choose one..." prompt to always be visible, looks like we need to use `include_blank` instead.

[https://apidock.com/rails/ActionView/Helpers/FormOptionsHelper/select#271-Prompt-vs-Select](https://apidock.com/rails/ActionView/Helpers/FormOptionsHelper/select#271-Prompt-vs-Select)

app/views/admin/participants/edit.html.erb
```diff
@@ -42,9 +42,16 @@
...
+          <% if f.object.mentor_profile.errors[:mentor_type].present? %>
+            <div class="field_with_errors">
+              <span class="error"><%= f.object.mentor_profile.errors[:mentor_type].to_sentence %></span>
+            </div>
+          <% end %>
+
           <%= m.select :mentor_type,
             MentorProfile.mentor_types.keys,
-            prompt: "Choose one..." %>
+            prompt: "Choose one...",
```
sha11n: Btw, just FYI (I got stumped by this) the "Choose one..." prompt will only display if a mentor doesn't have a mentor type set. Looks like we need to use `include_blank` if we always want it to be visible.

[https://apidock.com/rails/ActionView/Helpers/FormOptionsHelper/select#271-Prompt-vs-Select](https://apidock.com/rails/ActionView/Helpers/FormOptionsHelper/select#271-Prompt-vs-Select)

PR #2311, opened 2019-11-22
upgrade to rails 5.2

stenington:
This hopefully addresses some of the security concerns in the `Gemfile.lock` manifest for the time being. Part 2 would be addressing the issues identified in `yarn.lock`.


stenington:
It seems like `datagrid` isn't 5.2 compatible yet.

Issue #2257, opened 2019-10-11
Remove technical checklist from scoring [8,3,2.67]


stenington:
From #2192, remove the technical checklist from submission scoring without removing it from the UI.

stenington:
This will be folded into #2259

PR #2319, opened 2019-12-04
#2257 remove technical checklist from overall score calculation

stenington:


app/controllers/judge/scores_controller.rb
```diff
@@ -127,6 +127,7 @@ def score_params
         :ideation_comment_word_count,
 
         :app_functional,
+        :demo,
```
stenington: This was removed as part of #2307, and it caused a subtle bug where it appeared that judge scores were calculating correctly, but the value of the `demo` score was always ultimately stored as 0. Calling it out in case there's a way to be more preventative, either with tests or with noisier failure somewhere.

app/models/submission_score.rb
```diff
@@ -385,9 +380,9 @@ def overall_total
   def total_possible
     case team_submission.team_division_name
     when "junior"
-      80
+      70
```
stenington: It would be nice if these magic numbers were gone.

app/models/team_submission.rb
```diff
@@ -264,7 +264,7 @@ def qualifies_for_participation?
   end
 
   def total_possible_score
-    senior_division? ? 100 : 80
```
stenington: More magic numbers.

app/views/regional_ambassador/score_details/show.en.html.erb
```diff
@@ -18,7 +18,7 @@
     <dt>Quarterfinals average:</dt>
     <dd>
       <%= @submission.quarterfinals_average_score %> /
-      <%= @submission.senior_division? ? 100 : 80 %>
+      <%= @submission.senior_division? ? 90 : 70 %>
```
stenington: Magic numbers.

app/controllers/judge/scores_controller.rb
stenington: Opened #2320 for dealing with this, removing the fix from this PR

Issue #2261, opened 2019-10-11
Checklist removal final sweep [8,8,1]


stenington:
From #2192, do a final sweep on removing the checklist after the other tickets are done:
* migrations to remove things from the DB
* if any libraries or config can be remove, do so
* check for rake tasks or other stray mentions

hellafitz:
@stenington if this wasn't completed, would it block us from accepting new submissions?

stenington:
No, this shouldn't be a blocker on opening submissions. The only user-facing piece in this ticket that I found was about how scores are displayed to students at the end of the season.

stenington:
QA context:

This ticket mainly removes code and templates relevant to the technical checklist that should no longer be in use anyway. I also found that the screen where students see their scores needed updating to not mention the checklist anymore.

Suggested QA:

Run through the process of creating a submission once more, and make sure you can throw together a submission and submit it as a student. If the site throws any errors along the way, report them. Otherwise we'll assume things are fine.

Also, check the student scores view by
1. turning on Scores and Certificates
2. log in as a student who submitted and received official scores
3. navigate to the scores screen and view your scores
4. make sure the technical checklist isn't mentioned and the score looks correct

rsgonzal:
Everything is working as expected. No issues when creating a submission or viewing scores. Just the known error of the technical score not recording.

stenington:
@rsgonzal Is this ready to deploy or have you asked someone else to take a look?

nickiredescent:
If you need another test,
Same as Roxanna, things have worked like expected and I do not see the Technical checklist mentioned as a student.

stenington:
Awesome, thanks @nickiredescent!

PR #2344, opened 2019-12-20
#2261 clear out technical checklist and code checklist

stenington:
This is attempting to clear out the code and assets associated with the technical checklists, which are no longer accessible through the UI. It includes a migration to drop the table, and removes tests. It also updates the student score screen, which wasn't otherwise caught.


shaunxp20:
FWIW I pulled this branch down locally, and it's looking good to me! I couldn't find any other instances of the technical checklist, so thorough, so many red lines! 🙌 💯 ☑️ 🙅‍♂ 

One thing I did notice though, which I don't think is related to this PR, is that after I ran the migration, `structure.sql` had a change in it:
```
-SET default_with_oids = false;
+SET default_table_access_method = heap;
```
From my initial web search on it, I think this is just a side effect of using a newer version of PostgreSQL (I'm using version 12.1). 🤔 I'll try downgrading and see if I can verify that.

stenington:
> side effect of using a newer version of PostgreSQL

Yeah, that seems plausible. I'm not 100% clear on the purpose of the `structure.sql` file, honestly. I've seen small diffs like this crop up occasionally and ignored them, not sure if that will come back to bite me one day but so far it hasn't.

shaunxp20:
> > side effect of using a newer version of PostgreSQL
> 
> Yeah, that seems plausible. I'm not 100% clear on the purpose of the `structure.sql` file, honestly. I've seen small diffs like this crop up occasionally and ignored them, not sure if that will come back to bite me one day but so far it hasn't.

:+1: I was having some difficulties running the migrations on an older version of PostgreSQL, and then the settings page bug came in; I abandoned the downgrade idea in order to get my local env up and running. Anywho, long story short, sounds like we're good. 😅 😀

Issue #2345, opened 2020-01-09
Content & Settings tab not loading on TG production


rsgonzal:
**Describe the bug**
The content & settings tab on TG Production site is not loading.

**To Reproduce**
1. Go to http://my.technovationchallenge.org/
2. Click on 'Content & Settings'
3. See nothing loaded

**Expected behavior**
I expect to see all of the toggle controls for registration, submissions, surveys, etc.

**Screenshots**
![Screen Shot 2020-01-08 at 3 25 12 PM](https://user-images.githubusercontent.com/22801026/72096341-e1226180-32ce-11ea-9804-a691760ef10e.png)

**Additional context**
Tried with multiple accounts and computers and seeing the same error. It was working around 1/8 12 EST when we put up the pre-surveys

stenington:
The issue is coming from the way settings data is passed into the Vue app for that page. If you look [here](https://github.com/Iridescent-CM/technovation-app/blob/44790be23e1ca65099fbe5543037ddc67516d199/app/views/admin/season_schedule_settings/edit.html.erb#L16) you can see we're putting an object literal into an HTML element attribute, pulling in various season toggle values that the Vue app needs. The issue is that strings are being quoted with single quotes, but not being escaped so that if the string itself contains a single quote, the page has invalid HTML, Vue blows up, and the Vue app doesn't load.

shaunxp20:
Wow, great find @stenington ❕🥇 💯

stenington:
QA steps:

1. log in as an admin
2. go to **Content & Settings**
3. Under **Notices**, set all notices to something with a `'` in it, like "Let's test"
4. Under **Surveys**, set all 6 fields to something with a `'`
5. Click **Review**, then **Save these settings**
6. Click **Content & Settings** again, and make sure the settings controls load

stenington:
@shaunxp20 do those QA steps seem reasonable to you?

shaunxp20:
> @shaunxp20 do those QA steps seem reasonable to you?

QA steps look good to me. Thanks for adding those! 🙌 I verified them locally too. 👍 🚥 🍏 📗 💚

stenington:
I confirmed the steps above on staging, but it would be great for one of you to do the same, @rsgonzal or @shaunxp20.

shaunxp20:
> I confirmed the steps above on staging, but it would be great for one of you to do the same, @rsgonzal or @shaunxp20.

[...]

PR #2346, opened 2020-01-09
#2345 fix single quote bug in content and settings page

stenington:
I don't think this is a particularly elegant solution, but it gets the job done for now. There seems to be a double-quote problem as well, though it doesn't prevent the Vue app from loading.


stenington:
@shaunxp20 Want to take a quick look at this one?

app/helpers/vue_helper.rb
```diff
@@ -0,0 +1,9 @@
+module VueHelper
```
shaunxp20: I was able to reproduce the error locally, by hard-coding the return value of `#dashboard_text` to `"'blah"`. I verified this branch/fix fixes it! 👍 🙌 

A couple of small things (no blockers, just throwing some random thoughts out there): 
 * it would be nice to have some tests around this
 * it looks like this will just replace the first single quote, I can still break it w/ the string `"'blah'blah"` (not sure how likely this is, just throwing it out there)
 * this seems more like a string helper than a Vue helper 🤷‍♂ 

Anyhow, nice find, great work! 👍 🥇 🍰 🎉 😄

stenington: Thanks @shaunxp20! I take your point about it being a string helper, I think I was trying to highlight that the greater context for cramming an object literal into an HTML element attribute was for Vue, but I think you're right.

Good catch on multiple single quotes, I'll change that too.

What kind of tests do you want to see? Testing the helper method directly seems easy enough. Would you want to see some sort of integration test as well?

shaunxp20: Nice! Changes look good to me, thanks for working on those! 🙌 I think at least testing the helper method would be good and an easy win, imho. 🤷‍♂ 😄 ✌️

stenington: @shaunxp20 Let me know what you think of the tests I just added!

shaunxp20: Tests look good to me! 🎉 💯 I verified the fix again too! Everything looks good to me! 👍 👍

Issue #2107, opened 2019-04-19
RAs should not be able to switch to judge [8,2,4]


stenington:
> An RA account is unable to switch into judge mode. The flow is RA account> click mentor mode> click "switch to judge mode"> "Uh oh, something went wrong". 

It appears that we explicitly don't auto-create judge accounts for RAs, but then the link to switch doesn't work unless they already have a judge profile.

stenington:
@ophelieh @kmorton2017 I wrote this one up on behalf of Mackenzie last week, but I'm not 100% clear on the request here. Do we need a way for RAs to switch to judge mode in their account? Are there any special considerations I need to know about with RAs being judges, or do you think the system handles that already?

stenington:
We will handle individually cases manually for now, this ticket is going in the icebox as a reminder for a long-term fix to be prioritized later.

stenington:
The current recommendation is that RAs should never be judges, and we would enforce that on the platform and message it out to participants.

stenington:
I think the work here just amounts to hiding the "Switch to judge" button if `current_mentor.is_an_ambassador?` is true.

stenington:
It looks like RAs logged in to their RA dashboards always have a "Mentor mode" link, and optionally have a "Judge mode" link **only if they have a judge profile**. The problem was that any RA, even one without a judge profile, could switch to their mentor dashboard and then see a "Switch to Judge mode" link, and try to become a judge that way.

This year, the decision is that RAs shouldn't judge at all. This ticket focuses on the RA > Mentor > Judge path, but I wonder if we want to hide the "Judge mode" link from the RA dashboard as well, just in case some RAs already have judge profiles? @rsgonzal or @ophelieh what do you think?

stenington:
Looks like 55 current RAs also have judge profiles, so I'm guessing we should disable that button so they cannot switch to judge mode this year.

ophelieh:
@stenington That works for me!

rsgonzal:
Looks like Ophelie approved! So all good :)

stenington:
## Context for QA:

There are two relevant pieces of config: `ENABLE_SWITCH_TO_JUDGE` and `ENABLE_RA_SWITCH_TO_JUDGE`. 

When `ENABLE_SWITCH_TO_JUDGE` is turned on...
* mentors (who are not also RAs) should be able to switch to the judge dashboard from the mentor dashboard
[...]

Issue #2321, opened 2019-12-05
RA infinite redirect


stenington:
**Describe the bug**
I don't know if this happens in production, but it has happened multiple times locally. When I try to log in as `ra@ra.com`, I sometimes get an infinite redirect error page.

```
Started GET "/regional_ambassador/location_details?return_to=%2Fregional_ambassador%2Fdashboard" for 127.0.0.1 at 2019-12-05 12:17:47 -0500
[SPOOFING IP] 200.94.199.219
Processing by RegionalAmbassador::LocationDetailsController#show as HTML
  Parameters: {"return_to"=>"/regional_ambassador/dashboard"}
  Account Load (0.5ms)  SELECT  "accounts".* FROM "accounts" WHERE "accounts"."deleted_at" IS NULL AND "accounts"."auth_token" = $1 LIMIT $2  [["auth_token", "LoMay2ShhLhAbuY5F7L1buHu"], ["LIMIT", 1]]
  SignupAttempt Load (0.3ms)  SELECT  "signup_attempts".* FROM "signup_attempts" WHERE "signup_attempts"."account_id" = $1 LIMIT $2  [["account_id", 6], ["LIMIT", 1]]
  RegionalAmbassadorProfile Load (0.3ms)  SELECT  "regional_ambassador_profiles".* FROM "regional_ambassador_profiles" WHERE "regional_ambassador_profiles"."account_id" = $1 LIMIT $2  [["account_id", 6], ["LIMIT", 1]]
Redirected to http://localhost:3000/regional_ambassador/profile#!location
Completed 302 Found in 57ms (ActiveRecord: 1.1ms)


Started GET "/regional_ambassador/profile" for 127.0.0.1 at 2019-12-05 12:17:47 -0500
[SPOOFING IP] 200.94.199.219
Processing by RegionalAmbassador::ProfilesController#show as HTML
  Account Load (1.1ms)  SELECT  "accounts".* FROM "accounts" WHERE "accounts"."deleted_at" IS NULL AND "accounts"."auth_token" = $1 LIMIT $2  [["auth_token", "LoMay2ShhLhAbuY5F7L1buHu"], ["LIMIT", 1]]
  SignupAttempt Load (1.3ms)  SELECT  "signup_attempts".* FROM "signup_attempts" WHERE "signup_attempts"."account_id" = $1 LIMIT $2  [["account_id", 6], ["LIMIT", 1]]
[...]

PR #2348, opened 2020-01-13
#2107 hide switch to judge link for RAs acting as mentors

stenington:
RAs have two paths to switch to judging:
1. if they already have a judge profile, a "Judge mode" link appears on the RA dashboard
2. whether they have a judge profile or not, if they switch to Mentor mode, they see a "Switch to Judge mode" link on the judge dashboard

This PR takes away the mentor dashboard link for RAs logged in as mentors, and puts the RA dashboard link behind config named `ENABLE_RA_SWITCH_TO_JUDGE` which we can leave off this year to prevent RAs from judging, I think.


app/views/mentor/dashboards/_header.en.html.erb
```diff
@@ -7,11 +7,13 @@
 <% end %>
 
 <% if ENV.fetch("ENABLE_SWITCH_TO_JUDGE", false) %>
-  <template slot="judge-switch-link">
-    <%= link_to "Switch to Judge mode",
-      judge_dashboard_path,
-      data: { turbolinks: false } %>
-  </template>
+  <% unless current_mentor.is_an_ambassador? %>
```
shaunxp20: I think I'm finally getting my head wrapped around this change/PR. 😊 😅 I don't know if this makes sense but I'll throw it out there anyway! I wonder if we should incorporate the new ENV here?

Something like:
`if ! current_mentor.is_an_ambassador? || (current_mentor.is_an_ambassador? && ENV.fetch("ENABLE_RA_SWITCH_TO_JUDGE", false))` (or something like that, but I think you get the idea)

That way this link can also be controlled by the new ENV? 🤔 🤷‍♂ 😅

app/views/regional_ambassador/_navigation.en.html.erb
```diff
@@ -51,7 +51,8 @@
     mentor_dashboard_path,
     data: { turbolinks: false } %>
 
-  <% if current_ambassador.is_a_judge? %>
+  <% if current_ambassador.is_a_judge? and
```
shaunxp20: I don't think the `and` here will do any harm, but just throwing this out there:
https://github.com/rubocop-hq/ruby-style-guide#no-and-or-or

It's been a while since I used Ruby, but I remember that was one thing we incorporated where I was/when using it before. 😄 😅

spec/features/regional_ambassador/switch_to_mentor_spec.rb
```diff
@@ -74,3 +94,33 @@
     expect(page).to have_content("You don't have permission to go there!")
   end
 end
+
+RSpec.feature "Config prevents RA switch to judge mode" do
```
shaunxp20: I wonder if it's worth adding tests for the positive case? On a side note, I tried to validate this locally, but I'm getting a bunch redirect failures when I try to login as an RA. 😖 😕 Anywho, I'll keep trying to test this out locally. 😄

stenington: Ah, the redirect failures might be this? https://github.com/Iridescent-CM/technovation-app/issues/2321

I've had that happen in dev as well, but haven't quite figured out exactly what conditions have to exist to make it happen. We have some code that redirects users to an interruption page where they enter their location if they don't have a location on file, but I'm not sure what's then sending them back away from that page.

stenington: Let me do a little more work improving tests. We're also testing judge mode in a file named mentor mode, so let me split that out and cover all the bases.

app/views/regional_ambassador/_navigation.en.html.erb
stenington: Oh, good note. I didn't know about that!

stenington:
@shaunxp20 Can you take a look at this again? Start with the tests, do these look good to you? I'm not sure if there are conventions around RSpec or feature tests or anything that I'm failing to follow, so if you see something let me know. And then let me know if you think the coverage here is better in terms of testing the positive and negative cases.

Two additional pieces of work I'm considering doing here are
1. pulling some of the link logic out in to some sort of method, like `Account#may_switch_to_judge?` or something like that so it's easier to test directly, and
2. add a test helper for setting `ENV` config more succinctly

What would you think about those improvements?

stenington:
@shaunxp20 I think I'm done tweaking this one. Adding a helper for manipulating `ENV` variables proved a little more involved than I thought. I found https://github.com/thoughtbot/climate_control if we really want to pursue it, but it has somewhat limited capabilities.

app/models/account.rb
```diff
@@ -728,6 +732,16 @@ def is_admin?
...
+  def can_switch_to_judge?
+    if is_an_ambassador?
+      !!ENV.fetch("ENABLE_RA_SWITCH_TO_JUDGE", false) && is_a_judge?
+    elsif is_a_mentor?
+      !!ENV.fetch("ENABLE_SWITCH_TO_JUDGE", false)
+    else
+      false
+    end
+  end
+
```
shaunxp20: 👍 Nice! I really like the approach you took! Good call on extracting this logic out! Makes the view code a lot nicer/cleaner too! 💯 💇‍♂ 😁

spec/features/regional_ambassador/switch_to_judge_spec.rb
```diff
@@ -0,0 +1,161 @@
...
+
+      expect(regional_ambassador.is_a_judge?).to be_falsey
+
+      click_link "Mentor Mode"
+      expect(current_path).to eq(mentor_dashboard_path)
+
+      expect(page).not_to have_link("Switch to Judge mode")
+    end
+  end
+end
```
shaunxp20: Wow! Nice test coverage! 🙌

spec/models/account_spec.rb
```diff
@@ -737,4 +737,49 @@
...
+
+  describe "#can_switch_to_judge?" do
+    describe "when config is off" do
+      before do
+        allow(ENV).to receive(:fetch).and_call_original
+        allow(ENV).to receive(:fetch).with("ENABLE_SWITCH_TO_JUDGE", any_args).and_return(false)
+        allow(ENV).to receive(:fetch).with("ENABLE_RA_SWITCH_TO_JUDGE", any_args).and_return(false)
+      end
+
+      it "allows noone" do
```
shaunxp20: Minor quibble, "no one" is two words. 🕴 🙅‍♂ 😀

spec/models/account_spec.rb
```diff
@@ -737,4 +737,49 @@
       expect(profile.reload).not_to be_onboarded
     end
   end
+
+  describe "#can_switch_to_judge?" do
```
shaunxp20: Nice coverage on this new method too! 👍

spec/features/regional_ambassador/switch_to_mentor_spec.rb
shaunxp20: > Ah, the redirect failures might be this? #2321
> 
> I've had that happen in dev as well, but haven't quite figured out exactly what conditions have to exist to make it happen. We have some code that redirects users to an interruption page where they enter their location if they don't have a location on file, but I'm not sure what's then sending them back away from that page.

Just wanted to follow up on this thread, that redirect bug/issue is exactly what I'm experiencing locally, my logs are pretty much the same as what's in #2321. 😬 😩 ☹️

Issue #2264, opened 2019-10-11
Change the judging rubric and score calculation [20,8,2.5]


stenington:
From #2245, change the judging rubric and score calculation: 

* **Ideation**
    * add 1 question
    * change text on existing  3 questions
* **Technical**
    * add 2 questions
    * change text on existing 2 questions
* **Pitch**
    * remove 2 questions
    * change text on remaining 2 questions
* **Entrepreneurship**
    * change text on existing 4 questions
* **Overall**
    * remove 3 questions
    * change text on remaining 2 questions

I believe that the rubric questions are defined in `app/models/questions.rb`, so the basic process for working with questions would be something like:
* update `questions.rb` appropriately
* update the `SubmissionScore` model, including migrations to add/remove fields as needed
* update the `ScoresController` to add/remove parameters as needed
* update RA printable scores
* update RA score details? (not sure if this is actually unreachable)

![uploads/93f4d35d-3807-48a1-a0fb-6aa69a34843c/Screen Shot 2019-09-26 at 2.52.43 AM.png](https://s3-us-west-2.amazonaws.com/prod.huboard.com/uploads%2F93f4d35d-3807-48a1-a0fb-6aa69a34843c%2FScreen+Shot+2019-09-26+at+2.52.43+AM.png)

@rsgonzal Do we have finalized copy for the new rubric already?

rsgonzal:
copy is here: https://docs.google.com/document/d/1OxmYxzSD7bBUnkWBevzuVtynAjUG2mS65FjefPFzXfg/edit

stenington:
This ticket **will** include help text changes, we're closing #2263 which would have removed that text.

stenington:
Things we will want to develop QA plans for in this ticket:
* rubric changes
    * confirm all questions are correct
    * confirm all section summaries are correct
    * confirm all section rubric links are correct
* scoring
    * confirm max scores are correct for junior and senior divisions
    * confirm you can do scoring as a judge, and your score is saved correctly
* RA/admin
    * confirm that scores in the admin display correctly when viewing score details
    * confirm the printable scores sheets for an event still display correctly

Another consideration is that this change will almost certainly break dataclips. Do we want to try to deal with that now, or fix it later?

stenington:
So maybe QA looks something like this:

### Setup
1. log in as an RA
2. create 2 events, 1 junior division, 1 senior division
3. add teams to each event and save
[...]

Issue #2320, opened 2019-12-05
Add feature specs for scoring submissions [8,8,1]


stenington:
Add features specs for:
* scoring a junior submission
* scoring a senior submission

These feature specs will help catch the original bug: `"Demo" score element not recorded"`.

The original bug was fixed in #2208 

For posterity, the original bug is below:

**Describe the bug**
When scoring a submission, whatever value you choose for the second question of the Technical section is not included in the score total.

**To Reproduce**
1. Assuming you have a judge account and a submission that that judge will be able to see...
2. Log in as a judge
3. Attempt to give the submission a perfect score (it's easiest to see the problem this way), finish the score
4. Review the score just created
5. Notice it's 5 points lower than what it had previously shown you
6. Go to the Technical section and see that it has no value for the second question.~~

**Expected behavior**
That score should be part of the total.

**Screenshots**

![score2](https://user-images.githubusercontent.com/55319/70252615-baad5b80-174f-11ea-9522-82dde5a50568.gif)

stenington:
Addressing testing around this should be part of the ticket. I'm not sure if we can drive the Vue app UI through capybara or not, but it would be nice to have a "perfect score" smoke test of some sort at least.

shaunxp20:
> Addressing testing around this should be part of the ticket. I'm not sure if we can drive the Vue app UI through capybara or not, but it would be nice to have a "perfect score" smoke test of some sort at least.

I totally agree! 💯 I'll see about adding one. 👍

PR #2351, opened 2020-01-16
#2264 updating the judging rubric

stenington:
Opening this as a WIP PR for now. I decided to make the rubric updates without restructuring the code too much, but there are some potential refactorings to make.

In `questions.rb`, the `field` attribute is starting to feel redundant when initializing `Question` objects. Similarly, the way `score` is pulled feels slightly redundant. I also considered moving a bunch of what's essentially rubric config into a YAML file (section, idx, text), but decided to postpone doing those sorts of changes.

And then methods like `ideation_total` in the `SubmissionScore` model feel a little silly with the new naming convention, but again I went for minimal update to get everything working with the new rubric.


stenington:
Incidentally, I think this PR will also fix #2321, which I encountered again while working on the rubric.

stenington:
@shaunxp20 If you have time today to review this before your travel, I'd appreciate it. I know there's room for refactoring, but I'm particularly interested if you see something you would hesitate to accept at all.

db/seeds.rb
```diff
@@ -233,6 +233,7 @@
     job_title: "Software Engineer",
     bio: "I am passionate about tech and empowering girls",
   )).valid?
+  Geocoding.perform(ra.account).with_save
```
shaunxp20: I just copied this line over locally and verified that this fixes #2321! 🙌 🏅 🍰 🎉 🎈

app/controllers/judge/scores_controller.rb
```diff
@@ -120,35 +120,34 @@ def reject_suspended_judge
 
     def score_params
       params.require(:submission_score).permit(
-        :sdg_alignment,
-        :evidence_of_problem,
-        :problem_addressed,
+        :ideation_1,
+        :ideation_2,
+        :ideation_3,
+        :ideation_4,
```
shaunxp20: 👍 I like these changes/the renaming in this PR! Imho, it aligns better with the scoring section/pages, and makes it clearer which section/page they belong to! 📏 🤓 📝 👍

app/controllers/judge/scores_controller.rb
```diff
@@ -120,35 +120,34 @@ def reject_suspended_judge
...
         :ideation_comment,
         :ideation_comment_word_count,
 
-        :app_functional,
+        :technical_1,
+        :technical_2,
```
shaunxp20: Btw, this was the bug in #2320 (question `#2` the demo score), I like how it's a little more clear here that it's the second question on the technical section. 👍 #️⃣ 2️⃣

app/javascript/judge/scores/QuestionSection.vue
```diff
@@ -3,7 +3,7 @@
     <div :class="solo ? 'grid__col-md-6' : 'grid__col-12'">
       <h3>{{ title }}</h3>
 
-      <p class="refer-to">Refer to the {{ referTo }}</p>
+      <slot name="section-summary" />
```
shaunxp20: Still getting ramped up on Vue.js, anywho this `slot` extraction looks like a nice win! 🏁 🏎

app/models/submission_score.rb
```diff
@@ -232,9 +232,9 @@ def self.total_possible
   def self.total_possible_for(division)
     case division
     when "junior"
-      70
+      60
     when "senior"
-      90
+      80
```
shaunxp20: Oh good to know! I didn't know there was a different max score depending on division! 🌈 🌟

app/models/submission_score.rb
```diff
@@ -341,38 +341,36 @@ def raw_total
   end
 
   def ideation_total
-    sdg_alignment +
-      evidence_of_problem +
-        problem_addressed
+    ideation_1 +
+      ideation_2 +
+        ideation_3 +
+          ideation_4
```
shaunxp20: I wonder how these ☝️ 👇 would look on one line, and if a `reduce` or some method extraction would make sense here. *shrug, just popping a little to me with all the indentation. #no-biggie #random-thoughts

app/views/regional_ambassador/printable_scores/show.en.html.erb
```diff
@@ -91,140 +91,63 @@
...
-              <td><%= score.evidence_of_problem %></td>
-              <td></td>
-            </tr>
-
-            <tr>
-              <td>The team addresses the problem well.</td>
-              <td><%= score.problem_addressed %></td>
-              <td></td>
-            </tr>
+            <% Questions.for(score).in_section(:ideation).each do |question| %>
```
shaunxp20: Nice clean up here! I was trying to think of an adjective to add to `score` or some other way to make it clearer that we're getting the questions associated to saved scored (if I'm understanding the code correctly), but nothing is coming to me at the moment. 😖 😵 ☹️

db/migrate/20200115211026_change_judging_rubric_fields.rb
```diff
@@ -0,0 +1,39 @@
...
+    add_column :submission_scores, :technical_3, :integer, default: 0
+    add_column :submission_scores, :technical_4, :integer, default: 0
+    add_column :submission_scores, :pitch_1, :integer, default: 0
+    add_column :submission_scores, :pitch_2, :integer, default: 0
+    add_column :submission_scores, :entrepreneurship_1, :integer, default: 0
+    add_column :submission_scores, :entrepreneurship_2, :integer, default: 0
+    add_column :submission_scores, :entrepreneurship_3, :integer, default: 0
+    add_column :submission_scores, :entrepreneurship_4, :integer, default: 0
+    add_column :submission_scores, :overall_1, :integer, default: 0
+    add_column :submission_scores, :overall_2, :integer, default: 0
```
shaunxp20: Nice @ adding the defaults! 👍

shaunxp20:
> @shaunxp20 If you have time today to review this before your travel, I'd appreciate it. I know there's room for refactoring, but I'm particularly interested if you see something you would hesitate to accept at all.

I did first-pass, the changes are looking good to me! 👍 😄

app/views/regional_ambassador/printable_scores/show.en.html.erb
stenington: Yep, you're understanding it correctly. I agree the terminology is a little murky, but don't have any suggestions either. Maybe something will arise over time as we discuss it.

db/migrate/20200115211026_change_judging_rubric_fields.rb
stenington: Yeah, this is actually required or tests fail 🤷‍♂

app/models/submission_score.rb
stenington: Yeah, it was really tempting to do something like
```
def ideation_total
  (1..4).reduce { |sum, n| sum + public_send("ideation_#{n}) }
end
```

and possibly even auto-generate those based on the section name and question counts, but I think you end up trading off brevity for ease of understanding the model. I'm leaving it as-is here, and we could continue to talk about this sort of change later if we want.

stenington: Oh, but I will add a test testing the section totals, if we want to refactor later.

PR #2352, opened 2020-01-17
2320 Allow demo scores to be saved

shaunxp20:
As we discussed on Stand, it looks like this bug will also be fixed in #2351. Anywho, there's also a fix for it in this branch/PR.

I also added a feature spec for scoring a submission.

I have two feature specs for you to look at, @stenington - the original style, and then a new/different style. A little background: it has been a while since I wrote a feature spec and stumbled upon this other way when looking at feature specs. I don't have any attachment or experience using this new style; and I'm not sure it's even popular in the Rails/Ruby world. Just thought I would give it a whirl, and wondering what you think of it. I can see pros/cons for each way.

Here are the two articles I was looking at:
https://about.futurelearn.com/blog/how-we-write-readable-feature-tests-with-rspec
https://www.madetech.com/blog/feature-testing-with-rspec

### DIFF
--- a/spec/features/judge/scoring_submissions_spec.rb
+++ b/spec/features/judge/scoring_submissions_spec.rb
@@ -1,0 +1,116 @@
+require "rails_helper"
+
+feature "scoring submissions", js: true do
+  before do
+    SeasonToggles.judging_round = :qf
+  end
+
+  after do
+    SeasonToggles.judging_round = :off
+  end
+
+  scenario "when a junior team receives a perfect score" do
+    given_there_is_a_submission_from_a_junior_team
+    and_a_user_is_logged_in_as_a_judge
+
+    when_the_judge_starts_a_new_scoring_session
+    and_scores_the_ideation_section_perfectly
+    and_scores_the_technical_section_perfectly
+    and_scores_the_pitch_section_perfectly
+    and_scores_the_overall_section_perfectly
+    and_submits_the_score
+
+    then_the_review_page_displays_a_perfect_score_of 60
+  end
+
+  scenario "when a senior team receives a perfect score" do
+    given_there_is_a_submission_from_a_senior_team
+    and_a_user_is_logged_in_as_a_judge
+
+    when_the_judge_starts_a_new_scoring_session
+    and_scores_the_ideation_section_perfectly
+    and_scores_the_technical_section_perfectly
+    and_scores_the_pitch_section_perfectly
+    and_scores_the_entrepreneurship_section_perfectly
+    and_scores_the_overall_section_perfectly
+    and_submits_the_score
+
+    then_the_review_page_displays_a_perfect_score_of 80
+  end
+  private
+
+  def given_there_is_a_submission_from_a_junior_team
+    FactoryBot.create(:team_submission, :junior, :complete)
+  end
+
+  def given_there_is_a_submission_from_a_senior_team
+    FactoryBot.create(:team_submission, :senior, :complete)
+  end
+
+  def and_a_user_is_logged_in_as_a_judge
+    judge = FactoryBot.create(:judge, :onboarded)
+
+    sign_in(judge)
+  end
+
+  def when_the_judge_starts_a_new_scoring_session
+    click_link("Start a new score")
+    click_link("Start scoring")
+  end
+
+  def and_scores_the_ideation_section_perfectly
+    find("#judge-scores-app li.score-question:nth-child(1) li.score-value", text: "5").click()
+    find("#judge-scores-app li.score-question:nth-child(2) li.score-value", text: "5").click()
+    find("#judge-scores-app li.score-question:nth-child(3) li.score-value", text: "5").click()
+    find("#judge-scores-app li.score-question:nth-child(4) li.score-value", text: "5").click()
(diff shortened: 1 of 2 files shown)


stenington:
Cool! I can see why targeting the actual scoring buttons was a pain. 😟 

I'm comfortable with either style here. Style 2 is more descriptive for sure, I think it works nicely here. I wouldn't reject style 1 though, if that was all you had proposed. Given the options, I think I'd lean towards keeping 2.

shaunxp20:
> Cool! I can see why targeting the actual scoring buttons was a pain. 😟
> 
> I'm comfortable with either style here. Style 2 is more descriptive for sure, I think it works nicely here. I wouldn't reject style 1 though, if that was all you had proposed. Given the options, I think I'd lean towards keeping 2.

Cool, I'm onboard for style 2. I removed style 1, and made a few slight tweaks to style 2 - feel free to update anything in there though! Btw, I forgot to include the two links/articles I was looking at, if you're interested:
https://about.futurelearn.com/blog/how-we-write-readable-feature-tests-with-rspec
https://www.madetech.com/blog/feature-testing-with-rspec

Sorry about all of the build/test failures! I can't seem to figure it out, it passes for me locally, I hate the idea of adding another flaky test to the suite. 😕 😞 🤢

shaunxp20:
Gah! Sorry @stenington! About ready to pull my non-existent hair out over here! So many random failures! I tried updating Capybara to the latest version, that didn't seem to do anything though. I gotta call it quits for now, sorry about the poor hand off of this issue/PR. 😦 I can definitely pick it back up when I get back though.

spec/features/judge/score_a_submission_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+    and_scores_the_overall_section_perfectly
+    and_submits_the_score
+
+    then_the_review_page_displays_a_perfect_score_of 90
+  end
+
+  private
+
+  def given_there_is_a_submission_that_needs_to_be_scored
+    FactoryBot.create(:team_submission, :complete)
```
stenington: I believe this is creating a junior division submission

spec/features/judge/score_a_submission_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+    find("#judge-scores-app li.score-question:nth-child(1) li.score-value", text: "5").click()
+    find("#judge-scores-app li.score-question:nth-child(2) li.score-value", text: "5").click()
+    find("#judge-scores-app li.score-question:nth-child(3) li.score-value", text: "5").click()
+    find("#judge-scores-app li.score-question:nth-child(4) li.score-value", text: "5").click()
+    find("#judge-scores-app textarea").set("Lorem ipsum dolor sit amet consectetur adipiscing elit iaculis suspendisse natoque magna senectus, tempus nulla maecenas rutrum cursus euismod ante cras posuere proin himenaeos. Nisi primis ullamcorper penatibus vivamus dapibus, risus vel lobortis nam sed convallis, velit a cubilia hendrerit.")
+
+    find("#judge-scores-app a.button.btn-next").click()
+  end
+
+  def and_scores_the_entrepreneurship_section_perfectly
```
stenington: Junior division teams don't get judged on entrepreneurship, so this method is actually driving setting scores in the "Overall" section. This was tricky to see, because I believed this method was handling entrepreneurship when it wasn't. The problem is that each of these `and_scores_..._perfectly` methods isn't unequivocally navigating to the section mentioned, they're relying on the previous method to have done the navigation, and that previous method is actually just clicking the "Next" button.

stenington:
Ok, so my recommendations here are to test both junior and senior division submissions separately for completeness, and then have the methods scoring each section navigate unequivocally to the proper section, or fail loudly if it can't find that section. 

That aside, when I run this locally it's popping open a second Firefox window and running the tests there. I think all the other capybara tests run headless, so I'm not clear why it's doing that. Maybe the `js: true`? 

That would explain the `VCR::Errors::UnhandledHTTPRequestError:` seen in one of the earlier CI runs though. I believe it's trying to go grab Firefox but VCR isn't ignoring that url like it is `chromedriver.storage.googleapis.com` in `spec/vcr_helper.rb`.

I'd sort of prefer to run everything headless if possible, but I don't know if headless chrome handles js correctly, so maybe it's not an option. I think it's worth checking though.

stenington:
Oh, and I would roll back the capybara upgrade. Looks like that breaks tests that pass on the old version, so probably some method was deprecated or something? Best handled independently.

shaunxp20:
> Oh, and I would roll back the capybara upgrade. Looks like that breaks tests that pass on the old version, so probably some method was deprecated or something? Best handled independently.

👍 Agree! I removed it. 🙅‍♂ 😄

spec/features/judge/score_a_submission_spec.rb
shaunxp20: I added a feature spec for both junior and senior teams. 👧 👵 👍

spec/features/judge/score_a_submission_spec.rb
shaunxp20: Good catch! I made some changes to make it more explicit which division we're creating/scoring. 🥇 😃

shaunxp20:
@stenington Thoughts on renaming the branch/PR and/or creating another issue for the new feature specs? Since #2351 fixed the demo scoring bug. 🤔 😅

stenington:
Yeah, feel free to rename both to be more accurate. I think renaming the ticket and commenting that another ticket fixed the original bug but we're still writing the test is sufficient. A new ticket would also be fine with me, but I think introduces a little bit of weirdness in our iteration analytics to understand that ticket X replaced ticket Y and get the counts right.

shaunxp20:
I'm going to close this PR, I created a new PR #2371 to reflect the work done here (the feature specs). I also updated the original issue #2320 to reflect this as well.

Issue #2337, opened 2019-12-11
Security: Update rubyzip [8,8,1]


stenington:
As an ongoing part of #2166, we should update `rubyzip` to resolve [this security alert](https://github.com/Iridescent-CM/technovation-app/network/alert/Gemfile.lock/rubyzip/open)

stenington:
So this ticket was addressed with a set of dependency upgrades that should all be safe based on versioning, but I guess QA is just to let it sit in staging for a bit and see if people encounter strange problems as they QA other tickets.

PR #2357, opened 2020-01-22
#2337 updating locked dependencies to fix rubyzip alert

stenington:
This one confuses me a bit. It seems like by forcing bundler to re-evaluate the Gemfile, rubyzip ends up at the patched version we want. I think this involves more library upgrades than is strictly necessary to just fix rubyzip, but since the `Gemfile` itself wasn't change, this should be okay?


stenington:
`bundle update webdrivers` seems to be the trick for a smaller upgrade, but let's see how this bigger one does anyway.

Issue #2365, opened 2020-01-28
Replace outdated curriculum links on the submission screens [5,2,2.5]


rsgonzal:
We currently link to a number of curriculum units from the submission screens for girls. The links are not pointing to the correct units from this season.

This doc points to where we can find these links and what they should be replaced with: https://docs.google.com/document/d/1cC84zpkehmO_gX9Az_u2THajEJFEIWZmf8LFahs90jI/edit

rsgonzal:
@nickiredescent Could you help me test this to make sure the links are now going to the correct location, please?

nickiredescent:
All the links have been updated to the new, correct places. Looks good!

rsgonzal:
thanks, @nickiredescent ! looks good to go.

PR #2370, opened 2020-02-03
#2365 update curriculum links

stenington:
Making the link changes outlined here: https://docs.google.com/document/d/1cC84zpkehmO_gX9Az_u2THajEJFEIWZmf8LFahs90jI/edit

### DIFF
--- a/app/views/team_submissions/pieces/app_description.en.html.erb
+++ b/app/views/team_submissions/pieces/app_description.en.html.erb
@@ -55,7 +55,7 @@
     Get some help writing your app description in the
     <%= link_to(
       "ideation curriculum unit.",
-      "http://technovationchallenge.org/curriculum/ideation/",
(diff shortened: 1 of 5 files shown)


app/views/team_submissions/pieces/app_description.en.html.erb
```diff
@@ -55,7 +55,7 @@
     Get some help writing your app description in the
     <%= link_to(
       "ideation curriculum unit.",
-      "http://technovationchallenge.org/curriculum/ideation/",
+      "https://technovationchallenge.org/curriculum/ideation-6/",
```
shaunxp20: I'm wondering if we should/can make them all `https` links? 🤔 😈 🤓

stenington: Oh good call, the site will redirect to https but I do prefer to link directly, so I'll fix that.

PR #2371, opened 2020-02-05
Add feature specs to score submissions

### LINKED ISSUE #2320: Add feature specs for scoring submissions [8,8,1]

shaunxp20:
This will add feature specs to score both a junior and senior team's submissions.

This will also run these feature specs (and all other feature specs tagged w/ `js: true`) in a headless browser.

Fixes #2320

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -21,6 +21,7 @@ Dir[Rails.root.join('spec/support/**/*.rb')].each { |f| require f }
 ::Timezone::Lookup.lookup.default("America/Los_Angeles")
 
 Capybara.automatic_label_click = true
+Capybara.javascript_driver = :selenium_chrome_headless
 
 require 'rake'
 Rails.application.load_tasks
--- a/spec/features/judge/scoring_submissions_spec.rb
+++ b/spec/features/judge/scoring_submissions_spec.rb
@@ -1,0 +1,116 @@
+require "rails_helper"
+
+feature "scoring submissions", js: true do
+  before do
+    SeasonToggles.judging_round = :qf
+  end
+
+  after do
+    SeasonToggles.judging_round = :off
+  end
+
+  scenario "when a junior team receives a perfect score" do
+    given_there_is_a_submission_from_a_junior_team_that_needs_scoring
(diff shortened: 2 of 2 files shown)


spec/rails_helper.rb
```diff
@@ -21,6 +21,7 @@
 ::Timezone::Lookup.lookup.default("America/Los_Angeles")
 
 Capybara.automatic_label_click = true
+Capybara.javascript_driver = :selenium_chrome_headless
```
shaunxp20: Btw, here's the change I made to make these specs (and all other feature specs w/ the `js: true` attribute) in a headless browser. 👍 🎃🕴🐴

stenington: Nice!

shaunxp20:
First PR I'm going to merge in! 😅 😀  Just wanted to verify: looks like I just do a fast-forward (no merge commit) merge? And then delete the branch, close PR? Is that it? @stenington 🤔 🌞

stenington:
@shaunxp20 yeah, that's a conversation we can have if you want. I've done it before where you never merge your own PRs, the approver merges, so that's an option.

But for now, I usually select **Squash and merge** and click the button, which will close the PR for you, and then I delete the branch to keep things tidy. Then sometimes I move the ticket directly to test, other times if I'm being conscientious I wait for the CI build to deploy to staging.

We can do this on a screenshare today while we're looking at that support issue.

Issue #2366, opened 2020-01-29
Update calendar for RPE event creation [8,8,1]


rsgonzal:
The calendar when setting up a RPE highlights which dates are available. It is currently out of date and should be set with the range from April 22nd through May 25th 2020

![Screen Shot 2020-01-29 at 9 09 58 AM](https://user-images.githubusercontent.com/22801026/73379537-4a1a4b00-4277-11ea-950f-f1efc792363c.png)
![Screen Shot 2020-01-29 at 9 10 04 AM](https://user-images.githubusercontent.com/22801026/73379545-4dadd200-4277-11ea-8db7-23b9e9c36513.png)

hellafitz:
Decision: We will create new config variables to hold these dates, and this calendar widget should read that config. Right now the dates are hardcoded into the platform.

Need Program Team to confirm the relationship between these dates and Quarters (currently one day off). 
Currently we don't think RPEs work if they don't happen during the Quarterfinal window. 
Note that, if a dev or admin WAS to set RPE dates outside of the quarterfinals window there may be parts of the platform that won't work (so the Program Team will need to check it).

We'd like an instructive error message to display if dates are set outside of the quarters window.

rsgonzal:
@shaunxp20 checked in with the team and they're ok with the dates falling outside of the judging window.

shaunxp20:
> @shaunxp20 checked in with the team and they're ok with the dates falling outside of the judging window.

Thanks for the info! @rsgonzal! 

Hmm, does that mean we want to remove the date restrictions all together (that would be pretty easy to do)? So a RA can pick any date? 👉 📆 

Or do we want the new date restrictions of: April 22nd through May 25th 2020?

🤔 ⛰🌲🌞

rsgonzal:
The latter, we should change the new date restrictions to be April 22nd through May 25th 2020

shaunxp20:
> The latter, we should change the new date restrictions to be April 22nd through May 25th 2020

Sounds good! 👍 Roger Wilco! 🚀

shaunxp20:
I added some new configuration settings to make the RPE date range dynamic:
```
DATES_REGIONAL_PITCH_EVENTS_BEGINS_YEAR
DATES_REGIONAL_PITCH_EVENTS_BEGINS_MONTH
DATES_REGIONAL_PITCH_EVENTS_BEGINS_DAY

DATES_REGIONAL_PITCH_EVENTS_ENDS_YEAR
[...]

PR #2373, opened 2020-02-07
#2366 Make RPE date range configurable

shaunxp20:
This will make the RPE date range dynamic via new configuration settings:
```
DATES_REGIONAL_PITCH_EVENTS_BEGINS_YEAR
DATES_REGIONAL_PITCH_EVENTS_BEGINS_MONTH
DATES_REGIONAL_PITCH_EVENTS_BEGINS_DAY

DATES_REGIONAL_PITCH_EVENTS_ENDS_YEAR
DATES_REGIONAL_PITCH_EVENTS_ENDS_MONTH
DATES_REGIONAL_PITCH_EVENTS_ENDS_DAY
```

### DIFF
--- a/app/javascript/ra/events/EventForm.vue
+++ b/app/javascript/ra/events/EventForm.vue
@@ -174,8 +174,8 @@
           altInput: true,
           altFormat: "l, F J",
           dateFormat: "Y-m-d",
-          minDate: "2019-04-27",
-          maxDate: "2019-06-02",
+          minDate: this.minDate(),
+          maxDate: this.maxDate(),
         },
 
         timeOpts: {
@@ -461,6 +461,18 @@
           0
         );
         return newTime;
+      },
+
+      minDate() {
+        return process.env.DATES_REGIONAL_PITCH_EVENTS_BEGINS_YEAR + '-' +
+          process.env.DATES_REGIONAL_PITCH_EVENTS_BEGINS_MONTH + '-' +
+          process.env.DATES_REGIONAL_PITCH_EVENTS_BEGINS_DAY
+      },
+
+      maxDate() {
+        return process.env.DATES_REGIONAL_PITCH_EVENTS_ENDS_YEAR + '-' +
+          process.env.DATES_REGIONAL_PITCH_EVENTS_ENDS_MONTH + '-' +
+          process.env.DATES_REGIONAL_PITCH_EVENTS_ENDS_DAY
       },
     },
 


spec/technovation/configuration_settings_spec.rb
```diff
@@ -0,0 +1,51 @@
+  require "rails_helper"
```
shaunxp20: Let me know what you think about these specs, @stenington 🙏 

My thought was to ensure these configuration settings are setup correctly before deploying to staging/prod, and if they're not then the the tests will fail, which will break the build which will stop the deploy. 🤔 🤷‍♂ I also added a few simple specs to validate the date formats too.

If we like this, maybe we can add another issue for the rest of them?

A couple of random thoughts:

I struggled w/ referring to them as configuration settings vs environment variables, but I think it makes as configuration settings.

I like having specs explicit/wordy, and that's what I started doing, but your specs for the `curriculum-checks` specs inspired me to make them more dynamic. I think it's not so bad in this case. 🤔 🤷 😄

app/javascript/ra/events/EventForm.vue
```diff
@@ -462,6 +462,18 @@
         );
         return newTime;
       },
+
+      minDate() {
+        return process.env.DATES_REGIONAL_PITCH_EVENTS_BEGINS_YEAR + '-' +
+          process.env.DATES_REGIONAL_PITCH_EVENTS_BEGINS_MONTH + '-' +
+          process.env.DATES_REGIONAL_PITCH_EVENTS_BEGINS_DAY
```
shaunxp20: I tried string interpolation here, but (imho) it didn't make it any more readable. 😕 Open to suggestions!

shaunxp20: Having one environment variable contain the entire date would probably help, but I was just following the convention of our other environment variables where the date is split up. 💇‍♂ 💅 💇 🤷‍♂ 😺

spec/technovation/configuration_settings_spec.rb
shaunxp20: Hmm, it looks like the build is failing b/c the environment variables need to be setup in CircleCI as well. 🤔 😒 I'll wait to see if we want to go down this path before adding them/moving forward. 👍 😃

stenington: Well, the problem here is that we don't run tests in production, so these tests won't actually ensure that our production environment is correct, just that our test environment is correct (whether that's a local dev environment or CI). While that may be a little bit useful, I worry it gives a false sense of confidence that things will work in production without thinking about it carefully. What do you think?

app/javascript/ra/events/EventForm.vue
stenington: I had absolutely no idea you could use `process.env` from front end code, so this is interesting. 

I was assuming we'd read the config [somewhere in here](https://github.com/Iridescent-CM/technovation-app/blob/qa/app/constants/important_dates.rb) and render it as a data attribute onto a page or whatever it is this Vue app otherwise does to get initialized. (Note, I'd prefer that Vue apps initialize with an initial ajax call but many of them don't currently.)

This is in a way much cleaner, but at the cost of decentralizing some of our reading-dates-from-the-env setup. I guess my question is whether or not this will "fail fast" if that config is missing? Or will the front end just display a broken date picker? What if instead of string concatenation you pass the variables to `new Date(...)`, would that fail fast if the config wasn't there, or will the app still run with a broken date picker?

spec/technovation/configuration_settings_spec.rb
shaunxp20: > Well, the problem here is that we don't run tests in production, so these tests won't actually ensure that our production environment is correct

Ah! Good to know! I guess I went a little off reservation!! 😂 😊 😁 FWIW, I would be interested in exploring the pros/cons of running some of our smoke tests in production. I think running these (and some others) would provide value. 😄 

And as you pointed out, these new specs would ensure our local environments are setup correctly (which I think would add some value), but falls short of my main goal of ensuring staging/prod is setup correctly (where it stand now anyway). I think I'll remove/stash these specs for now, to not give the false sense of confidence (I agree), and it would be consistent w/ our other ENVs in that none of the others are currently being tested. 😀 😁 

Anywho, I found something else that we might be able to use w/ our `dotenv-rails` gem. There's a section on [Required Keys](https://github.com/bkeepers/dotenv#required-keys):

> If a particular configuration value is required but not set, it's appropriate to raise an error.
```
Dotenv.require_keys('BLAH')
```
> If any of the configuration keys above are not set, your application will raise an error during initialization. This method is preferred because it prevents runtime errors in a production application due to improper configuration.

Maybe this could meet the "fail fast" requirement too. 🤔 Thoughts?

stenington: Yeah, that's a possibility. Right now I believe we don't even install `dotenv-rails` in staging or production, but I'm very open to something like an env initializer that blows up on missing required settings regardless of environment, that would be a great thing to have.

stenington: We discussed this in stand briefly, but writing it here as well: It occurred to me that if we were able to run ENV checks as part of a Heroku build, we might be able to prevent a deployment from evening happening if certain criteria weren't satisfied, which is sort of nicer than deploying an app that then immediately refuses to run. Something to look into maybe... I'm not entirely sure how it might be done or if there's a downside I'm not thinking of.

shaunxp20: I'm not sure if this is what you were thinking, but I'll throw this idea out there... Apparently the ENV vars on Heroku are available via an API: [Config Vars Info for App](https://devcenter.heroku.com/articles/platform-api-reference#config-vars)

We could pull down the ENV vars via the API and do some validations against them, then make this a part of the build process (so the build will fail if the validations don't pass). I have no idea if this is what you had in mind! 😊 😄 🤔 🏖 🍹 🌞

shaunxp20: @stenington and I discussed this. In summary, I'm going to open a new issue to test our ENV vars; it's a little out of scope for this PR, and that way it doesn't hold up this PR either. 👍 ▶️

shaunxp20:
@stenington Unfortunately, no tests for this PR. 😭 😞 But I did verify the functionality again locally. I also added the appropriate config/ENV key/values to both staging and production, and added QA steps in [#2366](https://github.com/Iridescent-CM/technovation-app/issues/2366#issuecomment-587923099). Let me know what you think! 😄

stenington:
👍 Works for me!

PR #2387, opened 2020-02-14
Rake task to fix onboarding students

stenington:
This is a minimal rake task to do the fix I would have otherwise done by hand for the issue of students being marked onboarding when really they're fully onboarded.

Doing this uncovered a can of worms in the form of that other rake task, `check_onboarding`. Firstly, it's broken. It throws an exception when I try to run it, **not** related to the way it reports problems with `raise`. But secondly, it shows a couple more ways that onboarded/onboarding issues can present on students or teams that we may end up needing to consider.

I also took a stab at a basic test for this task. There are probably improvements to make, but I'm trying to find the minimal acceptable solution for now. So in that vein, please let me know how close or far away I am from something you think you could live with @shaunxp20 😄


lib/tasks/check_onboarding.rake
```diff
@@ -1,3 +1,12 @@
+task onboard_missed_students!: :environment do
+  StudentProfile.current.onboarding.find_each do |student|
```
shaunxp20: Nice @ `find_each`! 👍

stenington: Yeah, that was a lesson learned the hard way in past seasons. 😀

stenington:
This PR is for #2388

Issue #2378, opened 2020-02-13
Show official/unofficial event status to RAs [8,8,1]


rsgonzal:
Currently, all events are official when created. Admin has the ability to toggle them unofficial if they do not meet the event requirements

RAs are not given this information on their dashboards.

Proposal: When an event is toggled un-official by admin, show messaging on the RA event page telling them their event is un -official. 

![Screen Shot 2020-02-13 at 3 52 58 AM](https://user-images.githubusercontent.com/22801026/74432962-5aecc400-4e14-11ea-868c-89a46ad02f41.png)

Unknown: How to provide messaging for official events 
- After a certain date, all events not toggled un-official will replace above text with text that says "this event is an official event and will be scored)
- Program team just provides messaging that if they don't see if their event is un-official, then it is

stenington:
Similar to #2377, as a counter-proposal we could show **Status: official/unofficial/pending, check back on...** to RAs on the event in their dashboard. They probably don't need extra clarification on what those things mean, that's communicated to them through training or somewhere like that.

stenington:
We would need the actual date, which we agree could be hardcoded at this point.

rsgonzal:
@stenington April 22nd is the date.

stenington:
@rsgonzal What do you think about timezone issues here? April 22nd isn't the same set of 24 hours around the world, and if we want the platform to automatically do something different after a certain time, I think we need a specific date and time, including timezone.

stenington:
I think we have to make the date configurable to test this. Otherwise there's no good way I know of to make staging think it's after the finalization date to see that display.

stenington:
### QA steps

The relevant config settings here are
* `DATES_RPE_OFFICIALITY_FINALIZED_YEAR`
* `DATES_RPE_OFFICIALITY_FINALIZED_MONTH`
* `DATES_RPE_OFFICIALITY_FINALIZED_DAY`

**With config set to a date in the future, relevant to when you're testing...**

1. log in as an RA
2. create an event
3. you should see **Status: pending** on your event

**With config set to a date in the past, relevant to when you're testing...**

1. log in as an RA
2. create an event
3. you should see **Status: official** on your event
4. as an admin, mark the event as unofficial
5. as the RA again, you should see **Status: unofficial** on your event

**With config set to the date on which you're doing the testing...**

1. log in as an RA
2. create an event
3. you should see **Status: pending** on your event

[...]

PR #2397, opened 2020-02-20
#2378 show RA their event's status or pending depending on date

stenington:
This should handle #2378, showing RAs their event status as "pending" prior to a finalization date, and showing "official/unofficial" afterwards. The finalization date is set by `DATES_RPE_OFFICIALITY_FINALIZED_*` variables, which will be set to April 22, 2020 in production.

The way I have it written, I believe RPE statuses will be considered finalized on April 23, starting at midnight, Pacific time.


app/javascript/ra/events/Event.js
```diff
@@ -23,6 +23,12 @@ export default function (event) {
 
   this.capacity = event.capacity || null;
 
+  if (event.officiality_finalized) {
+    this.officiality = event.official ? "official" : "unofficial";
+  } else {
+    this.officiality = "pending";
+  }
+
```
shaunxp20: Not a blocker, just wondering if it makes sense to move this logic to `regional_pitch_event.rb`, and just expose one method/JSON item for `event.status`? 🤔 😅 😬

stenington: Yeah, I could be talked into that.

stenington:
@shaunxp20 maybe this looks a little better?

shaunxp20:
> @shaunxp20 maybe this looks a little better?

Oh nice! I like how it makes it easier to reuse this functionality now too, it's not tied to the Vue.js layer. Looks good to me! 👍 🙌

shaunxp20:
> This one's looking good to me! 🌞 Thanks for digging into our time zones! 🥇 Btw, I only ran the RPE specs (all passing for me), I didn't manually test it, but the changes look good. 👍

Gah, comment/approval for the wrong PR! 🤦‍♂ 😄 😂

Issue #2347, opened 2020-01-10
Platform allows same email for user and parent but then breaks [8,5,1.6]


rsgonzal:
**Describe the bug**
Student could not create a team or edit her profile. It was identified that she had set the same email for her account and for her parent consent. Apparently the platform does not like this and stops working when this happens. User isn't notified

**To Reproduce**
1. Go to http://my.technovationchallenge.org/
2. Sign up with an email
3. Set up a different email for your parent
4. Change your email to be the same as parent email
5. verify your email account
6. Sign your consent form
7. try to create a team

**Expected behavior**
You should not be allowed to use the same email for parent and student, the platform shouldn't break, and a message should be displayed (e.g. "Please enter a different email address")

**Additional context**
Nick was able to identify the issue and we told the student to use a different email for her parent.

#### QA steps to ensure students can't have the same email address as their parent's/guardian's
 * Sign up as a new student
 * Use a different email address for the parental consent form
 * Try to update the student's email address to be the same email address as their parent's/guardian's, this was possible before, verify it's not possible and you see an error

#### QA steps for existing students who have the same email address as their parents
 * Run the dataclip on staging to find students who have the same email address as their parents
* Sign in as one of these students
* Try to update the student's profile
* Verify you get redirected to the parental consent form, and see an error message about shared email addresses

#### Other QA steps
Changes were made to `profile_controller.rb` which all account types use. It would be a good sanity check to make sure the other account types are working as expected.

hellafitz:
Not able to reproduce as described. Need new steps to reproduce. @rsgonzal to investigate.

rsgonzal:
Updated steps to reproduce above.

Current broken account: roxana+ptest123@technovation.org

hellafitz:
When a student edits their email address the platform should check that they're not changing it to what they have as their parent address. Prevent the change and show an error message.

shaunxp20:
I have been testing this locally, and I found an issue.

It's possible right now for students and parents to share the same email address (after this issue #2347 gets resolved it shouldn't be possible for new accounts), if this is the case for a current student, I'm finding that no profile fields can be updated. The user will be redirected to the main "Edit profile" page with the error: **Sorry, there was a problem. Check the fields below.** and there won't be any fields w/ errors. I tested/verified this locally on the QA branch, which means it is existing behavior.

I made a dataclip of all the students who share the same email address as their parents and there seems to be quite a few of them (hundreds).

[...]

PR #2398, opened 2020-02-21
Validate student email address is different than parent/guardian's email address

shaunxp20:
I added a custom validator to ensure that the student's email address is different than the parent/guardian's email address.

Repro steps are in the issue #2347

### DIFF
--- a/app/controllers/concerns/profile_controller.rb
+++ b/app/controllers/concerns/profile_controller.rb
@@ -41,6 +41,8 @@ module ProfileController
       else
         render "passwords/edit"
       end
+    elsif profile.errors["account.email"].any?
+      render 'email_addresses/edit'
     elsif profile.errors["account.city"].any? or
             profile.errors["account.state_province"].any? or
               profile.errors["account.country"].any?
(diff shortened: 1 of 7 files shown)


shaunxp20:
@stenington I am still going to do more QA on this locally (later today), I just wanted get a PR up for you to look at sooner rather than later. 😅 It makes me nervous touching `profile_controller.rb`. 😬 😬 😅

app/controllers/concerns/profile_controller.rb
```diff
@@ -41,6 +41,8 @@ def update
       else
         render "passwords/edit"
       end
+    elsif profile.errors["account.email"].any?
+      render 'email_addresses/edit'
     elsif profile.errors["account.city"].any? or
             profile.errors["account.state_province"].any? or
               profile.errors["account.country"].any?
```
shaunxp20: Btw, my goal was to test each branch of the conditionals here for students, but I couldn't find a way to get any (validation) errors for `city`, `state_province`, or `country`, I couldn't even find any validators for these fields, anywho tldr; I'm not sure how to reach this branch! 🤷‍♂ 🌳 🚶 🎶

stenington: Yeah, they may be unreachable. I've encountered a fair amount of leftover stuff around locations.

shaunxp20:
> I'm good with the PR, if work is done. 👍

Cool, at this point, I think a good next step would be to put it on staging for more QAing. 😬 😅

PR #2402, opened 2020-02-25
#2378 make rpe_officiality_finalized config independent of user timezone

stenington:
Despite only specifying a date in configuration (year/month/day) what we're really talking about is a specific time in a specific timezone when we want RPE statuses to go from being pending to finalized. This change reflects that a little bit better: the config is now assumed to be expressing midnight of the date configured, in the application's default timezone which is Pacific time.

Now the comparison against an RA's local time can happen much more transparently, and all RA's will be able to see their statuses simultaneously rather than on a rolling basis throughout the configured day, which I think was the intention.

### DIFF
--- a/app/constants/important_dates.rb
+++ b/app/constants/important_dates.rb
@@ -24,7 +24,7 @@ module ImportantDates
     year = Integer(ENV.fetch("DATES_RPE_OFFICIALITY_FINALIZED_YEAR"))
     month = Integer(ENV.fetch("DATES_RPE_OFFICIALITY_FINALIZED_MONTH"))
     day = Integer(ENV.fetch("DATES_RPE_OFFICIALITY_FINALIZED_DAY"))
-    Time.find_zone(Rails.configuration.time_zone).local(year, month, day, 0, 0, 0)
+    Time.find_zone(Rails.configuration.time_zone).local(year, month, day, 0, 1)
   end
 
   def self.submission_deadline
--- a/app/constants/important_dates.rb
+++ b/app/constants/important_dates.rb
@@ -24,7 +24,7 @@ module ImportantDates
     year = Integer(ENV.fetch("DATES_RPE_OFFICIALITY_FINALIZED_YEAR"))
     month = Integer(ENV.fetch("DATES_RPE_OFFICIALITY_FINALIZED_MONTH"))
     day = Integer(ENV.fetch("DATES_RPE_OFFICIALITY_FINALIZED_DAY"))
-    Time.zone.local(year, month, day)
+    Time.find_zone(Rails.configuration.time_zone).local(year, month, day, 0, 0, 0)
   end
 
   def self.submission_deadline
--- a/app/constants/important_dates.rb
+++ b/app/constants/important_dates.rb
@@ -24,7 +24,7 @@ module ImportantDates
     year = Integer(ENV.fetch("DATES_RPE_OFFICIALITY_FINALIZED_YEAR"))
     month = Integer(ENV.fetch("DATES_RPE_OFFICIALITY_FINALIZED_MONTH"))
     day = Integer(ENV.fetch("DATES_RPE_OFFICIALITY_FINALIZED_DAY"))
-    Time.find_zone(Rails.configuration.time_zone).local(year, month, day, 0, 1)
+    Time.find_zone(Rails.configuration.time_zone).local(year, month, day, 23, 59)
   end
 
   def self.submission_deadline
--- a/app/models/regional_pitch_event.rb
+++ b/app/models/regional_pitch_event.rb
@@ -105,7 +105,7 @@ class RegionalPitchEvent < ActiveRecord::Base
   end
 
   def officiality
-    if Time.zone.today > ImportantDates.rpe_officiality_finalized
+    if Time.zone.now >= ImportantDates.rpe_officiality_finalized
       official? ? :official : :unofficial
     else
       :pending
--- a/spec/helpers/important_dates_spec.rb
+++ b/spec/helpers/important_dates_spec.rb
@@ -2,9 +2,9 @@ require "rails_helper"
 
 RSpec.describe ImportantDates do
   describe ".rpe_officiality_finalized" do
-    it "is 12:01 am of the configured date" do
-      expect(ImportantDates.rpe_officiality_finalized.hour).to eq(0)
-      expect(ImportantDates.rpe_officiality_finalized.min).to eq(1)
+    it "is 11:59 pm of the configured date" do
+      expect(ImportantDates.rpe_officiality_finalized.hour).to eq(23)
+      expect(ImportantDates.rpe_officiality_finalized.min).to eq(59)
     end
 
     it "is the application configured time zone regardless of user time zone" do
--- a/spec/helpers/important_dates_spec.rb
+++ b/spec/helpers/important_dates_spec.rb
@@ -1,0 +1,22 @@
+require "rails_helper"
+
+RSpec.describe ImportantDates do
+  describe ".rpe_officiality_finalized" do
+    it "is 12:01 am of the configured date" do
+      expect(ImportantDates.rpe_officiality_finalized.hour).to eq(0)
+      expect(ImportantDates.rpe_officiality_finalized.min).to eq(1)
+    end
+
+    it "is the application configured time zone regardless of user time zone" do
+      [
+        'Alaska',
+        'London',
+        'Tokyo',
+      ].each do |zone|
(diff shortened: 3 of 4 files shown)


app/constants/important_dates.rb
```diff
@@ -24,7 +24,7 @@ def self.rpe_officiality_finalized
     year = Integer(ENV.fetch("DATES_RPE_OFFICIALITY_FINALIZED_YEAR"))
     month = Integer(ENV.fetch("DATES_RPE_OFFICIALITY_FINALIZED_MONTH"))
     day = Integer(ENV.fetch("DATES_RPE_OFFICIALITY_FINALIZED_DAY"))
-    Time.zone.local(year, month, day)
+    Time.find_zone(Rails.configuration.time_zone).local(year, month, day, 0, 0, 0)
```
shaunxp20: 😎 👍 If I'm understanding correctly, this is where we're converting the date to midnight PST? Just wanted to double check that it's midnight, and not like 23:59.. the day of? Times/time zones are so complicated/hard for me! 😊 🤒 😵 💫 😅

stenington: I didn't realize this beforehand, but RA requests are wrapped in a `set_time_zone` action that sets the application time zone to whatever local time zone we've stored for the RA account. I think the semantics of this config (and probably a bunch of other config, but that felt out of scope) is that we're trying to specify things in a single time zone regardless of where the RAs might be, so that's why this is now referencing `Rails.configuration.time_zone`.

Additionally, making it a time and date (not just a date) preserves that timezone information for later when we want to compare against this.

spec/models/regional_pitch_event_spec.rb
```diff
@@ -6,11 +6,10 @@
     let(:official_rpe) { FactoryBot.create(:rpe, unofficial: false) }
     let(:unofficial_rpe) { FactoryBot.create(:rpe, unofficial: true) }
 
-    it "is pending before or on finalization date" do
+    it "is pending before finalization date" do
```
shaunxp20: I think it would be helpful here to add that it's the finalization date/time according to midnight PST. 🤔 ❔

app/models/regional_pitch_event.rb
```diff
@@ -105,7 +105,7 @@ def self.find(id)
   end
 
   def officiality
-    if Time.zone.today > ImportantDates.rpe_officiality_finalized
+    if Time.zone.now >= ImportantDates.rpe_officiality_finalized
```
stenington: This comparison is now getting the RA's current time, including time zone, and comparing that against the config which is expressed in the application timezone. That ends up being a safe, time-zone-aware comparison that causes the switch for all RAs simultaneously rather than rolling out over 24 hours depending on their local date.

spec/models/regional_pitch_event_spec.rb
```diff
@@ -19,10 +18,24 @@
...
     end
 
-    it "is appropriate status after finalization date" do
-      Timecop.freeze(finalization_date + 1.day) do
-        expect(official_rpe.officiality).to be :official
-        expect(unofficial_rpe.officiality).to be :unofficial
+    it "is appropriate status on or after finalization date" do
+      [
+        finalization_date,
+        finalization_date + 1.day,
```
shaunxp20: I kinda liked the `1.minute` difference test case above. 🤷‍♂ 😃

app/constants/important_dates.rb
stenington: Yeah, maybe it should be `local(year, month, day, 0, 0, 1)` so it's more unambiguously the start of the specified date, not the very end. Doing things right on the midnight boundary is confusing that way.

spec/models/regional_pitch_event_spec.rb
stenington: I can add that.

spec/models/regional_pitch_event_spec.rb
stenington: Ok, I can add that here too.

spec/models/regional_pitch_event_spec.rb
stenington: Well, I'll add in the midnight part, but the timezone is whatever is configured in `config/application.rb` which could change.

app/models/regional_pitch_event.rb
shaunxp20: Ah, ok, this is making sense now, thanks for explaining! I see the `around_action :set_time_zone` for RAs (that the RA's RPE controller uses). 👍 

Hmm, total side note,  I'm wondering how this will work for when we display the event status for students/mentors since they don't currently have that `around_action`. 🤔

stenington: Yep, I would expect we'll have time zone bugs to iron out, but at least now I feel like I have some understanding of how they might be happening!

spec/helpers/important_dates_spec.rb
```diff
@@ -0,0 +1,22 @@
+require "rails_helper"
+
+RSpec.describe ImportantDates do
+  describe ".rpe_officiality_finalized" do
+    it "is 12:01 am of the configured date" do
```
stenington: So I added a test here to clarify that the finalization happens at 12:01 am. That way the RPE model specs can test the behavior before, on, and after the finalization point without actually caring what the point is in the tests. Does that work for you?

app/models/regional_pitch_event.rb
shaunxp20: Yes! Totally! 👍

spec/helpers/important_dates_spec.rb
shaunxp20: Nice! I like this! 🚩I still think it's weird though that it's the start of the day and not the end of the day. 🚩

spec/models/regional_pitch_event_spec.rb
```diff
@@ -2,27 +2,48 @@
 
 RSpec.describe RegionalPitchEvent do
   describe ".officiality" do
-    let(:finalization_date) { ImportantDates.rpe_officiality_finalized }
+    let(:finalization_point) { ImportantDates.rpe_officiality_finalized }
     let(:official_rpe) { FactoryBot.create(:rpe, unofficial: false) }
     let(:unofficial_rpe) { FactoryBot.create(:rpe, unofficial: true) }
 
-    it "is pending before or on finalization date" do
+    it "is pending before finalization point" do
```
shaunxp20: 👍

spec/models/regional_pitch_event_spec.rb
```diff
@@ -2,27 +2,48 @@
...
           expect(unofficial_rpe.officiality).to be :pending
         end
       end
     end
 
-    it "is appropriate status after finalization date" do
-      Timecop.freeze(finalization_date + 1.day) do
-        expect(official_rpe.officiality).to be :official
-        expect(unofficial_rpe.officiality).to be :unofficial
+    it "is appropriate status on or after finalization point" do
```
shaunxp20: 👍

app/constants/important_dates.rb
shaunxp20: 👍 Thanks for explaining this! That makes sense - making all of our start/end dates/times based on a single time zone: `Rails.configuration.time_zone` (PST). 🌎 🕥

spec/helpers/important_dates_spec.rb
stenington: That's fair, there's no reason it can't be 11:59 pm. I think midnight technically belongs to the following day, so when I initially used midnight that's why it became a start-of-day thing, but maybe I should make it 11:59pm, and then it behaves more like what @rsgonzal and others already tested.

spec/models/regional_pitch_event_spec.rb
shaunxp20: > Well, I'll add in the midnight part, but the timezone is whatever is configured in `config/application.rb` which could change.

Ah yeah, that makes sense. 👍 😃

spec/helpers/important_dates_spec.rb
shaunxp20: Cool, I'm onboard w/ that, just wanted to be sure it was working as expected (however that should be). 👍 

Btw, TIL (from wikipedia):

> The "midnight" term can be avoided altogether if the end of day is noted as 24:00 and the beginning of day as 00:00. While both notations refer to the same moment in time, the choice of notation allows its association with the previous night or with the following morning. This approach follows the 24-hour time specification of ISO 8601.

stenington: Oh wow, that's pretty clever.

Issue #2377, opened 2020-02-13
Show official/unofficial event status to students and mentors [13,8,1.6]


rsgonzal:
Currently, all events are official when created. Admin has the ability to toggle them UN-official when they reach the correct amount of teams

![Screen Shot 2020-02-13 at 3 22 15 AM](https://user-images.githubusercontent.com/22801026/74429886-9d140680-4e10-11ea-9179-af3039caa894.png)

This information is not passed on to students and creates a lot of confusion for teams. They assume they are attending an official event but then only receive online scores.

Proposal: add messaging to the student dashboard and event page showing the UN-official status once an admin has toggled an event UN-official

![Screen Shot 2020-02-13 at 3 55 07 AM](https://user-images.githubusercontent.com/22801026/74433153-b7e87a00-4e14-11ea-8bbf-8551d96dd5f8.png)

![Screen Shot 2020-02-13 at 4 34 07 AM](https://user-images.githubusercontent.com/22801026/74435788-1bc17180-4e1a-11ea-8ea6-81bf9262eca6.png)

![Screen Shot 2020-02-13 at 4 48 17 AM](https://user-images.githubusercontent.com/22801026/74436916-1bc27100-4e1c-11ea-8a5c-5fd44cab2612.png)

## QA
Based on the screenshots above, log in as both a **student** and as a **mentor** and make sure you see the expected messages where you expect to see them. 

**One note on the dates**: it is set in config, which I'm currently setting to 4/23/2020 to mimic production. However, if you see a different date in the screens as you QA, as @rsgonzal or @stenington to check what the QA site has been configured to. It may have been changed.

stenington:
We discussed a modification here where we might show **Status: official/unofficial/pending, check back on...** alongside the Time and Address. We might also want to add messaging about what those words mean, or link to an FAQ, or something like that. The next step for this ticket is to mock up what that alternative could look like, and get more concrete about the copy.

rsgonzal:
Question about how emails get triggered. Might need to edit auto email that gets sent to come back and check for status.

rsgonzal:
@stenington @shaunxp20 

I did some quick mock ups here.

**Before an event has been selected**
First event screen, can we add some sort of text here? (same image as before)
![Screen Shot 2020-02-19 at 4 42 32 PM](https://user-images.githubusercontent.com/22801026/74889759-db408700-5336-11ea-9510-ee61a1729c67.png)

[...]

PR #2405, opened 2020-02-27
#2377 show students and mentors event status and status disclaimers

stenington:
This PR for #2377 shows mentors and students the event officiality status when they're looking at an event, and it adds some disclaimers to the site as well about when that status gets finalized. 

I'm opening this PR without tests with the understanding that I might need to write some, but I wanted to get some guidance on that. @shaunxp20 do you think it would be appropriate to write some feature specs here that navigate some of the event selection process for mentors and students? I don't see any existing ones to build on, so maybe it would be good to get some started.


app/views/mentor/dashboards/onboarded/_regional_pitch_events.html.erb
```diff
@@ -5,7 +5,9 @@
   pitch to a live panel of judges!
 </p>
 
-<div class="step-actions">
+<%= render 'regional_pitch_events/dashboard_disclaimer' %>
```
shaunxp20: I was always wondering if there is a convention around when we use single vs double quotes. #nbd-jc (custom acronym i just made up for: no big deal, just curious) 😄 😁 😄

stenington: I don't have one, do you have a proposal? A quick google search does come back with lots of "always use double quotes" type discussions so maybe that's a habit I need to pick up?

stenington:
Yeah I don't disagree.

app/views/mentor/dashboards/onboarded/_regional_pitch_events.html.erb
shaunxp20: Well, my personal preference (which is in the minority from my experience) is to always use single quotes, and then use double quotes if you need to do string interpolation (or need to use single quotes inside the string, or some other reason). This provides helpful information, imho, double quotes - there's something different about this string, whereas double quotes everywhere, no extra information. Having said all of that, I'm (like w/ merging branches in) more a fan of consistency. And fwict, it looks like we use double quotes everywhere. But honestly, either way fine w/ me. #nbd! 👍 😄 😁 😃

stenington: Ha ha, yeah looks like [Thoughtbot](https://github.com/thoughtbot/guides/tree/master/style/ruby) and [Github](https://github.com/github/rubocop-github/blob/master/STYLEGUIDE.md#strings) ruby style guides both prefer double quotes. 

I guess I don't actually have a preference on which way to do it, my only preference is that something like rubocop enforces it so I don't have to police myself, which I will constantly forget to do for this kind of detail.

shaunxp20: +1 for rubocop enforcement. 💂 🙅 🛑

Issue #2401, opened 2020-02-25
Update TG Emails that get sent out for RPEs (5,5,1)


rsgonzal:
When a user joins an event or when they are added to an event by an RA, they get sent an email from TG.

The team wants to add messaging about the event status in those emails, letting them know that the status of their email will be finalized after April 21st.

Updated copy is here: https://docs.google.com/document/d/1iU-8y7H1Xl7Q5xf7BsZKpkzAa_EoLAt4Ew5Diu-6G-A/edit#

I pulled out the coded emails, but to edit I used what the user sees.

I have a question about one from Team_Mailer --> Notify added to an event
- not sure if this one is sent out, I think the one being sent is the one from Event_Mailer above

## QA
As an **RA**...
1. create an event if necessary
2. add a judge to your event
3. add a team to your event
4. confirm the judge receives an email titled "You are invited to judge a Technovation Girls event" with the new disclaimer
5. confirm the student(s)/mentor(s) receive an email titled "Your team is invited to attend a Technovation Girls event" with the new disclaimer

As a **team member** (student or mentor)...
1. log in
2. select a regional pitch event to attend
3. confirm the team member(s) receive an email titled "<Team name> has joined a Technovation Girls event: <event name>" with the new disclaimer

nickiredescent:
Tested.
Can confirm that participants joining the event receive the updated emails with the new headers and disclaimers. This is true for team members that were either directly invited to attend or chose an event themselves.

rsgonzal:
This one looks good to me too! Think Ophelie is taking a look tomorrow.

ophelieh:
**Student:** works as expected. 

Side Note: For the student & the RA, the status of the event showed as pending but on the admin event tab the status of the events shows as official or unofficial. 
![image](https://user-images.githubusercontent.com/32206421/76098137-89326f00-5f8e-11ea-8641-6f03a8635002.png)
![image](https://user-images.githubusercontent.com/32206421/76098165-96e7f480-5f8e-11ea-91d1-b71c5d1e8782.png)


**RA:** Everything worked as expected. 

Important Notes:
* For judges though, I would remove the part about RPE vs celebration event as it doesn't really impact them. 
![image](https://user-images.githubusercontent.com/32206421/76098500-30afa180-5f8f-11ea-89a8-26a010fd76cf.png)
[...]

PR #2407, opened 2020-03-03
#2401 update some rpe emails

stenington:
Update some RPE email titles and add the official/unofficial disclaimer to RPE invitation emails.


app/views/event_mailer/invite.en.html.erb
```diff
@@ -25,6 +25,17 @@
...
         </td>
       </tr>
+      <tr>
+        <td class="content-block">
+          Your event will be confirmed as either a Regional Pitch Event or Celebration Event
+          after <%= ImportantDates.rpe_officiality_finalized.strftime("%B %-d") %>. Please sign
+          into your account to verify the status of your event after that date. For more
+          information about pitch events, please check out the
+          <%= link_to "Technovation FAQ",
+            "https://iridescentsupport.zendesk.com/hc/en-us/articles/115002731414-All-about-Regional-Pitch-Events"
```
shaunxp20: Hmm, I think this link is broken? 🤔 🔗 🙅‍♂ 😄

stenington: It sure is! It's the link from the doc, but not working. @rsgonzal is this a page that is going to eventually exist, or do we need a new url?

rsgonzal: @stenington team accidentally switched the link. Here is the new one: https://iridescentsupport.zendesk.com/hc/en-us/sections/360007469154-Regional-Pitch-Events

stenington:
@shaunxp20 How do you feel about this one with a corrected link? Thanks for catching that!

Issue #2391, opened 2020-02-19
Allow Admin to know when an event has been created by an RA (by 3/21) [5,5,1]


rsgonzal:
Program team wants to know when an RA created an event.

They are giving the RAs a specified time frame for when their events can be created, but nothing prohibits them from creating an event after March 22nd.

The current plan is to download all of the events on the 22nd, and mark and events that show up after that as unofficial. 

It would be helpful to either have a time stamp of when the event was created on the event page or to prohibit the RAs from creating events after a specific date (3/22).

## QA steps
1. log in as an RA
2. create an event
3. log in as an admin
4. go to **Events**
5. add **Created at** and **Last updated at** as "More columns", and search
6. confirm the new event has the expected timestamps, and is unofficial by default
7. view that event
8. confirm that you see the timestamps here too

hellafitz:
Other options: 
- hide the 'create new event' button. That'd mean even admins couldn't create one. 
- also validate the date on the backend

Showing timestamps option is easy but requires the program team to monitor it (which they do say they're prepared to do). Hide the button option(s) are more robust but would need additional definition. 

Q: What happens if we do want to create an event after the deadline, as a special case (which no one says will happen until it does)?

Decision: Given the timeline, we should show the timestamps on the event page and as an optional column on the list of events page, and have program team monitor the numbers of late and 'exception' events

stenington:
I worked ahead on this one a bit, and have a potential PR. I'm not pulling it into the iteration yet though.

stenington:
So the base change here is just showing admins the creation date (and latest update date, because that's already there as well) of events. I think another easy thing would be to default all new events to being unofficial. That way our team doesn't have to patrol for new events and mark them as unofficial after the date, they'll automatically be unofficial and only the exceptions need to get marked as official.

nickiredescent:
Tested.
Can confirm that the time stamps are available at all the stated places and that the time columns are appearing as expected.

rsgonzal:
Wooo. Looks good. Ophelie is still taking a look.

Question @stenington, so all events are no unofficial by default? No matter what the RPE officiality date is set to?

stenington:
@rsgonzal I'm not 100% sure what you're asking, but here's how it works. It used to be that all events defaulted to being official, and an admin would have to change them to be unofficial. Now they're going to default to being unofficial, and an admin will have to change them to be official.

[...]

PR #2410, opened 2020-03-04
#2391 default events to unofficial and show event timestamps

stenington:
This should help out the program team, because now events will be defaulted to "unofficial" until they go through and promote certain events to be official. After that point, RAs aren't supposed to create more events, but if they do, they will be unofficial by default with no action from the program team, which is appropriate for the program policy.

@shaunxp20 one thing that bugged me while doing this one... can you think of a good place to pull out the timestamp formatting? Should that live in a helper somewhere?


stenington:
Okay, then I'll leave it. Thanks!

Issue #2411, opened 2020-03-04
Admin judge suspension controls [20,13,1.54]


stenington:
Implementation ticket for #2272 

The following are the specs for this implementation:
* add a **Judge suspension** section to the admin participants page for judges (`app/views/admin/participants/_judge_debugging.html.erb`)
* Help text to go in the section:
> Suspending a judge does not mean that their scores will be deleted, but they will not be able to create new ones. Please note any changes you make here in the appropriate tracking sheet.
* The section contains one of two buttons, with confirmation text:
    * **Suspend this judge**: You are about to SUSPEND this judge
    * **Enable this judge**: You are about to ENABLE this suspended judge
* Both are the "danger" color
* create the appropriate admin controller(s), and maybe method(s) on `JudgeProfile`, to back the two actions
* Success should redirect back to the admin participants page for the judge, with the following success messages respectively:
    * This judge has been suspended.
    * This judge is no longer suspended.
* If changing the judge's status isn't successful, probably set a generic "An error has occurred" message on the page
* Adjust the suspension message in `app/controllers/judge/scores_controller.rb` to say "Your account has been suspended, please visit our [FAQs](https://iridescentsupport.zendesk.com/hc/en-us/articles/360036011914-Rules-for-Judges) to learn more about our monitoring process."
* Add tests

## QA Steps

1. Log in as admin
2. Suspend a judge
3. Log in as the suspended judge
4. Confirm you cannot start a score with the correct message
5. Log back in as admin
6. Unsuspend the judge
7. Log in as the judge
8. Make sure you can start a score
9. Log in as an RA
10. Look up a judge, confirm you do not see the suspension controls

stenington:
Mockup screenshots for context. See https://github.com/Iridescent-CM/technovation-app/issues/2272#issuecomment-592214478

<img width="1280" alt="Screen Shot 2020-02-27 at 5 28 36 PM" src="https://user-images.githubusercontent.com/55319/75492901-bc6d7080-5986-11ea-826e-4d0dc317c3c7.png">

<img width="1280" alt="Screen Shot 2020-02-27 at 5 28 42 PM" src="https://user-images.githubusercontent.com/55319/75492905-be373400-5986-11ea-9412-2ad7c8e46b0a.png">

[...]

PR #2426, opened 2020-03-13
#2411 Give admins ability to suspend and unsuspend judges

stenington:
Here's my proposal for #2411. I think adding routes to a resource this way is a new pattern for the app, but this felt more natural to me than the alternatives I considered.


stenington:
CircleCI suddenly started complaining about duplicate keys in `circle.yml`, and indeed there were duplicate keys, so I guess that's getting fixed here too. 🤷‍♂

shaunxp20:
> CircleCI suddenly started complaining about duplicate keys in `circle.yml`, and indeed there were duplicate keys, so I guess that's getting fixed here too. 🤷‍♂

Thanks for taking care of this! @stenington 🎉 🎂 💯 🙌

shaunxp20:
>  I think adding routes to a resource this way is a new pattern for the app, but this felt more natural to me than the alternatives I considered.

I like it, makes sense to me! This post/q/a seems relevant: https://stackoverflow.com/a/35051767

config/routes.rb
```diff
@@ -280,7 +280,12 @@
     resources :scores, only: [:index, :show, :destroy]
     resources :suspicious_scores, only: :index
     resources :score_approvals, only: :create
-    resources :judges, only: :index
+    resources :judges, only: :index do
+      put 'suspend'
+      patch 'suspend'
+      put 'unsuspend'
+      patch 'unsuspend'
```
shaunxp20: Wondering if we need both put and patch here? Looks like they should be patch? Since we're updating an existing record? Also, I think we can use symbols instead of strings here. 🤔 🤓 📰 😄

shaunxp20:
I did a little bit of testing on this one, I was able to suspend/unsuspend a judge. 👍 I logged in as a suspended judge (via the admin), I think the suspension message needs to be updated? https://github.com/Iridescent-CM/technovation-app/issues/2411#issuecomment-595475373 🤔 🌮 😋 😄

config/routes.rb
stenington: I was taking my lead from the way rails will map a [put or a patch to the update method](https://guides.rubyonrails.org/routing.html#crud-verbs-and-actions), but I agree that a PATCH feels more technically correct since it's a partial update. I'll change it, and use symbols.

stenington:
Good catch, thanks. Yeah I forgot about the message.

Issue #2383, opened 2020-02-14
Add admin controls to delete scores [13,8,1.63]


stenington:
From #2368, this ticket handles adding a score deletion button for admins.

![73944688-fd5ef180-48c0-11ea-9c88-3011b28fe004](https://user-images.githubusercontent.com/55319/74547143-ab017e80-4f19-11ea-8953-5fe4c31c9918.png)

### Details & pointers
* The button should exist on the page displaying a single score's full details. Admins already have ways to reach that page.
* RAs should not see this button or be able to delete scores.
* The app uses [paranoia](https://github.com/rubysherpas/paranoia) to handle soft-deletions, see that documentation for details.
* Some summary details about scores are calculated and cached, like averages by judging round. Some/most/all? of that happens with [counter_culture](https://github.com/magnusvk/counter_culture). Add tests to ensure recalculation of these things happens when a score is deleted, if necessary.
* We haven't defined what should happen when you click **Delete this score**. I suggest it should stay on the score detail page with a flash message confirming success or explaining failure, if failure is possible. But use your judgement if something else would work better.

## QA STEPS

* Find a submission that has a completed (non-suspicious) score; or score a submission
* As an admin, view the score, you should see a delete score button (at the bottom of the score page)
* Viewing the score as any other user type should not have the delete score button

To verify the back-end delete functionality:

* Find a submission w/ multiple scores, and then delete one of the scores
* Verify the stats get updated correctly - completed scores / averages

hellafitz:
Include a confirmation popup (e.g. do you really want to delete this score?)

rsgonzal:
Deleted a score as an admin, it went away. Currently can not see deleted scores.

Tested on a submission with multiple scored, average was updated correctly.

Logged in as an RA, I do not see the option to delete score.

shaunxp20:
> Deleted a score as an admin, it went away. Currently can not see deleted scores.
> 
> Tested on a submission with multiple scored, average was updated correctly.
> 
> Logged in as an RA, I do not see the option to delete score.

Thanks @rsgonzal! 🎉 🍰 🙌

PR #2428, opened 2020-03-16
#2383 Allow admins to delete scores

shaunxp20:
This will give admins the ability to delete submitted scores. Some notes:

 * It re-uses the existing delete button/functionality for deleting suspicious scores (I moved these buttons into a partial)
 * I refactored/renamed `awaiting_review(account, &block)` to `pending_approval?`


spec/models/submission_score_spec.rb
```diff
@@ -244,6 +244,78 @@
...
+          end
+
+          it "updates the team's total number of completed scores" do
+            expect(team_submission.
+              public_send("complete_#{judging_round}_submission_scores_count")).to eq(2)
+          end
+
+          it "updates the team's total number of offical completed scores" do
+            expect(team_submission.
+              public_send("complete_#{judging_round}_submission_scores_count")).to eq(2)
```
stenington: This is checking the same thing as the previous test, I think you wanted `complete_#{judging_round}_official_submission_scores_count` right?

app/assets/stylesheets/_forms.scss
```diff
@@ -288,4 +288,8 @@ form[data-remote] {
   display: flex;
   justify-content: space-between;
   margin: 1rem 0;
-}
\ No newline at end of file
+
+  .delete-score {
```
stenington: I prefer utility classes for this kind of thing, something like `.margin--l-auto` could be added [here](https://github.com/Iridescent-CM/technovation-app/blob/qa/app/assets/stylesheets/_utilities.scss#L99), although I'm willing to do it this way too. I think utility classes are easier to reuse for this kind of tweak though. Your call! 😄

spec/models/submission_score_spec.rb
shaunxp20: Yup! Great catch, thanks @stenington! 🦅 👁 I updated this! 👍

app/assets/stylesheets/_forms.scss
shaunxp20: Thanks for the feedback, I was actually going to ask about this! I made it a utility class! 👍 What's funny though is that very recently I was looking at Tailwind and Tachyons; a couple of weeks ago I wouldn't have known the term utility class. 😄 😅 😀

stenington: I really just know it from Bootstrap, which has a pretty complete set of spacing utility classes: https://getbootstrap.com/docs/4.4/utilities/spacing/

stenington:
Changes look good 👍

PR #2429, opened 2020-03-17
#2411 don't show RAs the suspension status or controls on judge

stenington:
I realized just as I was writing up QA that RAs could see (but not actually use) the judge suspension controls, so this fixes that.

### DIFF
--- a/spec/views/admin/participants/judge_debugging_spec.rb
+++ b/spec/views/admin/participants/judge_debugging_spec.rb
@@ -13,8 +13,13 @@ RSpec.describe "admin/participants/judge_debugging", type: :view do
     )
   }
 
+  let(:current_profiles_account) {
+    FactoryBot.build_stubbed(:account)
+  }
   let(:current_profile) {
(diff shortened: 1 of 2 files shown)


spec/views/admin/participants/judge_debugging_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  let(:current_account) {
+    instance_double(
+      Account,
+      is_admin?: current_account_admin
+    )
+  }
+
+  let(:current_profile) {
+    FactoryBot.create(:judge).judge_profile
```
stenington: Couldn't figure out how to do this with an `instance_double` or `FactoryBot.build_stubbed` because of the way `Account` and `JudgeProfile` interact. But I didn't spend much time on it, so maybe it's worth trying a bit harder?

stenington: Ah, I see how it was done in `account_spec.rb`. I'll try that.

Issue #2382, opened 2020-02-14
Add admin controls to see deleted score details and optionally restore [13,13,1]


stenington:
From #2368, this ticket accomplishes two things:
* add a **View** column on the **Score export** admin page that lets you see score details for a deleted score, and
* add a button on deleted scores to restore the score

<img width="960" alt="73944686-fcc65b00-48c0-11ea-9438-340638f64cdd" src="https://user-images.githubusercontent.com/55319/74546122-0f234300-4f18-11ea-933a-dccf2a6ae6dc.png">

![73944687-fd5ef180-48c0-11ea-8746-281349b664bd](https://user-images.githubusercontent.com/55319/74546120-0e8aac80-4f18-11ea-8743-fea0fe86f76f.png)

### Details & pointers

* Add a **View** column similar to the column on the **Scores** page. You do not have to use the same icon.
* Reuse the existing view for the detailed score information.
* RAs and admins have similar-but-different views of the admin. RAs should **not** have the ability to restore deleted scores.
* The app uses [paranoia](https://github.com/rubysherpas/paranoia) to handle soft-deletions, see that documentation for details.
* Some summary details about scores are calculated and cached, like averages by judging round. Some/most/all? of that happens with [counter_culture](https://github.com/magnusvk/counter_culture). Add tests to ensure recalculation of these things happens when a score is restored, if possible.
* We haven't defined what should happen when you click **Restore this score**. I suggest it should stay on the score detail page with a flash message confirming success or explaining failure, if failure is possible. But use your judgement if something else would work better.

## QA STEPS

Prerequisite: scores that have been deleted

* Log in as an admin, navigate to the "Scores export" page (found in the navigation links on the right side)
* Search for "Include deleted scores" or "Only show deleted scores"
* From the search results, click the View icon (scroll all the way right to the last column in the table) for one of the deleted scores
* Click the "Restore this score" button
* Verify the score has been restored and the submission's score averages and numbers are adjusted correctly

hellafitz:
Add a confirmation popup. Restore flow sounds good.

stenington:
This might be a good ticket to add to the iteration, in terms of thematic cohesiveness with the other work.

shaunxp20:
This one is one staging, I added QA steps too. 👍 🌞

mdancho:
With the admin account I was able to click into score details from the judge profile page, view the score and delete it. I was then able to look up the judge and see the score was deleted, view the details again and restore it.

[...]

PR #2434, opened 2020-03-18
#2382 Allow admins to restore a deleted score

shaunxp20:
Functionality to restore a deleted score, from #2382. 😄

### DIFF
--- a/app/controllers/admin/scores_controller.rb
+++ b/app/controllers/admin/scores_controller.rb
@@ -13,7 +13,7 @@ module Admin
     }, only: :index
 
     def show
-      @score = SubmissionScore.find(params.fetch(:id))
+      @score = SubmissionScore.with_deleted.find(params.fetch(:id))
     end
 
     def destroy
@@ -21,6 +21,14 @@ module Admin
       score.destroy
       redirect_to admin_scores_path,
         success: "You deleted a score by #{score.judge_name}"
+    end
+
+    def restore
+      score = SubmissionScore.with_deleted.find(params.fetch(:score_id))
+      score.restore
+
+      redirect_back fallback_location: admin_scores_path,
+        success: "This score has been restored"
     end
 
     private
--- a/app/data_grids/scores_grid.rb
+++ b/app/data_grids/scores_grid.rb
@@ -98,4 +98,15 @@ class ScoresGrid
   column :deleted do
     deleted? ? "yes" : "no"
   end
+
+  column :view, html: true do |submission_score|
+    link_to(
+      web_icon("list-ul", size: 16, remote: true),
+      [current_scope, :score, id: submission_score.id],
+      {
+        class: "view-details",
+        data: { turbolinks: false }
+      }
+    )
+  end
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -277,7 +277,10 @@ Rails.application.routes.draw do
       controller: :regional_pitch_event_participations,
       only: :destroy
 
-    resources :scores, only: [:index, :show, :destroy]
+    resources :scores, only: [:index, :show, :destroy] do
+      patch :restore
+    end
+
     resources :suspicious_scores, only: :index
     resources :score_approvals, only: :create
     resources :judges, only: :index do
--- a/app/views/admin/scores/_score_controls.html.erb
+++ b/app/views/admin/scores/_score_controls.html.erb
@@ -3,23 +3,30 @@
     <% if score.pending_approval? %>
       <%= button_to "Approve this score",
         admin_score_approvals_path(id: score.id),
-        method: "post",
+        method: :post,
         class: "button",
         data: {
           confirm: "You are about to APPROVE this score",
           positive: "true",
         }
       %>
+  <% elsif score.deleted? %>
+    <%= button_to "Restore this score",
+      admin_score_restore_path(score),
+      method: :patch,
+      class: "button",
+      data: { confirm: "You are about to RESTORE this score" }
+    %>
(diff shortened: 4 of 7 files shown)


app/controllers/admin/scores_controller.rb
```diff
@@ -13,7 +13,7 @@ class ScoresController < AdminController
     }, only: :index
 
     def show
-      @score = SubmissionScore.find(params.fetch(:id))
+      @score = SubmissionScore.with_deleted.find(params.fetch(:id))
```
shaunxp20: I added this just to get things working/wired up; we could add some logic so only admins get the `with_deleted` scope. Right now, it looks like an RA can just punch in random ids and view any (active/non-deleted) scores, this will just extend that functionality so they can view deleted ones too. 😕 🤔 Thoughts?

app/data_grids/scores_grid.rb
```diff
@@ -98,4 +98,15 @@ class ScoresGrid
...
+  column :view, html: true do |submission_score|
+    link_to(
+      web_icon("list-ul", size: 16, remote: true),
+      [current_scope, :score, id: submission_score.id],
+      {
+        class: "view-details",
+        data: { turbolinks: false }
+      }
+    )
+  end
```
shaunxp20: This was basically just copy/pasted from the other scores grid. 😄

spec/feature/admin/restore_deleted_score_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+      round: 'quarterfinals',
+      completed_at: Time.current)
+
+    submission_score.delete
+  end
+
+  def when_the_admin_views_the_deleted_score
+    visit admin_score_exports_path("scores_grid[round]": "quarterfinals",
+                                   "scores_grid[deleted]": "with_deleted",
+                                   commit: "Search")
```
shaunxp20: I couldn't seem to interact w/ the select field on the "Scores export" page. 😢 😭 🤕 So, instead of selecting values and then clicking Search, I found that I could just pass the values in as query parameters, which isn't ideal, but works. Any tips/pointers on how to interact w/ this select field? 😸

config/routes.rb
```diff
@@ -277,7 +277,10 @@
       controller: :regional_pitch_event_participations,
       only: :destroy
 
-    resources :scores, only: [:index, :show, :destroy]
+    resources :scores, only: [:index, :show, :destroy] do
+      patch :restore
```
shaunxp20: I followed the suspend/unsuspend a judge pattern here. 😄

app/controllers/admin/scores_controller.rb
stenington: I think RA requests will route to controllers under `app/controllers/regional_ambassador/`, so this shouldn't be introducing any additional viewing power to RAs. That said, the RA controller doesn't do a better job of restricting access, so the problem you found is still relevant, but I wouldn't recommend trying to fix it here.

spec/feature/admin/restore_deleted_score_spec.rb
```diff
@@ -0,0 +1,48 @@
+require "rails_helper"
```
stenington: This file should be in `spec/features/` not `spec/feature/`

spec/feature/admin/restore_deleted_score_spec.rb
stenington: I think you can interact with the hidden elements underlying the javascript controls.

```
    visit admin_score_exports_path

    select "Quarterfinals", from: "scores_grid[round]", visible: false
    select "Include deleted scores", from: "scores_grid[deleted]", visible: false
    click_button("Search")
```
seems to work, but I don't mind going straight to the search results page the way you're doing.

spec/feature/admin/restore_deleted_score_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    find("tbody tr td.view .view-details").click
+  end
+
+  def and_clicks_the_restore_score_button
+    click_button("Restore this score")
+  end
+
+  def then_the_score_is_restored
+    expect(page).to have_content("This score has been restored")
```
stenington: Is it worth looking up the model and confirming it's not deleted any more?

app/controllers/admin/scores_controller.rb
shaunxp20: Oh right! I forgot this controller is admin-scoped! Doh! 🤦‍♂ 😊 😅 I can add a placeholder ticket for the RA issue. 👍

spec/feature/admin/restore_deleted_score_spec.rb
shaunxp20: Ah yeah, I just tried this, and it works! Thanks @stenington! I'll use this way, seems more inline w/ what a feature spec should be doing, imho! 😄 😁

spec/feature/admin/restore_deleted_score_spec.rb
shaunxp20: Ah yes! Good catch, I'll update this! 👍 💯

spec/feature/admin/restore_deleted_score_spec.rb
shaunxp20: Great idea! 💯 I'll update it to do this! 👍

app/controllers/admin/scores_controller.rb
shaunxp20: I added #2437 for the RA issue.

Issue #2433, opened 2020-03-18
Add deleted scores information to the judge profile page [?,5?,?]


stenington:
![](https://user-images.githubusercontent.com/55319/76892801-9f2f1200-6861-11ea-8776-50d5d9d1cf84.png)

When viewing the participants summary page for a judge...
* this should be for **admins only** (but I don't think RAs see the scores section at all)
* add deleted scores to both sections of the scores area, quarterfinals and semifinals
* add a column titled "Deleted?" that says yes or no for each row
* highlight deleted score rows with a pink color (exact color for you to choose)

Mockups from #2385 

## QA STEPS

Prerequisites: a judge needs to have deleted scores

* Login as an admin, and click "Judges" (from the navigation links on the right-hand side of the page)
* Find and view a judge with deleted scores
* Verify the deleted scores are being displayed
* Verify the deleted scores are only displayed to admins

shaunxp20:
This one is on staging, I added QA steps too. 👍 🌝

mdancho:
Able to see that a score has been deleted. 
<img width="1235" alt="Screen Shot 2020-03-26 at 1 40 32 PM" src="https://user-images.githubusercontent.com/52216992/77694544-6b41a400-6f67-11ea-8f5d-b0f22ebeff9d.png">

rsgonzal:
This one also works as expected!

As an RA, I can not see scores on a judge profile. When a score is deleted, it gets removed from their score page as well.

shaunxp20:
Thanks for testing this, i appreciate it! @mdancho @rsgonzal 🙇 :bowtie: 🍪 🍰 🎂

shaunxp20:
This one has been deployed to production! 👍

PR #2439, opened 2020-03-24
#2433 Add deleted score info to judge profile admin page

shaunxp20:
This will allow admins to view deleted scores on a judge's profile page.

#2433

### DIFF
--- a/app/views/admin/participants/_judge_scores.html.erb
+++ b/app/views/admin/participants/_judge_scores.html.erb
@@ -1,23 +1,25 @@
 <% if scores.any? %>
-  <table>
+  <table class="width--100-percent">
     <th>
       <tr>
         <td>Team Name</td>
         <td>Submission Name</td>
         <td>Score</td>
-        <td>Details </td>
+        <td>Deleted</td>
+        <td>Details</td>
       </tr>
     </th>
     <tbody>
       <% scores.each do |score| %>
-        <tr>
+        <tr class="<%= score.deleted? ? "background-color--subtle-red" : "" %>">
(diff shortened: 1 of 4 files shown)


app/views/admin/participants/_judge_scores.html.erb
```diff
@@ -1,23 +1,25 @@
...
         <td>Score</td>
-        <td>Details </td>
+        <td>Deleted</td>
+        <td>Details</td>
       </tr>
     </th>
     <tbody>
       <% scores.each do |score| %>
-        <tr>
+        <tr class="<%= score.deleted? ? "background-color--subtle-red" : "" %>">
```
shaunxp20: I thought about putting this into a helper, but I couldn't figure out a good name for it! 😬 😖

app/views/admin/participants/_judge_scores.html.erb
```diff
@@ -1,23 +1,25 @@
...
           <td><%= score.team_name %></td>
           <td>
             <%= link_to score.team_submission_app_name,
                         admin_team_submission_path(score.team_submission_id),
                         class: "cta-link" %>
           </td>
-          <td> <%= score.total %> / <%= score.total_possible %> </td>
+          <td><%= score.total %> / <%= score.total_possible %></td>
+          <td class="deleted-info"><%= score.deleted? ? "yes" : "no" %></td>
```
shaunxp20: There's seems to be a dozen different ways/answers for how to turn a boolean into yes/no, I didn't know which direction to take! And it seems that this way is common in our app, so 🤷‍♂ 😕 🤕

Issue #2422, opened 2020-03-10
Rake task for resetting surveys (5,8,0.63)


stenington:
Surveys on the platform do things like track how many times a user was reminded and whether the user finally took it or not, and those things don’t automatically reset. That means that users who took a previous survey, or saw the reminder the max number of times, will not be alerted as clearly on the platform when a new survey goes up. #2310 has more details.

As a first step in simplifying this, we could write a rake task or set of tasks that reset surveys for a particular type of user. [This documentation](https://docs.google.com/document/d/1hVaugqKipnD8pYmCcMl5KWCHrrqBBSr-86MQZTdm5P8/edit#heading=h.2yry0iduh9ht) has details on what that would look like.

### QA Steps
The solution for this ticket introduces 3 new tools: one to reset students only, one to reset mentors only, one to reset everyone. It also resets both users who aren't shown a survey because they took it, as well as users who aren't shown the survey pop-up because they've been reminded too many times. 

It will be challenging on staging to set up users who have been shown the pop up too many times, because there's a 2 day cooldown between each time it's shown. Here are steps for the user-took-the-survey scenario, and we can discuss how to convince ourselves the other scenario works as well.

When a step says "set up a student/mentor as having taken the survey", log in as that type of user and then go to https://technovation-qa.herokuapp.com/survey_completion directly. You should see a green message at the top of the dashboard thanking you for taking the survey. 

**With surveys configured for both mentors and students...**

Student reset:
1. set up a student as having taken the survey
2. set up a mentor as having taken the survey
3. have a developer run the **student reset task** 
4. log in as the student and confirm the survey pop-up appears
5. log in as the mentor and confirm the survey pop-up does not appear
    
Mentor reset:
1. set up a student as having taken the survey
2. set up a mentor as having taken the survey
3. have a developer run the **mentor reset task**
4. log in as the student and confirm the survey pop-up does not appear
5. log in as the mentor and confirm the survey pop-up appears

Reset all:
1. set up a student as having taken the survey
2. set up a mentor as having taken the survey
3. have a developer run the **reset task for all users**
4. log in as the student and confirm the survey pop-up appears
5. log in as the mentor and confirm the survey pop-up appears

rsgonzal:
Tested these scenarios with Mike and all looks great!

shaunxp20:
This one has been deployed to production! 👍

PR #2441, opened 2020-03-24
#2422 add survey reset task and tests

stenington:
It turns out `needs_survey_reminder?` and `took_program_survey?` are used in tandem to determine if a user should be shown a survey reminder. Specifically, `needs_survey_reminder?` doesn't check at all whether the user has taken the survey, which tripped me up for a while. 

There are also some magic numbers in the test setup that I'm not wild about, but the max number of survey reminders and the period after registering where you're not shown the survey are both declared privately in `account.rb`.


stenington:
I'm not sure there's a **good** reason for doing it the way we're doing it, but there are reasons. 😬

Issue #2445, opened 2020-03-25
Expand test coverage for critical platform functions [40,20,2]


hellafitz:
from #2414 and epic #2166

Updated groundwork ticket (est. effort 20):
- identify "critical" platform functionality (e.g. submissions)
- is there an integration test that covers it? 
- if yes it's critical, and no, there's not an existing test, write the test

stenington:
As a first step, I wanted to update capybara to get rid of test deprecation warnings and make the test suite a little easier to work with. The upgrade resulted in a few unexpected test failures.

It seems like we had a few tests loading pages with Vue components without javascript, and then asserting things about the uninitialized Vue templates. Those tests seem to fail with capybara 3.x, which is a good thing. Running those tests with javascript (and potentially making a few extra tweaks) gets them passing again.

stenington:
A couple of notes for posterity:
* if a non-js test is failing because it can't find an element, but you can see the element in e.g. a `puts page.html`, check that it's not embedded in a Vue component
* if you convert a test to use javascript and suddenly it becomes flaky, turbolinks might be getting in your way
    * with turbolinks active, the url can change before the page does so looking at the url doesn't guarantee page content has updated
    * add an assertion for some piece of page content ( a header, or a link, etc.) that lets you know the next page loaded

stenington:
Working on mapping from the javascript packs to a rough idea of where that code appears on the site [here](https://docs.google.com/spreadsheets/d/1_Q2pwfgq3gkzLROhmEQk1TWO6FJSjH0920ejqTYHTvA/edit#gid=0)

PR #2447, opened 2020-03-26
#2445 tests

stenington:
Trying to get rid of deprecation warnings by upgrading capybara.


stenington:
@shaunxp20 I probably should have had you review my initial changes without adding any new tests, so let me know if you want me to roll back my latest commits. But this PR updates capybara to get rid of the deprecation warnings, fixes a few tests that broke as a result of that upgrade, and introduces a new system spec for students submitting.

spec/system/student/submit_a_project_spec.rb
```diff
@@ -0,0 +1,94 @@
+require "rails_helper"
```
shaunxp20: Regarding our convo about what goes in `feature/` vs `system/`. Just making this up as I go, but I wonder if this style of test goes in `feature/`? Anywho, just a suggestion to help draw that line. #no-big-deal #random-thought 😃

spec/system/student/submit_a_project_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+    given_there_is_a_team_with_a_complete_submission
+    that_hasnt_been_submitted
+    and_a_student_from_that_team_is_logged_in
+
+    when_they_review_it_from_the_dashboard
+    and_submit_it
+
+    then_they_see_that_their_app_will_be_judged
+    and_it_is_marked_as_published
+  end
```
shaunxp20: Woah! I really like the way these specs read!!!! :100: 😍 🙌 🏅 💯

spec/system/student/submit_a_project_spec.rb
stenington: I found [this post](https://rossta.net/blog/why-rails-system-tests-matter.html) discussing it. It looks like system tests are a new concept with rails 5.1, and they allow capybara tests to run in transactions which previously wasn't possible. So it seems to me like the major difference is a technical aspect of database cleanup.

I'm thinking of moving this test to `feature/` and defaulting to putting stuff there.

Issue #2444, opened 2020-03-25
Reduce word minimum on judge feedback [8,5,1.6]


rsgonzal:
Judges are currently required to leave at least 40 words of feedback on each section of the judging rubric.

"People are copy and pasting the same test or giving poor feedback just to be able to submit their score. We do not need to force judges to have to leave more than 20 words of feedback. "

Proposal to reduce word minimum to 20 words in all sections, EXCEPT for the 'Overall Impression' section which should stay at 40 words.

## QA STEPS

* As a judge, score a submission
* Verify at least 20 words are required for the following sections: 
  * Ideation
  * Technical
  * Pitch
  * Entrepreneurship
* Verify at least 40 words are required for the "Overall Impression" section 
* Verify you can finish/submit the score

shaunxp20:
This one is on staging, I added QA steps too! 😃

nickiredescent:
Tested. Can confirm that the word minimums for each section are as described.

rsgonzal:
@shaunxp20 not sure what's going on with my scoring. it won't let me complete and says there's something wrong with ideation. i'll tag in slack

shaunxp20:
> @shaunxp20 not sure what's going on with my scoring. it won't let me complete and says there's something wrong with ideation. i'll tag in slack

Great catch @rsgonzal! 🥇 There are additional validation checks on the last/review page, doh! I just updated those. I'll let you know when it's on staging again. 👍 😃

shaunxp20:
@rsgonzal This one is back on staging, the issue you found should be fixed now. 🤞 😀

rsgonzal:
thanks, @shaunxp20 ! Looks good to me! I asked Mackenzie to take a final look.

mdancho:
All good! Confirmed that all sections require a minimum of 20 words before being able to submit, except overall which required 40. Also required numerical scores for each section before being able to finish the score.

PR #2448, opened 2020-03-26
#2444 Adjust minimum word count for judge feedback form

shaunxp20:
This will change the minimum word count for the "Overall Impression" section to 40, and for all other sections it will be 20.

#2444

### DIFF
--- a/app/javascript/judge/scores/QuestionSection.vue
+++ b/app/javascript/judge/scores/QuestionSection.vue
@@ -29,7 +29,7 @@
 
             <br>
 
-            <span class="font-weight--300">Please write at least 40 words</span>
+            <span class="font-weight--300">Please write at least {{ minWordCount }} words</span>
           </p>
         </div>
       </div>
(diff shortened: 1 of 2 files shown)


shaunxp20:
Btw, this one is functionally complete, but I still want (to try) to write some more tests. It seems like there will be room at the end of the iteration I was planning to use some of that time to look more into Vue/Vue testing. 😃

app/javascript/judge/scores/QuestionSection.vue
```diff
@@ -148,12 +152,12 @@ export default {
...
-      if (this.wordCount >= 40) {
+      if (this.wordCount >= this.minWordCount) {
         return 'darkgreen'
-      } else if (this.wordCount >= 30) {
+      } else if (this.wordCount >= (66 / 100) * this.minWordCount) {
+        return 'goldenrod'
+      } else if (this.wordCount >= (33 / 100) * this.minWordCount) {
         return 'orange'
-      } else if (this.wordCount >= 20) {
-        return 'darkyellow'
```
shaunxp20, line 156: Btw, just fyi, there isn't a named color called "darkyellow".

https://developer.mozilla.org/en-US/docs/Web/CSS/color_value

So, basically, this part of the code didn't work as intended. 😬 😅 Anyways, I updated it while I was in here. 😃

Issue #2425, opened 2020-03-13
Allow Expertise column in admin and RA exports of Participants [5,5,1]


hellafitz:
We ask mentors to indicate their areas of expertise on signup but RAs don't have access to that data when they're trying match mentors and teams. Add the ability to include Expertise as a csv column when exporting from the Participant page. 

During QA, make sure to check the csv export, not just the page

This feature request originated from Ophelie in the Platform Issues spreadsheet. Be sure to close the loop with her. 

## QA STEPS

* Login as an admin, click the "Participants" link
* Verify there is a new option for "Mentor expertise" in the "MORE COLUMNS" dropdown
* Verify "Mentor expertise" displays in the search results table, and in the CSV
* Verify this functionality as an RA too

hellafitz:
- send a screenshot mockup to Ophelie to confirm approach
- implement (but if the approach is widely divergent this will need a new ticket)

shaunxp20:
@ophelieh Here are some screenshots/mockups, let me know if this is what you had in mind for this! 😃 

### Here's "Mentor expertise" as a new selectable column:
<img width="1103" alt="Screen Shot 2020-03-26 at 5 43 12 PM" src="https://user-images.githubusercontent.com/58957909/77707560-64458080-6fbd-11ea-8647-3552eb768cc7.png">

### Here's "Mentor expertise" in the search results:
<img width="1118" alt="Screen Shot 2020-03-26 at 5 43 50 PM" src="https://user-images.githubusercontent.com/58957909/77707709-da49e780-6fbd-11ea-8805-487e6d9930e0.png">

And here's "Mentor expertise" in a the CSV file:
<img width="711" alt="Screen Shot 2020-03-26 at 5 46 58 PM" src="https://user-images.githubusercontent.com/58957909/77707567-690a3480-6fbd-11ea-9a45-daebc1840676.png">

Let me know what you think!

ophelieh:
Hi @shaunxp20 this is perfect! Thank you :)

shaunxp20:
This one is on staging, and I added QA steps too. 👍

ophelieh:
Tested as an admin and an RA - both worked as expected. 

Thank you!

shaunxp20:
Thanks @ophelieh! 👍 🙇 😀  I'll move this one to the Ready to Deploy column. 👍

PR #2449, opened 2020-03-27
#2425 Add mentor expertise to participants page

shaunxp20:
This will make "Mentor expertise" a selectable column for admins and RAs to optionally view on the participants page. It will also be on the CSV report.

#2425

### DIFF
--- a/app/data_grids/accounts_grid.rb
+++ b/app/data_grids/accounts_grid.rb
@@ -25,6 +25,14 @@ class AccountsGrid
     end
   end
 
+  column :mentor_expertise do
+    if mentor_profile.present?
+      mentor_profile.expertise_names.join(",")
+    else
+      "-"
+    end
+  end
+
   column :judge_industry do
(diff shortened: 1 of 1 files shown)


shaunxp20:
I didn't see any existing tests. I could add some, I could go either way. 🤔 😐 🤷‍♂

shaunxp20:
> This is pretty much what I had in mind.
> 
> I don't think it's worth writing an entire test suite just for this change.
> 
> Also, it's possible this won't export well. I believe `expertise_names` has to join to an expertise table, which isn't too bad on a page with 20 results or whatever, but trying to do an export of 100s of records including this column may be a memory hog, I'm not sure.
> 
> But again that's an endemic problem in the admin right now, so I don't think there's anything to be done in this PR.
> 
> With all those caveats, this looks 👍!

Makes sense! Thanks for the feedback/review! 👍 🌞 I'm going to merge this one in! :shipit: 😀

Issue #2418, opened 2020-03-10
Rake task and tests for unsubmitting [5,5,1]


stenington:
We have identified "unsubmitting" submitted but incomplete submissions as a potential developer task this season. This ticket would be for creating a rake task that "unsubmits" a submission or group of submissions based on id or list of ids, with tests. 

[This documentation](https://docs.google.com/document/d/1BEHSSY9gSwKpUNJUZaVSwcixPC6HRPqNVy_33t9p7Ow/edit#bookmark=id.jha3lxost4so) provides some hints about how to unsubmit, and should be updated as part of this ticket as well.

### QA Steps

To create a submission that is submitted but not complete, you will first have to complete it, then submit it, then the platform will allow you to continue editing and you can do something like blank out the app name and save.

In the admin under Submissions, adding "Complete?" and "Submitted?" as _More Columns_ is a good way of confirming your setup and the results of the rake task.

1. Find or create two different submissions that are submitted, but are not complete
2. Ask a dev to run `submissions:submitted_but_incomplete` and confirm that both apps show up in the output
3. Ask a dev to unsubmit one of them with `submissions:unsubmit![<id>]`
4. Confirm that the one you picked has been unsubmitted, while the other hasn't changed

stenington:
Unsubmitting is something we've identified as a possible scenario, but we aren't expecting many teams, if any, to end up in this state. As a manual task, it's not particularly painful. (We're just trying to avoid having to do these sorts of tasks manually, period.)

stenington:
Just noting that the language around a "complete" or "published" submission is sloppy in the code:

* the `complete` scope only looks to see if `published_at` is set
* a submission is `complete?` if all required fields are complete AND it's published
* a submission is `published?` if `published_at` is set
* and `percent_complete` is only 100% if the app has been published

This leads to strange code like
```
TeamSubmission.complete.find_each do |sub|
    if !sub.complete?
        ...
    end
end
```
where you wouldn't think you'd find an not-complete submissions when you've scoped to complete, but you do.

rsgonzal:
Tested with Mike and all is working as expected!

PR #2451, opened 2020-03-27
#2418 add rake task to help with unsubmitting submissions

stenington:
This adds two new rake tasks, `submissions:submitted_but_incomplete` and `submissions:unsubmit![<ids>]`

The first lists out the submissions that it thinks should be unsubmitted, and gives a copy/paste-able snippet for the `unsubmit!` task if you want to fix them all.

The `unsubmit!` task unsubmits all Submissions specified by id, whether they're complete or not.

I think these can be used in tandem with the admin to unsubmit (aka unpublish) submissions if we need to do that.


stenington:
I was thinking this pattern of one task to identify records and another to change them could be used to break apart the `fix_submissions` task as well, if it seems like a more comfortable pattern.

spec/tasks/submissions_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+      sub
+    }
+    let!(:past) {
+      sub = FactoryBot.create(:submission, :half_complete)
+      sub.publish!
+      sub.update(seasons: [Season.current.year - 1])
+      sub
+    }
+
+    it "lists current submitted but incomplete app" do
```
shaunxp20: Minor quibble at the lingo here, it says "submitted" but the expectation references a "published" submission.

I saw your comment in #2418, so it makes sense to me.

> Further muddying the water is the language of "submitting" or "un-submitting", which translates into the code as publishing or un-publishing, really.

Wondering if we can mention something like that^ in the `it` or somehow make that part clearer in the specs? 🤔 😃

stenington: Do you have a suggestion for what would make it clearer? I could change `:incomplete_but_published` to be `:incomplete_but_submitted` so the test name and assertion line up a little better. That moves the published/submitted language discrepancy into the data setup where perhaps it's not as jarring?

Another option is to stick with "published" naming throughout the tests (and maybe even the rake task) and let the translation from "submitted" to "published" be something we document elsewhere, in a project glossary or something (although that doesn't currently exist). 

Or the other extreme might be to alias `publish!` and `published?` with `submit!` and `submitted?` equivalents, although my gut tells me the flexibility that would introduce might just make things more difficult in the long run.

I'd vote for option 1 if it feels like it would clarify anything, otherwise maybe we need a quick discussion about how to get the terminology straight, because it's definitely confusing.

shaunxp20: >  I could change `:incomplete_but_published` to be `:incomplete_but_submitted` so the test name and assertion line up a little better.

I think something as simple as that would help (imho)! 😁 👍 

And/or I wonder if mentioning in the `it` something like submitted/published or submitted (aka published) or something like that would help or just make it more confusing! 🤕 😃 

Anyhow, I also vote for option 1, or something that involves very little/minor changes. 😄

stenington: Cool, that should be easy enough! Thanks

PR #2456, opened 2020-04-01
#2445 additional critical specs

stenington:
Continuing with #2445, adding specs for critical scenarios.

### DIFF
--- a/spec/features/judge/starting_scores_spec.rb
+++ b/spec/features/judge/starting_scores_spec.rb
@@ -1,0 +1,133 @@
+require "rails_helper"
+
+feature "starting scores", js: true do
+
+  context "as an RPE judge" do
+    before do
+      SeasonToggles.judging_round = :qf
+    end
+
+    after do
+      SeasonToggles.judging_round = :off
+    end
+
+    scenario "TBD"
+  end
+
+  context "as a virtual judge" do
+    before do
+      SeasonToggles.judging_round = :qf
+    end
+
+    after do
+      SeasonToggles.judging_round = :off
+    end
+
+    scenario "starting a new score" do
+      given_there_is_a_submission_from_a_junior_team_that_needs_scoring
+      and_a_user_is_logged_in_as_a_judge
+
+      when_the_judge_views_their_dashboard
+      and_starts_a_new_score
+
+      then_they_see_the_scoring_screens_for_that_submission
+    end
+
+    scenario "unable to start new score with a score in progress" do
+      given_there_is_a_submission_assigned_to_a_judge
+      and_that_judge_is_logged_in
+
+      when_the_judge_views_their_dashboard
+
+      then_they_will_not_see_a_way_to_start_a_new_score
+    end
+
+    scenario "continuing with an unfinished score" do
+      given_there_is_a_submission_assigned_to_a_judge
+      and_that_judge_is_logged_in
+
+      when_the_judge_views_their_dashboard
+      and_resumes_scoring
+
+      then_they_see_the_scoring_screens_for_that_submission
+    end
+
+    scenario "reviewing a completed score" do
+      given_there_is_a_judge_with_a_completed_score
+      and_that_judge_is_logged_in
+
+      when_the_judge_views_their_dashboard
+      and_navigates_to_their_finished_scores
+      and_reviews_their_completed_score
+
+      then_they_see_the_scoring_screens_for_that_submission
+    end
+  end
+
+  private
+
+  def given_there_is_a_submission_from_a_junior_team_that_needs_scoring
+    @submission = FactoryBot.create(:team_submission, :junior, :complete)
+  end
+
+  def and_a_user_is_logged_in_as_a_judge
+    judge = FactoryBot.create(:judge, :onboarded)
+
+    sign_in(judge)
+  end
+
+  def given_there_is_a_submission_assigned_to_a_judge
+    @judge = FactoryBot.create(:judge, :onboarded)
+    score = FactoryBot.create(:score, :incomplete, :quarterfinals, judge_profile: @judge)
+    @submission = score.team_submission
+  end
+
+  def and_that_judge_is_logged_in
+    sign_in(@judge)
+  end
+
+  def given_there_is_a_judge_with_a_completed_score
+    @judge = FactoryBot.create(:judge, :onboarded)
+    score = FactoryBot.create(:score, :complete, :quarterfinals, judge_profile: @judge)
+    @submission = score.team_submission
+  end
+
+  def when_the_judge_views_their_dashboard
+    click_link "My Dashboard"
+    expect(page).to have_content "Judge Dashboard"
+  end
+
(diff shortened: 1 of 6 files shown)


stenington:
This one was failing like crazy yesterday, not always the same test either, and today it's passing on run after run. Maybe CI was running much more slowly yesterday or something?

Anyway, it introduces a feature spec for judges starting scores, which differs between RPE judges and virtual judges, as well as depending on things like scores in progress.

It also sets some new Capybara config: making the browser bigger for screenshots or headful selenium drivers, making it possible to specify selenium drivers from the commandline, and dumping the browser log to STDERR on failure. All were useful to me while working on these so thought I'd add them to the PR.

stenington:
@shaunxp20 On this branch, `JAVASCRIPT_DRIVER=selenium` was sufficient for me to get the tests to run in FF. Does that not work for you? If not, let me know what you have to do so I can figure out how to accommodate it.

spec/features/judge/starting_scores_spec.rb
```diff
@@ -0,0 +1,223 @@
...
+    scenario "no button to score virtual submissions" do
+      given_there_is_a_logged_in_judge
+      and_they_are_assigned_to_an_RPE
+
+      when_the_judge_views_their_dashboard
+
+      then_they_will_not_see_a_way_to_start_a_new_virtual_score
+    end
+
+    scenario "able to start scoring an assigned submission" do
```
shaunxp20, line 32: Noob here, the "assigned submission" lingo here is tripping me up. How is the judge assigned the submission? From the other scenarios it looks like the judge can score any submission from the RPE? 🤔 😕 🤔

spec/features/judge/starting_scores_spec.rb
```diff
@@ -0,0 +1,223 @@
...
+      given_there_is_a_logged_in_judge
+      and_they_are_assigned_to_an_RPE
+      and_the_RPE_has_a_submission
+      and_the_judge_has_a_completed_score_for_the_submission
+
+      when_the_judge_views_their_dashboard
+      and_reviews_their_completed_score
+
+      then_they_see_the_scoring_screens_for_that_submission
+    end
```
shaunxp20, line 75: Btw, loving the way all these feature specs read! 💯 😍

spec/rails_helper.rb
```diff
@@ -21,7 +21,7 @@
 ::Timezone::Lookup.lookup.default("America/Los_Angeles")
 
 Capybara.automatic_label_click = true
-Capybara.javascript_driver = :selenium_chrome_headless
+Capybara.javascript_driver = ENV.fetch("JAVASCRIPT_DRIVER", "selenium_chrome_headless").to_sym
```
shaunxp20, line 24: 👍 🙌

spec/rails_helper.rb
```diff
@@ -70,6 +70,22 @@
   end
 
   config.before(:each, type: :system, js: true) do
-    driven_by :selenium_chrome_headless
+    driven_by ENV.fetch("JAVASCRIPT_DRIVER", "selenium_chrome_headless").to_sym
```
shaunxp20, line 73: 💯 🎉

spec/rails_helper.rb
```diff
@@ -70,6 +70,22 @@
...
+  end
+
+  config.before(:each, js: true) do
+    Capybara.page.driver.browser.manage.window.resize_to(1200, 1200)
```
shaunxp20, line 77: Btw, I didn't try it, but we might be able to combine this w/ the previous `before(:each)`:
> The driven_by method takes three optional keyword arguments, :using, which specifies the browser for selenium, as in driven_by :selenium, using: :firefox, and :screen_size, which specifies the size of the browser window as a two-element array: driven_by :selenium, screen_size: [640, 480]. There’s also an :options keyword which is passed to the driver if the driver requires other options.

https://medium.com/table-xi/a-quick-guide-to-rails-system-tests-in-rspec-b6e9e8a8b5f6

spec/rails_helper.rb
```diff
@@ -70,6 +70,22 @@
...
+  end
+
+  config.after(:each, js: :true) do |example|
+    if example.exception
+      save_and_open_screenshot if ENV.fetch("OPEN_SCREENSHOT_ON_FAILURE", false)
+      save_and_open_page if ENV.fetch("OPEN_PAGE_ON_FAILURE", false)
```
shaunxp20, line 83: Nice additions! 💯 🍪 🏆

spec/system/mentor/location_spec.rb
```diff
@@ -15,7 +15,7 @@
 
     select_gender(:random)
 
-    fill_in_vue_select "School or company name", with: "John Hughes High"
+    fill_in "School or company name", with: "John Hughes High"
```
shaunxp20, line 18: 👍

shaunxp20:
> @shaunxp20 On this branch, `JAVASCRIPT_DRIVER=selenium` was sufficient for me to get the tests to run in FF. Does that not work for you? If not, let me know what you have to do so I can figure out how to accommodate it.

Yes! This does work! Thanks for adding this option/feature!!! 👍

spec/features/judge/starting_scores_spec.rb
stenington: Ah, yeah, I'll reword. I was just reusing the language on the page, which has a tab labeled "Assigned submissions" that lists the RPE submissions.

spec/rails_helper.rb
stenington: Nope, `driven_by` is only available in system specs. A feature spec would blow up.

stenington: Well, I guess you can still check the spec type from within the block, so it could be consolidated down that way.

spec/features/judge/starting_scores_spec.rb
stenington: Actually there's some nuance here. If an RPE is big enough, the RA can assign a subset of the submissions to a judge, and then calling them the "assigned submissions" is a bit more clear. In this case however, because the RPE is small, it sort of defaults to all submissions being assigned to all judges.

I'm happy to update the test titles here to say "RPE submissions" instead of "assigned submissions" or something like that, but I wonder if I should also break the "as an RPE judge" scenario out into "as a judge at a small RPE" and "as a judge at a large RPE" or something, or if testing that assignment functionality more clearly belongs somewhere else?

shaunxp20: Ah, good to know!

+1 for making the that distinction (big RPE -> assigned vs small RPE -> all) clearer, and documenting it (via some specs). Maybe just make this PR for small RPEs and do big RPEs in another PR? I'm fine w/ just leaving this as-is and making those changes in another PR too, either way. 👍

stenington: Ok, I added a comment in #2461 about expanding on these but will go with them as they are in this PR.

PR #2469, opened 2020-04-07
#2444 Adjust minimum word count for judge feedback

shaunxp20:
Apparently there are three separate word count validations for judge feedback:
 * one for each section
 * one for the "Finish score" button on the review page
 * one for the warning icons on the review page

I covered the first bullet point in my previous PR #2448, this PR will cover the remaining bullet points. 😳 😅 

#2444

### DIFF
--- a/app/javascript/judge/scores/store/actions.js
+++ b/app/javascript/judge/scores/store/actions.js
@@ -16,10 +16,11 @@ export const validateScore = ({ commit, state }) => {
 
   Object.keys(state.score.comments).forEach((section) => {
     const comment = state.score.comments[section]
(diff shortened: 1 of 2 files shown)


shaunxp20:
I would like to DRY this up, but not sure how; I'm thinking it would take a major refactoring (and for myself a better understanding of Vue.js, and how this is all wired together). 😕 🤷‍♂ 🔰

stenington:
I don't think it would take a major refactor, I think we would just need to introduce a helper function for determining the minimum word count, but that said I don't know where you'd expect that sort of thing to live in a standard Vue/Vuex project or the exact details of pulling it in where you'd need it, so I'm happy to let this stand as is for the time.

PR #2475, opened 2020-04-15
Test improvements

shaunxp20:
Guess I'm at a stopping point and ready to get some feedback on this one. 😅


spec/models/account_spec.rb
```diff
@@ -106,7 +106,7 @@
     account = FactoryBot.build(:account, email: "email@gmail.com")
     expect {
       account.save(validate: false)
-    }.to raise_error
+    }.to raise_error(ActiveRecord::RecordNotUnique)
```
shaunxp20, line 109: This was to fix this warning when running our specs:
```
WARNING: Using the `raise_error` matcher without providing a specific error or message risks false positives, since `raise_error` will match when Ruby raises a `NoMethodError`, `NameError` or `ArgumentError`, potentially allowing the expectation to pass without even executing the method you are intending to call. 
```

spec/rails_helper.rb
```diff
@@ -21,13 +21,14 @@
...
 Capybara.automatic_label_click = true
+Capybara.default_max_wait_time = 5
 Capybara.javascript_driver = ENV.fetch("JAVASCRIPT_DRIVER", "selenium_chrome_headless").to_sym
 
 require 'rake'
 Rails.application.load_tasks
 
 RSpec.configure do |config|
-  config.fail_fast = true
+  config.fail_fast = ENV.fetch("RSPEC_FAIL_FAST", true)
```
shaunxp20, line 31: It looks like this can be set to a number too, like fail after x number of failures.
https://relishapp.com/rspec/rspec-core/docs/configuration/fail-fast

I added it as an ENV so I can set it to `false` so I can run the entire test suite (regardless of failures).

spec/system/admin/invite_users_spec.rb
```diff
@@ -66,16 +66,16 @@
         fill_in "School name", with: "John Hughes High"
```
shaunxp20, line 66: This was one of our flaky specs, I made a couple of tweaks and it looks like it's more stable now. If I'm seeing things right, it hasn't failed since build [#6147](https://circleci.com/gh/Iridescent-CM/technovation-app/6147), which is almost 100 builds ago!

spec/system/submissions/upload_source_code_spec.rb
```diff
@@ -91,6 +91,8 @@
         fill_in "What is the shareable URL of your Thunkable project?",
```
shaunxp20, line 91: This was another flaky one, I'm not really a fan of the `sleep` band-aid, but it seems have helped, it hasn't failed in over 50 builds.

There's a blurb in this article about sleeping in feature specs:
http://johnpwood.net/2015/04/23/tips-and-tricks-for-dubugging-and-fixing-slowflaky-capybara-specs/

I guess I can agree and get onboard w/ the: use sleeps as an exception, not the rule mindset.

spec/tasks/check_onboarding_spec.rb
```diff
@@ -1,6 +1,16 @@
...
+  let(:output) { StringIO.new }
+
+  before(:each) do
+    $stdout = output
+  end
+
+  after(:each) do
+    $stdout = STDOUT
+  end
+
```
shaunxp20, line 13: This was just a copy/paste from our other rake task specs to remove the "Student#XXX should be onboarded" text that would output in our specs.

spec/features/judge/starting_scores_spec.rb
```diff
@@ -3,8 +3,8 @@
 feature "starting scores", js: true do
```
shaunxp20, line 3: This was another flaky one, these changes seemed to have helped, but it still fails, just not as frequently. It successfully passed for about 20 runs before it failed, and maybe it's more of a fluke/anomaly now when it fails, in the last failure I see a mention cloud_flare, and a bunch of `ga` related stuff, which seems to be abnormal.

shaunxp20:
@stenington When you have a moment, can you give me a sanity check on using `JAVASCRIPT_DRIVER` with Firefox. I thought it was working for me, but now it seems to just throw errors. When I set it to `selenium`, it gives me the error:
```
Webdrivers::BrowserNotFound:
Failed to find Chrome binary.
```
I can only seem to get it to work by adding an extra option to `driven_by`:
`using: :firefox`

😕 🤔 🤕

spec/rails_helper.rb
stenington: 💯

spec/models/account_spec.rb
stenington: 👍

spec/system/admin/invite_users_spec.rb
stenington: 🎉

spec/system/submissions/upload_source_code_spec.rb
stenington: Yeah, I'm a little confused as to why the sleep would actually help here, but I guess it's giving some javascript time to finish or something? I can live with this.

shaunxp20:
> You're right, `selenium` is now opening chrome for me. The capybara docs say it should open Firefox, but I guess we need to rework this.

Thanks for double checking! Good to know it's not local to my setup/machine, it was driving me bonkers, b/c I could have swore it was working before! 🤷 😃 

> I'm good with merging this and continuing to tinker, or making more changes here. Your call.

Cool, I'll get this one merged in, which hopefully will get our test suite to run/pass more reliably/consistently. 🤞 😅 😁

Issue #2458, opened 2020-04-03
Don't display 'View Your Certificate' if no certificate is available [5,5,1]


hellafitz:
from #2271

![Screen Shot 2020-04-03 at 12 12 25 AM](https://user-images.githubusercontent.com/22801026/78334619-ec3d0480-7540-11ea-9d80-8723149111a8.png)

The "View your certificate" button shouldn't appear if the judge has no certificates. This is probably a hangover from when judge's would be shown a button that generated the certificate when clicked. I suspect [this is the offending line](https://github.com/Iridescent-CM/technovation-app/blob/qa/app/guards/display_certificates.rb#L6) at this point.

### QA STEPS
Prerequisites: the "Scores & Certificates" setting needs to be enabled

* As an admin, find a judge without a certificate; or find one with a certificate (via the Certificates tab/page) and then suspend this judge
* Then impersonate the judge and make sure the judge doesn't see a "View your certificate" button on their dashboard

shaunxp20:
Btw, I see another file/message for when an account doesn't have a certificate: [app/views/dashboards/_no_certificates.erb](https://github.com/Iridescent-CM/technovation-app/blob/qa/app/views/dashboards/_no_certificates.erb) which mentions the finalists, and a link to stay informed.

I wonder...
a) if we should incorporate some of this into the page when a judge/account does have a certificate
b) if the program team would want to change/update this content for this season, and/or if CET would want to update this for judges this season

🤔 😃

stenington:
It's unclear to me whether a judge with no certificates should see `app/views/judge/dashboards/onboarded/_certificates.en.html.erb` with no button, or whether they should see `app/views/dashboards/_no_certificates.erb`. I was assuming we'd delete the line in `DisplayCertificates` mentioned in the ticket description, which I think would accomplish the latter. But if the first option is the desired choice, then @shaunxp20 your PR seems like the way to go, although I wouldn't mind adjusting `DisplayCertificates` anyway to get rid of the old, unnecessary logic.

shaunxp20:
I made some changes based off our meeting earlier today, which if I understood correctly:
 - if `display_scores` is turned on and a cert exists for the judge -> display the certs page
 - if `display_scores` is turned on and a cert does not exist for the judge -> display the no certs page

Hopefully I got that right! 😸 😅 

I also added a condition for this edge case:
 - if `display_scores` is turned on, and a cert does exist for the judge, but the judge is suspended -> display the no certs page

[...]

PR #2481, opened 2020-04-21
Conditionally show judge's certificate page

shaunxp20:
I updated the logic so that in order for a judge to see the certificate's page:
 * SeasonToggles.display_scores has to be turned on (true)
 * a certificate has to exist for the judge
 * the judge can't be suspended (this will handle the edge case where certificates are generated, and then the judge is suspended)

#2458

### DIFF
--- a/app/guards/display_certificates.rb
+++ b/app/guards/display_certificates.rb
@@ -1,9 +1,0 @@
-require "./app/constants/badge_levels"
-
-module DisplayCertificates
-  def self.for(judge)
-    SeasonToggles.display_scores? &&
-      (judge.current_complete_scores.count >= NUMBER_OF_SCORES_FOR_CERTIFIED_JUDGE ||
-        judge.current_judge_certificates.any?)
-  end
-end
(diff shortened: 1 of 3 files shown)


shaunxp20:
@stenington When you get a moment, this one is ready for another looksie, hopefully I got it right this time! 🤞 😁

stenington:
Just curious if there's something in particular you dislike about the `DisplayCertificates` guard module. It seems like a good idea to me to move logic out of the template to a place where it's more readily testable.

shaunxp20:
> Just curious if there's something in particular you dislike about the `DisplayCertificates` guard module. It seems like a good idea to me to move logic out of the template to a place where it's more readily testable.

I dislike `DisplayCertificates` because it LIES!!! 😁 😂 From the current implementation... Tell me, `DisplayCertificates` can I display a certificate? Yes? Okay, then I'll display the certificate page. No? Okay, then I'll display the no_certificate page. What? I can't display that page? (Because perhaps `display_scores` is off) You told me there are no certificates to display! Just tell me yes or no! 🤣 🤒 😵 I'm a fan of: just keep the logic in one place, it's confusing enough! 😂

Also, it's misleading; okay `DisplayCertificates` tell me if I can display a certificate for this student. Oh, you can't do that?! But why not!?! Why do you exist? Except to make things more confusing!?! 😭 😦 

Anywho, a little over dramatic, but that's pretty much what I was going through when I was working on this! 😊 😁

shaunxp20:
Cool, thanks @stenington, I'll get this one merged in! 👍

PR #2495, opened 2020-04-24
#2501 Allow skipping certificate filling in non-production environments

stenington:
This is a proposal for allowing us to conditionally skip the actual filling of PDF certificates in non-production environments. On my machine, attempting to fill a pdf causes an exception, and various tests of ours attempt to fill a pdf. So without some way around it, I can't run tests locally any more.


shaunxp20:
I briefly looked at this and the other certificates PR (the diff to see anything popped to me). I think I found something, although it probably raises more questions! It looks like the old certificates have a space in the full name field: "FULL NAME" vs "FULL_NAME" in the new ones. I was experimenting w/ the student's certificate spec, and it seems to work with a space. It even seems to work **without** changing the certificates by replacing the underscore w/ a space, I did this: `Hash[ pdf.fields.map { |f| [f.name.gsub('_', ' '), get_value(recipient, f.name)] } ]` in `fill_pdfs.rb`, and the student's certificate spec seems to pass for me locally. 🤔 This was just a quick/rough experiment, so maybe it's something, maybe nothing? 🤷 But I wonder if you're seeing/experiencing the same thing w/ the spaces vs underscores. And I wonder if we should have Audra put the spaces back in, and try it that way? Anywho, random thoughts/findings. It's all so strange! 🤯 😖 😕

stenington:
@shaunxp20 I edited in the change you mention above, and the tests then passed but the certificates generated no longer contained the necessary data, they just contained the original filler, so I don't think that'll work.

How do you feel about the proposal here to be able to turn off proper certificate generation in dev environments? Were you looking for an alternate solution because you don't like it, or are you good with this PR?

shaunxp20:
> How do you feel about the proposal here to be able to turn off proper certificate generation in dev environments? Were you looking for an alternate solution because you don't like it, or are you good with this PR?

Yeah, it's just a bit of a bummer that these tests were passing like a week ago and now they're not locally. And we've been making all these great improvements to our test suite lately. Feels incongruous to what we've been doing, and little bit of a step backwards to me. 😕 

Anywho, I think it's fine to move this one forward if you want to get it off your plate. From the little bit I saw, the certificates code is pretty gnarly stuff! 😬 But I would like to take another looksie at this one to see if I can get them working locally (seems like that would be most ideal, imho) if I get some extra time during this iteration. 😀  🔦 🔨 👷

stenington:
I'm fine with you continuing to try to address this, let me know if you want a more detailed run down of what I've found out so far. My guess is that the path to addressing this has to do with changing something about the fonts embedded in the pdf. Notice in the stacktrace the method name **doType1TT**:

```
java.lang.ArrayIndexOutOfBoundsException: 0
   at pdftk.com.lowagie.text.pdf.DocumentFont.fillEncoding(pdftk)
   at pdftk.com.lowagie.text.pdf.DocumentFont.doType1TT(pdftk)
   at pdftk.com.lowagie.text.pdf.DocumentFont.<init>(pdftk)
   at pdftk.com.lowagie.text.pdf.AcroFields.getAppearance(pdftk)
   at pdftk.com.lowagie.text.pdf.AcroFields.setField(pdftk)
   at pdftk.com.lowagie.text.pdf.AcroFields.setFields(pdftk)
```

That hints to me that it's trying to handle some type of TrueType font, and getting something wrong. Given that [this StackOverflow comment](https://stackoverflow.com/questions/23948647/font-issue-with-pdftk#comment36894874_23948647) from the creator in 2014 says the software was already 5 years out of date, it just doesn't surprise me that the tool wouldn't work perfectly any more, and so it seems practical to provide an escape hatch for the situation we're in.

shaunxp20:
> I'm fine with you continuing to try to address this, let me know if you want a more detailed run down of what I've found out so far. My guess is that the path to addressing this has to do with changing something about the fonts embedded in the pdf. Notice in the stacktrace the method name **doType1TT**:
> 
> ```
> java.lang.ArrayIndexOutOfBoundsException: 0
>    at pdftk.com.lowagie.text.pdf.DocumentFont.fillEncoding(pdftk)
>    at pdftk.com.lowagie.text.pdf.DocumentFont.doType1TT(pdftk)
>    at pdftk.com.lowagie.text.pdf.DocumentFont.<init>(pdftk)
>    at pdftk.com.lowagie.text.pdf.AcroFields.getAppearance(pdftk)
>    at pdftk.com.lowagie.text.pdf.AcroFields.setField(pdftk)
>    at pdftk.com.lowagie.text.pdf.AcroFields.setFields(pdftk)
> ```
> 
> That hints to me that it's trying to handle some type of TrueType font, and getting something wrong. Given that [this StackOverflow comment](https://stackoverflow.com/questions/23948647/font-issue-with-pdftk#comment36894874_23948647) from the creator in 2014 says the software was already 5 years out of date, it just doesn't surprise me that the tool wouldn't work perfectly any more, and so it seems practical to provide an escape hatch for the situation we're in.

Thanks for the background info! I had a chance to look at this, TLDR; let's get this one merged in! 😁 

-----

Brain dump of random, for funsies things I found (in case we need to revisit this):

The issue definitely seems to be with pdftk. The real fix seems to be to use a different library. There's another wrapper for pdftk for PHP w/ the exact same issue:
https://github.com/mikehaertl/php-pdftk/issues/31

The fix in the Stack Overflow post (which I tried but it didn't work) apparently has some other side effects:
https://github.com/mikehaertl/php-pdftk/issues/31#issuecomment-456705225

The linked issue from that thread is interesting:
https://github.com/mikehaertl/php-pdftk/issues/4

pdftk uses iText to fill in the PDFs. (Bruno Lowagie from the Stack Overflow post wrote/is involved w/ iText.) iText is still active/current, and there seems to be an actively supported gem to use iText in Ruby:

https://github.com/vkononov/fillable-pdf
`FillablePDF is an extremely simple and lightweight utility that bridges iText and Ruby in order to fill out fillable PDF forms or extract field values from previously filled out PDF forms. `

The version of pdftk we're using is written in C++, this one is the 6+ years since it's been updated version. I found a Java port of pdftk that seems to be actively maintained:

https://gitlab.com/pdftk-java/pdftk
`This is a port of pdftk
into Java. The current goal is to make a translation as faithful as it
is reasonable and to fix any issues present in the original
(correctness takes precedence over compatibility, see the differences),
leaving possible improvements and refactoring for
later. So far all code has been manually translated and it passes the
test suite of php-pdftk,
but a lot more testing is needed. Due to the differences between C++
and Java, it is likely that a few bugs have sneaked in with respect to
the original; any help in catching them will be appreciated.`

And here's an open source version of iText:

https://github.com/LibrePDF/OpenPDF
`OpenPDF is a free Java library for creating and editing PDF files with a LGPL and MPL open source license. OpenPDF is based on a fork of iText. We welcome contributions from other developers. Please feel free to submit pull-requests and bugreports to this GitHub repository.`

Issue #2508, opened 2020-05-06
Set up an auto-scaler [20,13,1.54]


stenington:
From #2505, let's start by configuring the Rails Autoscale add-on and seeing how it performs (monitoring throughout the iteration) with default settings (or simple tweaks if needed). This will start to give us a baseline for future modifications we might want to make.

Let's also be conscious of the cost: the add on costs some money per month, but ideally it's scaling down our dynos when the site is slow and netting us some savings.

shaunxp20:
I added a note and some screenshots in [Slack](https://iridescentlearning.slack.com/archives/C06CHBQ4E/p1589922066003700) about the cost savings. 💸  We can revisit the this to get better data/figures after Rails Autoscale has been running for an entire/complete month (at/after the end of June). 😃

PR #2511, opened 2020-05-07
#2508 Install rails_autoscale_agent gem

shaunxp20:
#2508

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -132,6 +132,7 @@ group :production do
   gem 'hiredis', "~> 0.6"
   gem "rack-timeout", "~> 0.5"
   gem 'heroku-deflater', "~> 0.6"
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -132,6 +132,7 @@ group :production do
   gem 'hiredis', "~> 0.6"
   gem "rack-timeout", "~> 0.5"
   gem 'heroku-deflater', "~> 0.6"
+  gem 'rails_autoscale_agent'
```
shaunxp20, line 135: It's not obvious from the diff, but this is in the `:production` group. Let me know if you want it in an other/all environments. 😸

stenington: Production only makes sense to me

Issue #2463, opened 2020-04-07
Confirming/improving the task to drop lowest scores [13,13,1]


stenington:
We have a rake task that drops lowest scores for semifinalists, skipping a hardcoded list of ids, but we should confirm with the team when they plan to drop lowest scores and how they want that to work. Will it apply across the board? Do they anticipate needing to omit certain teams (e.g. teams with fewer scores maybe)? Do they anticipate doing this for semifinals only, or other rounds? Basically figure out the parameters they might want to control and update the tool to be ready for that.

### QA

For both QFs and SFS, as both a dry run then a real run, we should:

1. identify two different submissions with different numbers of scores for that round
2. run the task with the threshold set for the submission with more scores, and confirm only that submission has a score dropped
3. run the task again with a lower threshold, and confirm that the submission with fewer scores gets one dropped, and the submission with higher scores does not have a second score dropped
4. confirm that the dropped scores can show as dropped throughout the admin

rsgonzal:
- Last year we had a hard time getting judges to judge semis, barely got 2 scores per team. 
- This year, we should have the option to run it - for BOTH quarters and semis - depending on how many scores we get
- Fairest thing is prob drop the lowest score for all teams uniformly
- But the call will be made day before or right after judging round closes
- For teams with X # of scores, drop the lowest ones

hellafitz:
- update the tool to be more flexible
- update the documentation

stenington:
It occurred to me while starting on this that admins can now see when scores have been deleted, but will have no indicator of which scores were purposefully dropped as the lowest score. That seems like it would be important info for them, right? @rsgonzal @hellafitz

stenington:
Here's my proposal for showing dropped scores along with deleted scores:

In **Score Exports** we add a filter for dropped scores, and a yes/no column for "Dropped" alongside the existing "Deleted" column. This should help us confirm that we're dropping the right scores when running the tool, I think:

<img width="1280" alt="Screen Shot 2020-05-07 at 3 27 24 PM" src="https://user-images.githubusercontent.com/55319/81336537-85e35e80-9077-11ea-9fc5-fc331f6d74ad.png">

When viewing the **Score details** for a submission, dropped scores will say "dropped":

<img width="1280" alt="Screen Shot 2020-05-07 at 3 28 44 PM" src="https://user-images.githubusercontent.com/55319/81336534-85e35e80-9077-11ea-8413-167f834aa705.png">

[...]

PR #2512, opened 2020-05-08
#2463 improve rake task for dropping low scores

stenington:
Some notes on this one:

Scores could already be marked as "dropped" prior to this PR with a timestamp called `dropped_at`. I added an `after_restore` hook to unset that field if a deleted score gets restored, because a score should only be considered dropped if it is also deleted, in my mental model of how this works.

Prior to this PR, submission also had a timestamp field for `lowest_score_dropped_at` that would prevent dropping multiple scores from a submission with multiple runs of the tool. That's no longer sufficient with the requirement that we be able to drop low scores for both quarterfinals and semifinals, so that field is gone.

The new `LowScoreDropping` class is an alternative to putting all those methods on `TeamSubmission`, and sort of acts as a facade over a submission with just the methods useful to the context of dropping the lowest score. The "Xing" naming convention is something we have elsewhere in `app/technovation/`, and to me it just signifies this sort of class that helps orchestrate some action on a model or collection of models. 

The `drop_lowest_score` rake task got rewritten to make use of the `LowScoreDropping` class, and to make the tool more powerful. It can do quarterfinals, semifinals, different thresholds for a minimum score count, and has a dry run mode that outputs what will happen without doing it.

If this PR gets accepted, I'll also update documentation ([probably here?](https://docs.google.com/document/d/1_rVW24C1-RPGF4V3q3I1kr5Cu3CVqg2PoPt8Eja6s5c/edit#heading=h.arur096xgw8u)) to cover some of the new capabilities of the rake task.

A note on the previously hardcoded exceptions: I'm assuming the exceptions were made for submissions without as many scores. I haven't yet heard another reason we might want to avoid dropping a team's lowest score, but if we really need to admins now have the ability to restore scores so we could always tweak things that way.


app/technovation/low_score_dropping.rb
```diff
@@ -0,0 +1,34 @@
+class LowScoreDropping
+  def initialize(submission, options={})
+    @submission = submission
+    @round = options.fetch(:round)
+    @score_scope = @submission.public_send("#{@round}_complete_submission_scores")
+    @minimum_score_threshold = options.fetch(:minimum_score_threshold)
```
shaunxp20, line 6: Looks like a good candidate to use some keyword arguments here, if you're feeling up for it. Also, fwiw, `@score_scope` looks like it can be a private method, it doesn't look like it's used outside of this class. 🤔 😄 😀

app/technovation/low_score_dropping.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  def able_to_drop?
+    has_enough_scores? && !already_dropped?
+  end
+
+  def already_dropped?
+    @score_scope.only_deleted.any? {
```
shaunxp20, line 14: I'm wondering if we can use the new `dropped` score here? 🤔 Also, fwiw, I think you can make it like `lowest_score` and do something like `...any?(&:dropped?)`. 🤔 😺

app/technovation/low_score_dropping.rb
```diff
@@ -0,0 +1,34 @@
...
+      |score| score.dropped?
+    }
+  end
+
+  def has_enough_scores?
+    @score_scope.count >= @minimum_score_threshold
```
shaunxp20, line 20: Not sure where to comment on this, but it looks like each time this (and other things) gets called we'll be making additional/redundant database queries? Anywho, it might be a good candidate for some memoizing, but also it doesn't seem like it will be the end of the world since this is only going to be run a few times. 😄 🤓 :neckbeard: 😀

app/views/regional_ambassador/score_details/_scores_table.en.html.erb
```diff
@@ -11,7 +11,17 @@
           <td class="official-info"><%= score.official? ? "official" : "unofficial" %></td>
 
           <% if current_account.admin? %>
-            <td class="deleted-info"><%= score.deleted? ? "deleted" : "" %></td>
+            <td class="deleted-info">
+              <%=
+                if score.deleted?
+                  if score.dropped?
+                    "dropped"
```
shaunxp20, line 18: 👍

db/migrate/20200508185107_remove_lowest_score_dropped_at_from_team_submission.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveLowestScoreDroppedAtFromTeamSubmission < ActiveRecord::Migration[5.1]
+  def change
+    remove_column :team_submissions, :lowest_score_dropped_at, :datetime
+  end
+end
```
shaunxp20, line 5: Btw, I see the changes locally in `structure.sql` for this change, but it looks like they're missing from the diff. 🤔

lib/tasks/drop_lowest_scores.rake
```diff
@@ -1,21 +1,57 @@
...
-  TeamSubmission.where
-    .not(team_id: EXCLUDE_TEAM_IDS)
-    .current
-    .semifinalist
-    .find_each do |submission|
-    DropLowestScores.(submission, STDOUT)
+VALID_QF_PHRASING = ["qf", "qfs", "quarterfinals", ":qf", ":qfs", ":quarterfinals"]
+VALID_SF_PHRASING = ["sf", "sfs", "semifinals", ":sf", ":sfs", ":semifinals"]
+
+desc "Drop the lowest current scores on submissions for a specified round and minimum score threshold, pass 'run' as the dry_run argument to turn dry run mode OFF."
```
shaunxp20, line 4: Quibble at the lingo - "minimum score threshold" here, that equates to `minimum_score_count`, and then gets passed to `minimum_score_threshold`. Also, "minimum score threshold" makes me think, drop a score that's less than a certain amount, but maybe that's just me. 🤷 🤷‍♂️ 😀 Anywho, not really a blocker.

lib/tasks/drop_lowest_scores.rake
```diff
@@ -1,21 +1,57 @@
...
+    rank = :quarterfinalist
+    round = :quarterfinals
+  elsif VALID_SF_PHRASING.include?(round.downcase)
+    rank = :semifinalist
+    round = :semifinals
+  else
+    raise "Unknown round: #{round}"
+  end
+
+  puts "DRY RUN: #{dry_run ? 'on' : 'off'}"
```
shaunxp20, line 24: I like the dry run functionality! 👍

spec/views/admin/regional_ambassador/score_details/scores_table_spec.rb
```diff
@@ -88,6 +94,23 @@
...
+        end
+      end
+
+      it "displays the row in a subtle red color" do
+        within "tr" do
+          expect(rendered).to have_css(".background-color--subtle-red")
+        end
+      end
+    end
+
```
shaunxp20, line 113: So much coverage! 💯

shaunxp20:
How do I run this rake task locally?! I'm getting the syntax wrong. #fail 😦 😢

app/technovation/low_score_dropping.rb
stenington: Yeah, I decided not to worry about optimization at first, especially since some things like `already_dropped?` **should** run the query each time to prevent dropping multiple scores.

db/migrate/20200508185107_remove_lowest_score_dropped_at_from_team_submission.rb
stenington: I don't understand how `structure.sql` gets used or updated. Is there something I have to do after creating a migration to keep things in sync?

lib/tasks/drop_lowest_scores.rake
stenington: Ah, yeah "count" does seem better than "threshold" when I take a step back from it.

app/technovation/low_score_dropping.rb
stenington: Good suggestions!

app/technovation/low_score_dropping.rb
stenington: Yep.

stenington:
To run, `bundle exec rails drop_lowest_scores[qf,3]` or `bundle exec rails drop_lowest_scores[qf,3,run]` for example.

stenington:
Not sure what the hell just happened, I was not trying to merge this branch. I was trying to update it. Reverting.

PR #2514, opened 2020-05-12
#2463 try again

stenington:
Okay, trying to get #2512 reopened here for a final review. Sorry to have to do it as a new PR, this is all a bit of a headache.


stenington:
@shaunxp20 I think I addressed all of your comments in #2512, can you give this one more look over?

app/technovation/low_score_dropping.rb
```diff
@@ -0,0 +1,36 @@
+class LowScoreDropping
```
shaunxp20, line 1: Nice changes here! 🙌 

Some peanut gallery commentary after looking at this again, I could see moving the dry run functionality here, which I think would simplify things in that only one method would need to be public, the rest could be private, and the rake task would have to do less work, and know less. Anywho, just #random-thoughts! 😃 🤓

stenington: But I want the rake task to output a bunch of info, and I **didn't** want this class to only be useful in the rake task. For example, if we hooked this up to a button in the admin, we could still use it and it wouldn't be filling the logs with useless info for that context. If I added the dry run here and made everything else private, this class would have to do the reporting that's needed in one context but not the other.

I don't know if I got it totally right, but that was one of the principles I was trying to use when drawing lines of responsibility.

I suppose I could have implemented that as logging, and had the rake task turn logging up somehow before calling this class to do its thing? Then in a web context, lower log levels would mean that extra info stayed silent? That might be an interesting adjustment to try out.

shaunxp20: Ah yeah, I wasn't thinking this class would log/puts output, I was thinking it could just return a result object for each submission, the result would include a message (plus other things) that the rake task (or admin) could use. 😄 😁

Issue #2440, opened 2020-03-24
Bug: deleting a judge with scores breaks the score exports admin screen [8,13,0.62]


stenington:
**Describe the bug**
As an admin, if you delete a judge who had scores by visiting their participant page and clicking the delete button, the score exports page will no longer work if you choose the "Include deleted scores" or "Only show deleted scores" filter options.

**To Reproduce**
1. Go to **Scores export** as an admin
2. Select "Only show deleted scores" and click **Search**
3. Note the **Judge name** of the first judge
4. Go to **Participants** and look up that judge
5. Click **view**
6. Scroll to the bottom and click **Delete <judge> from Technovation**
7. Return to the **Scores export** page
8. Select "Only show deleted scores" and click **Search**
9. You should get an error

**Expected behavior**
Deleted scores associated with a deleted judge should not appear in the Scores export list.

<img width="1280" alt="Screen Shot 2020-03-24 at 2 21 53 PM" src="https://user-images.githubusercontent.com/55319/77462696-d9a52b80-6dda-11ea-96c8-26a709a33260.png">

### QA

Once this is live, you should be able to follow the steps to reproduce without getting an error at the end. Instead that judge's scores should no longer appear on the **Scores export** page.

hellafitz:
Confirm that:
- deleting Judges is a capability we want to allow
- There are deleted scores from deleted Judges - and that causes a problem. Do we need to see those? Or do we only need to see deleted scores from non-deleted Judges?

rsgonzal:
Do not plan to delete judges during current season, will only suspend

Always a possibility that a judge will email us to delete their account, so this issue could arise.

stenington:
Yeah, so my take away is that we do have to handle this edge case, but we would not expect to see deleted scores from a deleted judge. And actually, if the user is requesting to be deleted that should be a hard delete anyway, I think?

stenington:
We're also uncertain what should happen if a judge who had scored submissions then asks to be deleted. It would be nice to preserve their scores and just sub in some sort of default judge info, but are the comments they left considered their personal data that should be deleted if they want to be removed?

stenington:
We can mitigate the biggest risk here and set aside some bigger picture conversation for later. In this ticket, let's make it so that deleting a judge who has scores (by clicking the delete button at the bottom of their participants page) doesn't break the score exports page. Deleted scores from active judges are okay, deleted scores from deleted judges should no longer be shown, period.

[...]

PR #2516, opened 2020-05-12
#2440 filter scores attached to deleted judges from scores datagrid

stenington:
This one wasn't as bad as I thought it might be. Initially I was just adding a check for `judge_profiles.deleted_at IS NULL` but that wasn't guaranteeing the presence of a judge profile because the sql query is using a LEFT OUTER JOIN, so in some cases **all** the judge profile fields were null. Asserting that we do have an id and we don't have the deleted flag works though.


app/models/submission_score.rb
```diff
@@ -182,6 +182,12 @@ class SubmissionScore < ActiveRecord::Base
     end
   }
 
+  scope :judge_not_deleted, -> {
+    includes(:judge_profile)
+      .references(:judge_profiles)
+      .where("judge_profiles.id IS NOT NULL AND judge_profiles.deleted_at IS NULL")
```
shaunxp20, line 188: Huh, that `judge_profiles.id IS NOT NULL` part is strange! 🤔 😕 Fwiw, I tried a left join with just `judge_profiles.deleted_at IS NULL` in a manual/non-Rails SQL query locally and it works as expected. I'm going to guess that `judge_profiles.id IS NOT NULL` is necessary b/c of the paranoia soft-delete gem is interfering w/ the query. 🤔 ❔  🤷 Anywho, looks good to me, and it works! And tests! 👍 💯 :shipit:

stenington: My understanding is that references + includes does a LEFT OUTER JOIN, which is why you have to affirm you have an id.

shaunxp20: Looks like Paranoia is the culprit, it's adding a clause to the join:

`LEFT OUTER JOIN "judge_profiles" ON "judge_profiles"."id" = "submission_scores"."judge_profile_id" AND "judge_profiles"."deleted_at" IS NULL`

When I was testing locally I just did a normal/non-paranoia join: `left join judge_profiles on submission_scores.judge_profile_id = judge_profiles.id`, (with `judge_profiles.deleted_at IS NULL` in the where clause too) and it works as expected w/o the `judge_profiles.id IS NOT NULL`. Anywho, I think it's fine. 👍

stenington: Ah, ok, makes sense. Thanks!

Issue #2406, opened 2020-02-28
Edit error messages for email address changes [5,5,1]


rsgonzal:
After #2347 -The error messages that pop up when entering personal and parent emails are a little wonky.

The multiple messages when changing the personal email are aggressive:
![Screen Shot 2020-02-28 at 2 05 02 PM](https://user-images.githubusercontent.com/22801026/75594152-a28e6500-5a3c-11ea-8cd9-4644373a7b84.png)
Trim the multiple identical error messages down to one.

The first screen that prompts you to change the parent email talks about another student account:
![Screen Shot 2020-02-28 at 2 48 44 PM](https://user-images.githubusercontent.com/22801026/75594178-b934bc00-5a3c-11ea-8b97-b5a5fada5f3c.png)
Instead it should say "cannot match your (or any other student's) email"

## QA STEPS
 * Login or impersonate a student

#### Part 1
 * Navigate to My Profile -> Change email
 * Update the email address without providing a password
 * Submit the form
 * Verify you only see one error message (not three!)

#### Part 2
 * Navigate to Parent Consent, and click Edit email address
 * Change the email address to the student's email address, or another student's email address
 * Verify the new error message reads: "cannot match your (or any other student's) email"

stenington:
I don't have my head totally wrapped around this, but here's what I've found so far. These profile editing forms are a little complicated in that they combine fields on both the profile model and the Account model. 

When you're editing your profile (My Profile > Change your basic profile details), errors are displayed once, below the input field. 
<img width="1280" alt="Screen Shot 2020-03-20 at 5 00 26 PM" src="https://user-images.githubusercontent.com/55319/77205942-5202a880-6acc-11ea-8cac-fa9698de7438.png">

When you edit your email, errors are displayed on the label, on the field, and again below the field as shown in the original ticket. Note that it's not specific to the parental email error though. Inputting the wrong password will also give you triple error messages about that.
<img width="1280" alt="Screen Shot 2020-03-20 at 5 02 37 PM" src="https://user-images.githubusercontent.com/55319/77206120-a017ac00-6acc-11ea-9d84-b91979a9367c.png">

The main difference I see behind these two pages is that the profile page uses `simple_form_for` whereas the email page uses `form_with`. Switching the email page to use `simple_form_for` and then e.g. `a.input :email` instead of separate label, field, and errors does seem to bring the validations back in line with other pages. 

[...]

PR #2519, opened 2020-05-13
#2406 Update email address error messages

shaunxp20:
This will convert the "Change your email" form to use Simple Form, which will ultimately eliminate the triple error messages! :grin:

Just FYI: this was part of the culprit (responsible for two of the three error messages):
`config/initializers/action_view.rb`.

#2406

### DIFF
--- a/app/views/profiles/_email.en.html.erb
+++ b/app/views/profiles/_email.en.html.erb
@@ -19,27 +19,14 @@
   to confirm your new email address and restore full access to this site.
 </p>
 
-<%= form_with model: current_profile, local: true do |f| %>
-  <%= f.fields_for :account, include_id: false do |a| %>
-    <%= a.label :existing_password, "Current password" %>
-    <%= a.password_field :existing_password,
+<%= simple_form_for current_profile, local: true do |f| %>
+  <%= f.simple_fields_for :account, include_id: false do |a| %>
+    <%= a.input :existing_password,
+      label: "Current password",
       id: "existing_password_field_email",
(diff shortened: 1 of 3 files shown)


app/views/profiles/_email.en.html.erb
```diff
@@ -19,27 +19,14 @@
...
-    <% end %>
-
-    <%= a.label :email %>
-    <%= a.email_field :email %>
-
-    <% unless a.object.errors[:email].empty? %>
-      <div class="field_with_errors">
-        <span class="error"><%= a.object.errors[:email].to_sentence %></span>
-      </div>
-    <% end %>
```
shaunxp20, line 42: Btw, I removed this custom error stuff. It doesn't look like we need it (unless I'm missing something), I tested it and Simple Form seems to output the errors just fine. 🤷 😄 😊 😀

config/locales/activerecord.en.yml
```diff
@@ -40,7 +40,7 @@ en:
         student_profile:
           attributes:
             parent_guardian_email:
-              found_in_student_accounts: "cannot match another student's email"
+              found_in_student_accounts: "cannot match your (or any other student's) email"
```
shaunxp20, line 43: @stenington I was about to write up the QA steps and realized I missed this requirement, doh! 🤦  You already approved this PR, not sure if you wanted to give it another looksie? 🤔 😄 😀

stenington: Yep, seems right to me.

Issue #2489, opened 2020-04-23
Write integration test for admin setting team contest rank through scores page [?,3,?]


stenington:
<img width="960" alt="Screen Shot 2020-04-23 at 5 38 03 PM" src="https://user-images.githubusercontent.com/55319/80152232-4f8ee500-8589-11ea-837c-b162b16b768f.png">
Admins have a control on the Scores page to set the **Contest rank** of a team directly. Write tests covering this functionality.

PR #2522, opened 2020-05-14
#2489 Add tests for admin setting contest rank

stenington:
Some more tests

### DIFF
--- a/spec/features/admin/setting_contest_rank_spec.rb
+++ b/spec/features/admin/setting_contest_rank_spec.rb
@@ -8,7 +8,7 @@ RSpec.feature "Admin updates contest rank", js: true do
     when_they_are_marked_as_a_semifinalist
 
     then_a_confirmation_shows
-    and_the_change_is_persisted
+    and_the_new_rank_is_remembered
   end
 
   scenario "setting a finalist" do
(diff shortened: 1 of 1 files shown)


spec/features/admin/setting_contest_rank_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+
+RSpec.feature "Admin updates contest rank", js: true do
+  scenario "setting a semifinalist" do
+    given_there_is_a_scored_quarterfinalist_team
+    and_a_logged_in_admin_viewing_scores
+
+    when_they_are_marked_as_a_semifinalist
+
+    then_a_confirmation_shows
+    and_the_change_is_persisted
```
shaunxp20, line 11: Minor quibble with the "persisted" part, seems more like a back-end/unit test term than a front-end/UI/feature spec term. Anywho, just a random quibble, not a blocker, going to approve it! :shipit: 😺 😁

stenington: Yeah, it felt too low level but I do want to check that the change is persisted. Is refreshing the page and confirming the rank through the UI better?

shaunxp20: I think that makes sense. If we're treating our feature specs as let's use the UI to drive out the behavior/expectations. 👍 And maybe even `and_the_semifinalist_rank_is_reflected` or shown or something?  🤔 😸 #random-idea

Issue #2520, opened 2020-05-14
Assign a judge to a specific submission [40,13,3.08]


hellafitz:
As an alternative to manually inputting scores, we want to allow admins to assign a specific judge to a specific submission (in need of judging). Currently submissions are assigned to judges only automatically. This should be done through the Score Details screen, accessible from the Scores page in the dashboard. 

The earlier this is available to CET, the more valuable it is (because we can engage engaged Judges).

Goal: significantly reduce "pen & paper" scores 

Q&A: 
- do you do this by looking up the judge, or looking up the submission? A: submission route, it'd be nice to add multiple judges to the submission at the same time (resulting in 2-3 future scores)
- would these judges already have accounts, or be required to? A: Yes
- how would an admin identify a specific judge (e.g. email address, ID, etc)? A: Email address field to paste in. Bryant has an active Judges spreadsheet he'd be referencing. 
- would any of our current limitations on judges being assigned particular types of submissions need to be voided in this use case? A: should be able overrule restrictions. nice ot have a 'confirm' notification about that fact that you're doing it. 
- what's the minimum level of alerting the Judge that we'd do? (we'd create a score in progress, then the 'resume my score' would take them to the new submission; CET would need to email them or we'd need to provide some other type of alert) A: CET would email. 

## QA STEPS

The new judge assignment controls are only available when judging is enabled; the new controls can be found on the score details page. To get there, as an admin, click Scores, and then in the scores/search results table, click the bulleted-list icon (on the very right-hand side of the table; that you may need to scroll-right to see).

Verify the new judge assignment controls for:
 * a valid judge email address; also verify the judge gets assigned the submission
 * a judge who already has a score in progress (this can be done by re-assigning a judge the same submission twice)
 * a suspended judge email address
 * a non-judge email address
 * a made-up email address

We should also verify an RA doesn't see the judge assignment controls.

stenington:
A technical note: judges can only have one score in progress, so we would not be able to give multiple submissions to a single judge in a useful way. We could theoretically create a bunch of complete scores and ask them to edit them, but on the off chance they don't, I wouldn't want to be polluting our score pool with bad completed scores. Setting them up with a single in-progress score seems to be the way to go.

The screen where you could add a judge onto a submission by email address would need to account for that then, and refuse to make the assignment if the judge has a score in progress already.

hellafitz:
- next step: mockups OR mockups + MVP implementation

hellafitz:
[...]

PR #2528, opened 2020-05-23
#2520 Assign a judge to a submission

shaunxp20:
As per #2520, this will allow a submission to be assigned to a judge. 🧑‍⚖️ 📝

### DIFF
--- a/app/technovation/submission_to_judge_assignor.rb
+++ b/app/technovation/submission_to_judge_assignor.rb
@@ -1,0 +1,37 @@
+class SubmissionToJudgeAssignor
+
+  def initialize(submission:,
+                 judge:,
+                 validator: SubmissionToJudgeValidator,
+                 score_questions: Questions)
+
+    @submission      = submission
+    @judge           = judge
+    @validator       = validator
+    @score_questions = score_questions
+  end
+
+  def call
+    if validator_result.success?
+      assign_submission_to_judge
+
+      Result.new(success?: true, message: { success: "This submission was successfully assigned to #{judge.account.email}" })
+    else
+      validator_result
+    end
+  end
+
+  private
+
+  Result = Struct.new(:success?, :message, keyword_init: true)
+
+  attr_reader :submission, :judge, :validator, :score_questions
+
+  def validator_result
+    @validator_result ||= validator.new(judge: judge).call
+  end
+
+  def assign_submission_to_judge
+    score_questions.new(judge, submission)
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -261,7 +261,9 @@ Rails.application.routes.draw do
     resources :profile_locations, only: :edit
 
     resources :teams, except: :destroy
-    resources :team_submissions, except: :destroy
+    resources :team_submissions, except: :destroy do
+      resource :judge_assignments, only: :create
+    end
     resources :team_memberships, only: [:destroy, :create]
     resources :team_locations, only: :edit
     resources :team_submission_certificates, only: :update
--- a/spec/technovation/submission_to_judge_validator_spec.rb
+++ b/spec/technovation/submission_to_judge_validator_spec.rb
@@ -1,0 +1,63 @@
+require "rails_helper"
+
+RSpec.describe SubmissionToJudgeValidator do
+  let(:validator_result) { SubmissionToJudgeValidator.new(judge: judge).call }
+  let(:judge) { instance_double(JudgeProfile, account: account, suspended?: judge_suspended) }
+  let(:account) { instance_double(Account, email: "judge345@alwaysbejudging.com") }
+  let(:judge_suspended) { true }
+
+  context "when a judge can be assigned another submission (not suspended and doesn't have another score in progress)" do
+    let(:judge_suspended) { false }
+
+    before do
+      allow(judge).to receive_message_chain("scores.current_round.incomplete.any?").and_return(false)
+    end
+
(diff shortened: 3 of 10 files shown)


config/routes.rb
```diff
@@ -261,7 +261,9 @@
     resources :profile_locations, only: :edit
 
     resources :teams, except: :destroy
-    resources :team_submissions, except: :destroy
+    resources :team_submissions, except: :destroy do
+      post :judge_assignment
```
stenington, line 265: I have a question about team norms here. I feel like the current code base mostly tries to avoid adding extra actions to a resource like this, but I don't know if that's some sort of best practice we should adhere to. Do you have any insight on that? 

The contrasting pattern I see is to "nounify" an interaction, give it its own controller, and reuse the standard REST actions, like the way judge assignments or event assignments are handled for RAs. I kind of like how it keeps controllers small and focused, but it can also feel like extra hoops to jump through when a custom action would do just fine.

Is this something you have any guidance on from experience @shaunxp20?

app/technovation/submission_to_judge_assignor.rb
```diff
@@ -0,0 +1,37 @@
...
+                 validator: SubmissionToJudgeValidator,
+                 score_questions: Questions)
+
+    @submission      = submission
+    @judge           = judge
+    @validator       = validator
+    @score_questions = score_questions
+  end
+
+  def call
```
stenington, line 14: It seems like there's a tight coupling here between `SubmissionToJudgeAssignor` and `SubmissionToJudgeValidator` since `call` might return a `Result` from either one. I'm on the fence as to whether I would suggest just folding `validate_judge` in here somehow. I think it's your call. Having them split does seem like it helps keep them tightly focused.

app/technovation/submission_to_judge_assignor.rb
```diff
@@ -0,0 +1,37 @@
...
+  Result = Struct.new(:success?, :message, keyword_init: true)
+
+  attr_reader :submission, :judge, :validator, :score_questions
+
+  def validator_result
+    @validator_result ||= validator.new(judge: judge).call
+  end
+
+  def assign_submission_to_judge
+    score_questions.new(judge, submission)
```
stenington, line 35: It bugs me that initializing a `Questions` object seems to be the way to assign a submission to a judge. (Just commenting on existing weirdness in the code base, I think you're doing the right thing in this PR given the weirdness already exists.)

spec/technovation/submission_to_judge_validator_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+  let(:validator_result) { SubmissionToJudgeValidator.new(judge: judge).call }
+  let(:judge) { instance_double(JudgeProfile, account: account, suspended?: judge_suspended) }
+  let(:account) { instance_double(Account, email: "judge345@alwaysbejudging.com") }
+  let(:judge_suspended) { true }
+
+  context "when a judge can be assigned another submission (not suspended and doesn't have another score in progress)" do
+    let(:judge_suspended) { false }
+
+    before do
+      allow(judge).to receive_message_chain("scores.current_round.incomplete.any?").and_return(false)
```
stenington, line 13: Interesting technique here, I was not aware of `receive_message_chain`. Seems like a nice tool for mocking out things that would otherwise rely on ActiveRecord.

app/technovation/submission_to_judge_assignor.rb
shaunxp20: My thinking w/ separating the judge validation functionality is that it doesn't have any side effects, versus the assignment functionality which has side effects, which yeah, makes the validator more focused/independent/easier to test/etc. 😄 😁  And just FYI, that `Result` response was just a hodgepodge - the `success?` method came from the [Braintree API](https://developers.braintreepayments.com/reference/general/result-objects/ruby), and the `message` format is just so it can be used as a flash message. 👺 🌈 🐤 🎩 😺  Maybe it can be extracted out later if we like it, I saw quite a few approaches to the response object approach too. 🤯 😀

config/routes.rb
shaunxp20: Ah yeah, I remember having a similar convo when we did the score restore, and judge suspend/unsuspend stuff. My approach here was to just to something similar (and to just get something working!). I'm definitely open to a different/standardized approach. I don't have too much of an opinion, except that I agree, if we keep adding to existing controllers it's going to make the controllers more messy, and that separate controllers would be more focused. 👍

I'm up for changing/standardizing this, I guess the question is what should I change it to? 🤔 I think making `judge_assignment` a resource would align it w/ our other assignment controllers if we like that approach? Or something else? 🤔 I'm open for either! 😁

app/technovation/submission_to_judge_assignor.rb
shaunxp20: A yeah, that `Questions` code was weird/a head scratcher for me too! 🤪  🧟 🤕

config/routes.rb
stenington: Yeah, it might be slightly misleading to have RA judge assignments and admin judge assignments use the same phrase to mean slightly different things, but I can't think of better terminology. I would probably go with a `judge_assignment` resource under the admin namespace.

shaunxp20: I added a new `judge_assignments` controller for this new functionality! 😃

stenington: Cool, I like it! 🎉

Issue #2517, opened 2020-05-12
Allow teams continued access to their submissions [13,8,1.63]


hellafitz:
Ophelie: "teams should be able to see their submission either after submission closes or after judging is done until next season. Otherwise, submission closes and it’s gone fooooreeeveeer (for teams at least, seems mentor can still see them)" 
https://docs.google.com/document/d/18a9ToY63JnyyksC7X13oAkdJ7okuOu_PugYMxKBkKsw/

## QA STEPS

* As an admin, in "Content & Settings", turn off this setting: `Team submissions are editable` 
* Find a student who is on a team, and the team's submission is complete; login as this student
* Verify there is a button on the student's dashboard that will display the team's completed submission; this should have the same functionality as the "My team's submission" link in the header navbar
* You can also verify this button doesn't show up when editing submissions is turn on (and other combinations of conditions on bullet point #2 (student not a team, submission incomplete))

Also, I updated the "My team's submission" link; the functionality should be the same, but that's also something that can be tested if time permits.

hellafitz:
potential solution: provide access to the public view of submissions after submissions close (instead of just giving them messaging about how everything is closed)
- gather screenshots of current state
- rough mockup of proposal

stenington:
Sorry for the broken images here, but this gives you an idea of the current flow. After submissions end, students can still see their submissions if they know to click the "My team's submission" navigation tab, but the main dashboard becomes a bit of a dead end.

## While team submissions are editable

Student dashboard
<img width="1213" alt="Screen Shot 2020-05-19 at 10 40 42 AM" src="https://user-images.githubusercontent.com/55319/82340658-7b04c400-99bd-11ea-80a2-aa8fd5174e7a.png">

My Team's Submission tab
<img width="1213" alt="Screen Shot 2020-05-19 at 10 40 48 AM" src="https://user-images.githubusercontent.com/55319/82340654-7a6c2d80-99bd-11ea-8340-e53890e1617a.png">

"View your complete submission" button
<img width="1213" alt="Screen Shot 2020-05-19 at 10 40 56 AM" src="https://user-images.githubusercontent.com/55319/82340653-7a6c2d80-99bd-11ea-80d2-82ccc36c4db0.png">

## While team submissions are not editable (i.e. judging onwards)

Dashboard (becomes a dead end)
[...]

PR #2532, opened 2020-05-27
#2517 Add link to student dashboard to view submission

shaunxp20:
This will add a new link to the student's dashboard to view their completed submission, when editing submissions is turned off.

Refs: #2517

### DIFF
--- a/app/views/student/dashboards/_view_submission_link.html.erb
+++ b/app/views/student/dashboards/_view_submission_link.html.erb
@@ -1,0 +1,9 @@
+<% if StudentSubmissionLinkGuard.new(team: current_team, student: current_student).
+  display_link_to_published? %>
+
+  <div class="margin--t-xlarge">
+    <%= link_to t("views.application.dashboards.buttons.team_submission_html"),
+      student_published_team_submission_path(current_team.submission),
+      class: "button" %>
+  </div>
+<% end %>
--- a/app/guards/student_submission_link_guard.rb
+++ b/app/guards/student_submission_link_guard.rb
@@ -1,0 +1,21 @@
+class StudentSubmissionLinkGuard
+  attr_accessor :team, :student, :season_toggles
+
+  def initialize(team:, student:, season_toggles: SeasonToggles)
+    @team           = team
+    @student        = student
+    @season_toggles = season_toggles
+  end
+
+  def display_link_to_new?
+    season_toggles.team_submissions_editable? && student.is_on_team? && team.submission.blank?
+  end
+
+  def display_link_to_in_progress?
+    student.is_on_team? && team.submission.present?
+  end
+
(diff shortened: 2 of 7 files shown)


app/views/student/dashboards/_view_submission_link.html.erb
```diff
@@ -0,0 +1,10 @@
+<% if !SeasonToggles.team_submissions_editable? &&
+    current_student.is_on_team? &&
+    current_team.submission.complete? %>
```
stenington, line 3: Since the if condition here is the same as part of the condition controlling the "My Team's Submission" tab, I would consider pulling it out into a guard. It feels to me like those two things would change together: if we want to adjust the tab, we probably want to adjust this new button similarly.

shaunxp20: I agree! But I can't think of a good/clear/explicit name for it! 😬 🤯 Any suggestions (factoring in the other logic in `app/views/student/navigation/_global_nav.html.erb`)? 🤔

stenington: Looking at the other template where there are three branches in the if/else, I could imagine doing something like
* `StudentSubmissionLink.should_link_to_published?`
* `StudentSubmissionLink.should_link_to_in_progress?`
* `StudentSubmissionLink.should_link_to_new?`

That goes a little bit against the pattern where guards just define a `self.call` method but I don't think I'm bother by that.

Although I'm starting to think we should address both templates at once, and test the logic used in the navigation as well, so maybe all of this work should be described in a new ticket and we can take this one as-is?

shaunxp20: I like where this is headed! 

I see how it goes against our current guard pattern; and it also goes against the "exit early" style guard pattern too. But I could see it working, perhaps by "newing" up a guard: `StudentSubmissionLinkGuard.new({ current_student: ... current_team: ... })`, and then you can ask it, "Hey guard, can I display the published link?" with something like `display_link_to_published?`, etc. I have no experience writing guards this way, but I could see it working - the guard guards the links. 🤷‍♂️ 😺 🤔 Thoughts? 🤔 

(A random side-tangent, one of my favorite things about Ruby is being able to use `?` in method names, which basically (by convention) tells you yes/no, true/false; so imho Ruby doesn't need "is" or "should" in method names, like other languages that don't have the `?` do, but that's just like my opinion. 😄 🎳)

stenington: Yeah, that sounds like it would be great!

shaunxp20: I made these changes, when you get a chance let me know what you think. 😄 🤔 🌞

Issue #2503, opened 2020-04-29
Edit screen judges & mentors see when they do not have a certificate [8,3,2.67]


rsgonzal:
Not everyone gets a certificate at the end of a season and currently those users would be seeing an outdated screen.

<img width="1228" alt="Screen Shot 2020-04-28 at 11 46 38 AM" src="https://user-images.githubusercontent.com/22801026/80651077-0d9eec80-8a2a-11ea-818d-c400b7ddfa6c.png">

It looks like multiple groups see this content block:
- mentors who were not matched with a team
- suspended judges
- judges that scored less than 5 submissions

Thank you for your participation in the 2020 Technovation season.
You can see this year's finalists at [technovationchallenge.org/season-results](technovationchallenge.org/season-results.).
The 2021 season will open in the Fall.
Stay informed about important dates and updates. [Sign up for our newsletter.](https://confirmsubscription.com/h/d/15B1DBF38F40C405)

We asked Bryant for feedback on this screen. At the very least, the two links need to be checked/changed.

We should also verify is anyone else sees this screen

## QA STEPS

As an admin, turn on the "Scores & Certificates Accessible" setting.

* Find a fully on-boarded judge with < 5 submissions, or a suspended judge, login as this judge
* Verify the new [content/url changes](https://github.com/Iridescent-CM/technovation-app/issues/2503#issuecomment-631577982) on the judge's dashboard
* You can also verify this content gets shown to a fully on-boarded mentor that isn't on a team; after logging in as the mentor, the content can be accessed/viewed from the "Scores & Feedback" link (in the sidebar)

hellafitz:
@rsgonzal do we have updated content or feedback from Bryant for this?

Bryant-Gomer:
I think these bullet points are fine on the screen are fine. I would add:
"Questions or feedback about staying involved with Technovation, reach out to volunteer@Technovation.org". 
"Make sure to visit [Technovation's blog](https://www.technovation.org/blog/) for program updates and announcements"

rsgonzal:
Thank you for your participation in the 2020 Technovation season.
You can see this year's finalists [here](technovationchallenge.org/season-results).
The 2021 season will open in the Fall.
Stay informed about important dates and updates. [Sign up for our newsletter](http://eepurl.com/gTS2KL).
Questions or feedback about staying involved with Technovation, reach out to volunteer@Technovation.org.
Make sure to visit Technovation's blog for program updates and announcements.

[...]

PR #2534, opened 2020-05-28
#2503 Update "no certificate" content

shaunxp20:
Content changes for the "no certificate" page.

#2503

### DIFF
--- a/app/views/dashboards/_no_certificates.erb
+++ b/app/views/dashboards/_no_certificates.erb
@@ -7,21 +7,11 @@
 
   <h1>Thank You</h1>
 
-  <p>
-    Thank you for your participation in the
-    <%= Season.current.year %> Technovation season.
-  </p>
+  <p>Thank you for participating in this season of Technovation Girls.</p>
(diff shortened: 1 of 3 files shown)


app/views/dashboards/_no_certificates.erb
```diff
@@ -7,21 +7,11 @@
...
-
-  <p>The <%= Season.next.year %> season will open in the Fall.</p>
-
-  <p>
-    Stay informed about important dates and updates.
-    <%= link_to "Sign up for our newsletter", ENV.fetch("NEWSLETTER_URL") %>.
-  </p>
-</div>
\ No newline at end of file
+  <p>Stay informed about important dates and updates. <%= link_to "Sign up for our newsletter", ENV.fetch("NEWSLETTER_URL") %>.</p>
```
stenington, line 14: `NEWSLETTER_URL` in production is set to https://confirmsubscription.com/h/d/15B1DBF38F40C405 whereas this ticket specifies something different. I'm assuming that current link can be replaced? Just making sure there aren't two different newsletters used in different contexts.

shaunxp20: We're good with the new URL, btw. 👍 :shipit: 😸

Issue #2474, opened 2020-04-09
Remove unused rebrand banner and "evolving" page content from site


stenington:
#2471 is repurposing the old rebrand banner for a covid 19 message. The work for this ticket would be to rename or remove:

* the `.rebrand__banner` and `.rebrand` styles
    * the banner styles should move to a more appropriately named file and get renamed drop the "rebrand" part, while the rebrand page styles can go away altogether I think
* the `RebrandController` controller
* the `general_info/rebrand` template
* any assets associated only with that page
* the routing for `/evolving`

shaunxp20:
This one is on the QA site, and a note about QA has been added as well. 😄 👍 😁

rsgonzal:
Banner looks great!

shaunxp20:
> Banner looks great!

Thanks @rsgonzal! 🙇 😁 👍 🙌

PR #2535, opened 2020-05-29
#2474 Remove/update code for the rebrand banner

shaunxp20:
As per #2474, I removed/updated the rebrand code.


app/assets/stylesheets/banners.scss
```diff
@@ -0,0 +1,23 @@
+.banner--homepage {
```
shaunxp20, line 1: This is the old `rebrand__banner` styles that I moved/renamed. I used the `notices` SCSS as a guide/reference, which uses selector names like `notice--alert`; and I used the view filename `_homepage_banner.en.html.erb` as reference for this selector name. 🤷‍♂️ Open to renaming any/all of it! Just throwing something up to see if it'll stick/work! 😁

shaunxp20:
> Cool, looking good. There are at least two assets you can remove as well, `evolving1.png` and `evolving2.png`.

Good catch, I removed those images too! 🙅 ❌ 🙅‍♂️

Issue #2460, opened 2020-04-06
Security: Update to resolve actionview alert


stenington:
As part of the ongoing #2166, resolve this alert: https://github.com/Iridescent-CM/technovation-app/network/alert/Gemfile.lock/actionview/open

PR #2541, opened 2020-06-02
Address actionview security vulnerability

shaunxp20:
This is to address the actionview security vulnerability:
https://github.com/Iridescent-CM/technovation-app/network/alert/Gemfile.lock/actionview/open

I tried upgrading actionview to `5.2.4.2`, but ran into this issue:
```
Bundler could not find compatible versions for gem "actionview":
  In Gemfile:
    actionview (>= 5.2.4.2)

    country_state_select (~> 3.0) was resolved to 3.0.5, which depends on
      rails was resolved to 5.1.7, which depends on
        actionview (= 5.1.7)
```

Which feels out of scope for this ticket to me, so instead, I applied the workaround patch provided in the link above.

#2460

### DIFF
--- a/config/initializers/action_view.rb
+++ b/config/initializers/action_view.rb
@@ -22,6 +22,30 @@ ActionView::Base.field_error_proc = Proc.new do |html_tag, instance|
   html.html_safe
 end
 
+ActionView::Helpers::JavaScriptHelper::JS_ESCAPE_MAP.merge!(
+  {
+    "`" => "\\`",
+    "$" => "\\$"
+  }
+)
+
+module ActionView::Helpers::JavaScriptHelper
+  alias :old_ej :escape_javascript
+  alias :old_j :j
+
+  def escape_javascript(javascript)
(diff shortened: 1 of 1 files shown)


config/initializers/action_view.rb
```diff
@@ -22,6 +22,30 @@
...
+    "`" => "\\`",
+    "$" => "\\$"
+  }
+)
+
+module ActionView::Helpers::JavaScriptHelper
+  alias :old_ej :escape_javascript
+  alias :old_j :j
+
+  def escape_javascript(javascript)
```
shaunxp20, line 36: I couldn't reproduce the vulnerability (only spent a little bit of time on it), but I did prove out this method is actually getting called by adding some log statements here, and then in another file calling `escape_javascript`; and I saw my log statements. 📜 🤓

shaunxp20:
Btw, I added a ticket to upgrade to Rails 5.2 (#2542), and a note to remove this patch when we upgrade. 👍 😀

PR #2559, opened 2020-06-23
specify node version

stenington:
We weren't telling Heroku what version of node we wanted, and it switched from 10.x to 12.x which broke the canvas build. This tells it to use 10.x which I'm hoping will fix deployments, although I'm not sure how much longer that version will be supported. See https://devcenter.heroku.com/articles/nodejs-support


stenington:
Heroku just began installing node 12 again. The fix seems to be adding the `heroku/nodejs` buildpack first in the list, which properly installs the specified node and lets the ruby buildpack use that instead of attempting the installation itself.

Issue #2550, opened 2020-06-16
Tests shouldn't fail when season rolls over [40,13,3.1]


stenington:
Last year, certain tests began failing when the season rolled over until date config was updated, and certificate templates were copied into a new directory. See https://github.com/Iridescent-CM/technovation-app/pull/2233 

I don't think the test suite should pass one day and fail the next, with no changes in between other than the actual date. These tests should probably be abstracted away from the actual app settings, and maybe we should be testing multiple cases (e.g. before a configured date, after a configured date).

stenington:
These are the failures I get by setting the 2020 dates in my `.env` back to 2019, so it doesn't include failures for missing certificate templates:

```
rspec ./spec/features/mentor/find_a_mentor_spec.rb:36 # Mentors find a team browse nearby mentors
rspec ./spec/features/mentor/find_a_mentor_spec.rb:117 # Mentors find a team visit the mentor page
rspec ./spec/features/mentor/find_a_team_spec.rb:114 # Mentors find a team request to join a team
rspec ./spec/features/mentor/find_a_team_spec.rb:132 # Mentors find a team request to join the same team again, even if you deleted an earlier request
rspec ./spec/features/student/find_a_team_spec.rb:31 # Students find a team request to join a team
rspec ./spec/features/student/find_a_team_spec.rb:45 # Students find a team onboarded student sees pending requests
rspec ./spec/system/student/request_to_join_a_team_spec.rb:79 # Students request to join a team a valid student requestor student accepts the request
rspec ./spec/system/student/request_to_join_a_team_spec.rb:118 # Students request to join a team a valid student requestor student accepts from team page
rspec ./spec/system/student/request_to_join_a_team_spec.rb:132 # Students request to join a team a valid student requestor student declines from team page
rspec ./spec/system/student/request_to_join_a_team_spec.rb:146 # Students request to join a team a valid student requestor mentor accepts the request
rspec ./spec/system/student/request_to_join_a_team_spec.rb:165 # Students request to join a team a valid student requestor mentor accepts from team page
rspec ./spec/system/student/request_to_join_a_team_spec.rb:179 # Students request to join a team a valid student requestor mentor declines from team page
rspec ./spec/system/student/request_to_join_a_team_spec.rb:193 # Students request to join a team a valid student requestor student declines the request
rspec ./spec/system/student/request_to_join_a_team_spec.rb:219 # Students request to join a team a valid student requestor mentor declines the request
```

[...]

PR #2570, opened 2020-07-08
#2550 rollover test fixes

stenington:
I think this will fix some tests that previously relied on config in the .env matching the actual date, i.e the `DATES_*_YEAR` config needed to agree with `Season.year.current` or tests would fail.

It also fixes some tests that started to break when I set the season rollover date to be June 1. Doing so somehow changed the division calculations so that students intended to be in the junior division were being calculated as senior.

### DIFF
--- a/spec/models/account_spec.rb
+++ b/spec/models/account_spec.rb
@@ -3,9 +3,9 @@ require "./lib/fill_pdfs"
 
 RSpec.describe Account do
   let(:season_with_templates) { Season.new(2020) }
-  before {
+  before do
     allow(Season).to receive(:current).and_return(season_with_templates)
-  }
+  end
 
   context "validations" do
     describe "student email address validations" do
--- a/spec/models/account_spec.rb
+++ b/spec/models/account_spec.rb
@@ -2,6 +2,10 @@ require "rails_helper"
 require "./lib/fill_pdfs"
 
 RSpec.describe Account do
+  before {
+    allow(Season).to receive(:current).and_return(Season.new(2020))
+  }
+
   context "validations" do
     describe "student email address validations" do
       let!(:student_account) { FactoryBot.build_stubbed(:account, email: student_email_address) }
--- a/spec/models/account_spec.rb
+++ b/spec/models/account_spec.rb
@@ -2,8 +2,9 @@ require "rails_helper"
 require "./lib/fill_pdfs"
 
 RSpec.describe Account do
+  let(:season_with_templates) { Season.new(2020) }
   before {
-    allow(Season).to receive(:current).and_return(Season.new(2020))
+    allow(Season).to receive(:current).and_return(season_with_templates)
   }
 
   context "validations" do
--- a/spec/models/account_spec.rb
+++ b/spec/models/account_spec.rb
@@ -2,7 +2,7 @@ require "rails_helper"
 require "./lib/fill_pdfs"
 
 RSpec.describe Account do
-  let(:season_with_templates) { Season.new(2020) }
+  let(:season_with_templates) { instance_double(Season, year: 2020) }
   before do
     allow(Season).to receive(:current).and_return(season_with_templates)
   end
--- a/spec/models/team_spec.rb
+++ b/spec/models/team_spec.rb
@@ -234,7 +234,7 @@ RSpec.describe Team do
     team.reload
 
     old_student = FactoryBot.create(:student, date_of_birth: 15.years.ago)
-    young_student = FactoryBot.create(:student, date_of_birth: 14.years.ago)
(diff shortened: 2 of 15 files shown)


stenington:
@shaunxp20 I'm happy for you to review this one for readability without worrying about changing your config to simulate the season rollover. If you want to check the rollover, what I did was leave all my date config in the `.env` as 2020, but make `DATES_NEW_SEASON_START_MONTH` and `DATES_NEW_SEASON_START_DAY` something in the past, 7/4 in my case. That way the software generally thinks it's the 2021 season, except where this PR makes changes to force it to think it's something else.

spec/models/account_spec.rb
```diff
@@ -2,6 +2,11 @@
 require "./lib/fill_pdfs"
 
 RSpec.describe Account do
+  let(:season_with_templates) { Season.new(2020) }
+  before {
+    allow(Season).to receive(:current).and_return(season_with_templates)
```
shaunxp20, line 7: 🤔 I see this pattern in various places. Just curious, wondering why the real `Season` is being called, and then stubbed? I'm probably missing something 😊 why not use the real `Season.current`? Or not use the real `Season` at all and return a double instead? 🤔 😸 Also, quibble at the block syntax, in Ruby the convention is multi-line block are do/end, and single-line blocks use the curlys, fwiw. 😁

spec/models/team_spec.rb
```diff
@@ -234,7 +234,7 @@
     team.reload
 
     old_student = FactoryBot.create(:student, date_of_birth: 15.years.ago)
-    young_student = FactoryBot.create(:student, date_of_birth: 14.years.ago)
+    young_student = FactoryBot.create(:student, date_of_birth: 13.years.ago)
```
shaunxp20, line 237: These age changes also caught my eye. 👀 Just curious why they had to change? 🤔 😄

spec/models/account_spec.rb
stenington: > why not use the real Season.current

That's because e.g. this year when the season ticks over to 2021, there will be no 2021 certificate templates in the repo yet, and the test suite will fail until they're added, which isn't particularly helpful in my opinion. Especially because we won't have real 2021 certificates at that point, so we'll just copy 2020 certificates as 2021 anyway most likely.

>Or not use the real Season at all and return a double instead

Yeah, could do that. Would there be an advantage to using a double here?

> quibble at the block syntax

Oh cool, I will try to follow that convention. I wasn't aware of it.

spec/models/team_spec.rb
stenington: 14 was too close to the boundary condition I guess. When I changed the season rollover date to early July, it broke the tests I expected it to break, but it also made the platform think `young_student`, in this case, was a senior division student when they need to be a junior division student. So making them even younger fixed that.

spec/models/account_spec.rb
shaunxp20: > > why not use the real Season.current
> 
> That's because e.g. this year when the season ticks over to 2021, there will be no 2021 certificate templates in the repo yet, and the test suite will fail until they're added, which isn't particularly helpful in my opinion. Especially because we won't have real 2021 certificates at that point, so we'll just copy 2020 certificates as 2021 anyway most likely.

My "why not use the real Season current" comment was referring to the line two lines above where we're setting the season via `Season.new`; I think I'm missing something, what's the point of stubbing `Season.current` to the thing that was just set?

> > Or not use the real Season at all and return a double instead
> 
> Yeah, could do that. Would there be an advantage to using a double here?

Imho, the advantage would clarity; and more isolation from `Season`, or fully committing to `Season`. Basically, my initial inquiry was just that I'm curious/confused why the mix of real vs stub, why not one or the other? 🤔

spec/models/team_spec.rb
shaunxp20: Huh, that seems odd! 👃 😂 👍 👌

spec/models/account_spec.rb
stenington: Oh, because later in the year `Season.current` will return 2021 which won't have templates and will break. Specifying 2020 is redundant for now, but will allow these tests to continue to pass in the future.

I'll try putting in a double instead and seeing if that makes more sense. I don't have a principled reason for using the real Season, it's just what came out of my investigating and allowed these tests to pass even when I configured the app to think the 2021 season had started.

spec/models/team_spec.rb
stenington: Yeah, it's very odd! Actually it makes me nervous that there's some weird boundary condition bug lurking in the app around divisions, I was considering opening a ticket for further exploration of that maybe.

Issue #2487, opened 2020-04-23
Write integration tests for admin Content & Settings controls [20,13,1.54]


stenington:
Write integration tests covering the admin Content & Settings section. This could range from a basic "able to change a simple setting" happy path, to a number of tests covering much more complex season scenarios. That part is TBD.

hellafitz:
- create a map together
- define the "basic path"
- cover as much as you can within the available time
- include documentation/notes about the things you don't cover so we can expand testing there in the future (this could be in the form of a new ticket)

stenington:
@shaunxp20 I'm curious how you would structure tests for this ticket. 

To my mind, the thing I care about is that when the user makes a series of selections, reviews it, and saves it, the appropriate values on `SeasonToggles` are updated to reflect those choices. So my knee-jerk reaction is to write integration tests where I manipulate the UI to make selections, review and save, and then assert against `SeasonToggles` for the values I would expect. You've called it out before though when I wrote assertions that peeked behind the curtain, rather than only asserting against what the user is able to see. So that's my first question: what would you query against?

Secondly, what do you think about exercising each setting individually versus making a larger batch of changes and assertions at once? I've read advice that recommended having fewer integration tests that do more in one go to try to minimize the time penalty you incur with slower integration tests, but the trade off then is less granularity when they fail.

Below are notes for myself on the basic correspondences between the settings UI and `SeasonToggles`:

## Registration
**Basic tests**
For each checkbox, test that the checkbox enables/disables `SeasonToggles.<type>_signup?`

**Additional considerations**
I don't think the platform actually obeys this setting for judges or regional ambassadors.

## Notices
**Basic tests**
For each text field, the text entered ends up in `SeasonToggles.dashboard_text(:<type>)`

**Additional considerations**
Handling of special characters?
I don't believe the platform makes use of these notices in all the dashboards.

## Surveys
**Basic tests**
For each user type, values put into the fields end up in `SeasonToggles.survey_link(:<type>, <key>)` for the appropriate keys.

## Teams & Submissions
**Basic tests**
*Forming teams allowed* should control `SeasonToggles.team_building_enabled?` and *Team submissions are editable* shoud control `SeasonToggles.team_submissions_editable?`

## Events
**Basic tests**
*Selecting regional pitch events allowed* should control `SeasonToggles.select_regional_pitch_event?`

## Judging
**Basic tests**
The selected radio button should end up in `SeasonToggles.judging_round`

## Scores & Certificates
[...]

PR #2575, opened 2020-07-10
#2487 Add integration spec for content and settings page

stenington:
This drives the content and settings admin page and checks that the appropriate `SeasonToggles` are set.

I tried the given/when/then format as well, but it felt to me like it was harder to see the connection between UI and `SeasonToggles` without adding much clarity.


shaunxp20:
I didn't see any specs for setting the judging round and how that turns on/off other settings. I think it would be good to test the more complicated scenarios. I'm fine w/ another ticket if we want to just get this one wrapped up. Thoughts?

stenington:
I'll try to add those here, but I'll kick it over to another ticket if it becomes too much. Good catch!

stenington:
@shaunxp20 Okay, this is ready for another look. I'm now testing that certain judging rounds:
* disable other toggles on the backend
* show alerts that things have been disabled, and
* uncheck and disable the actual checkboxes too

while other judging rounds don't do the above. The file has gotten pretty long, and I keep getting tempted to try to do more clever dynamic things, but I think the simple procedural style will ultimately be more comprehensible. What do you think?

shaunxp20:
> @shaunxp20 Okay, this is ready for another look. I'm now testing that certain judging rounds:
> 
>     * disable other toggles on the backend
> 
>     * show alerts that things have been disabled, and
> 
>     * uncheck and disable the actual checkboxes too
> 
> 
> while other judging rounds don't do the above. The file has gotten pretty long, and I keep getting tempted to try to do more clever dynamic things, but I think the simple procedural style will ultimately be more comprehensible. What do you think?

I definitely prefer readability in tests, so doing w/o more clever dynamic things is fine w/ me!

Issue #2492, opened 2020-04-24
Write integration tests on RA event creation/management screen [13,20,0.65]


stenington:
Write integration tests for basic scenarios involving RAs creating and managing their events: creating the event, managing teams and judges, assign judges to a team, etc.

hellafitz:
- map it out collaboratively
- work it
- document anything we aren't able to tackle within the allocated effort

shaunxp20:
@stenington, when you have a moment, here's the scenarios I was thinking for this one:

RPEs:
 * creating/viewing an event
 * editing an event
 * deleting an event

Teams:
 * a team can be added to an event
 * a team can be removed from an event
 * a team cannot be added if it goes over the event capacity

Judges:
 * a judge can be added to an event
 * a judge can be removed from an event

Other scenarios:
 * printing an event
 * exporting teams to CSV
 * exporting judges to CSV

Thoughts? Thoughts on "other scenarios"? Any other scenarios we need to cover?

stenington:
I think covering that list of scenarios would be great. 

There's a feature that gets enabled if an event has 10+ teams, which is that judges can be assigned directly to teams. I believe the logic was that in big RPEs, the organizers would sometimes partition groups of judges and have them each judge a subset of submissions to limit judging fatigue. So there might be an extra set of scenarios around RPEs with 10+ submissions to assert that a judge can be assigned and un-assigned from a team. I can't remember if the interface also lets you do it from the team side.

shaunxp20:
I wrote tests for the bullet points under the headers: RPEs, Teams, and Judges in https://github.com/Iridescent-CM/technovation-app/issues/2492#issuecomment-657807468

I created a new issue (#2579) for the "Other scenarios", and for the 10+ teams scenarios.

hellafitz:
@shaunxp20 this is just an it-27 ticket, correct? I think that it-26 label slipped by us in Planning. If so, please remove.

shaunxp20:
> @shaunxp20 this is just an it-27 ticket, correct? I think that it-26 label slipped by us in Planning. If so, please remove.

It's an it-26 ticket. 👍

shaunxp20:
After so many broken builds and the ups and downs with these tests, I am finally able to get them passing on CircleCI. A couple of issues worth mentioning that I ran into:
[...]

PR #2578, opened 2020-07-15
#2492 Add feature specs for Regional Pitch Events

shaunxp20:
This will add feature specs for these scenarios:

RPEs:
 * creating/viewing an event
 * editing an event
 * deleting an event

Teams:
 * a team can be added to an event
 * a team can be removed from an event
 * a team cannot be added if it goes over the event capacity

Judges:
 * a judge can be added to an event
 * a judge can be removed from an event

### DIFF
--- a/spec/system/chapter_ambassador/regional_pitch_event_judges_spec.rb
+++ b/spec/system/chapter_ambassador/regional_pitch_event_judges_spec.rb
@@ -1,0 +1,73 @@
+require "rails_helper"
+
+RSpec.describe "Regional Pitch Event Judges", :js do
+  let(:chapter_ambassador) { FactoryBot.create(:chapter_ambassador, :approved) }
+
+  context "when MANAGE_EVENTS is enabled" do
+    before do
+      allow(ENV).to receive(:fetch).and_call_original
+      allow(ENV).to receive(:fetch).with("MANAGE_EVENTS", any_args).and_return(true)
+    end
+
+    it "successfully adds a judge an event" do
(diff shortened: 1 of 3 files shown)


shaunxp20:
I think this is finally ready for a looksie, the build is passing more consistently now (five times in a row currently). I think I mentioned before about how specific/brittle the selectors are, and definitely not the best tests I have written, but it accomplishes the objective of wrapping some of the RPE stuff in tests. :-/ ¯\_(ツ)_/¯

stenington:
I got this failure on 3/3 runs locally: 

```
Failures:

  1) Regional Pitch Events successfully creates a new event
     Failure/Error: expect(find_field("Date").value).to be_present
       expected `"".present?` to return true, got false
     
     [Screenshot]: tmp/screenshots/failures_r_spec_example_groups_regional_pitch_events_successfully_creates_a_new_event_182.png

     
     # ./spec/system/chapter_ambassador/regional_pitch_events_spec.rb:14:in `block (2 levels) in <main>'

```

The screenshot is this:

failures_r_spec_example_groups_regional_pitch_events_successfully_creates_a_new_event_182

Any thoughts on what might be going on?

stenington:
Oh, I think maybe the test is trying to pick a date that's disabled based on my local environment maybe?

stenington:
The latest commit doesn't fix this issue for me. I think it's because on the js side, environment config isn't read in dynamically, it's essentially compiled in by webpack.

shaunxp20:
When you get a chance, can you see if this latest push fixed the issue you were seeing? 🤞

stenington:
@shaunxp20 Yep, they passed!

Issue #2542, opened 2020-06-02
Upgrade Rails to version 5.2 [20,8,2.5]


shaunxp20:
We're currently using Rails 5.1, and it looks like Rails 5.1 hasn't been updated in over a year:
https://weblog.rubyonrails.org/2019/3/28/Rails-5-1-7-has-been-released/
https://github.com/rails/rails/releases/tag/v5.1.7

One downside of using an older version of Rails is that fixes for security vulnerabilities aren't getting added to it (like #2460).

Here is a guide for upgrading to Rails 5.2 from Rails 5.1:
https://guides.rubyonrails.org/5_2_release_notes.html#upgrading-to-rails-5-2
https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-5-1-to-rails-5-2

When we do the upgrade, we should remove the security patches that were added for Rails 5.1:
 * #2541

## QA

No specific QA here.

PR #2587, opened 2020-07-17
#2542 upgrade rails to 5.2 with changes to get tests passing

stenington:
I believe these changes get us a passing build on rails 5.2


Gemfile
```diff
@@ -1,7 +1,7 @@
 source 'https://rubygems.org'
 ruby "2.6.4"
 
-gem 'rails', '~> 5.1.0'
+gem 'rails', '~> 5.2.0'
 # from GitHub security alert
 gem "actionview", ">= 5.1.6.2"
```
shaunxp20, line 6: Wondering if this can be removed?

Gemfile
```diff
@@ -16,10 +16,7 @@ gem 'bootsnap', '~> 1.3', require: false
 gem 'fast_jsonapi', "~> 1.4"
 gem 'httparty', "~> 0.16"
 
-# Patch applied: https://github.com/Iridescent-CM/textacular/commit/99cf5ef6cff0129b75bed261c6dd8e765f04a0c8
-gem 'textacular',
-  git: 'https://github.com/Iridescent-CM/textacular.git',
-  branch: 'change-assemble-query'
```
shaunxp20, line 22: 👍 Would be nice to remove that repo at some point too.

db/migrate/20191008152233_remove_deleted_account_email_index.rb
```diff
@@ -1,5 +1,5 @@
 class RemoveDeletedAccountEmailIndex < ActiveRecord::Migration[5.1]
   def change
-    remove_index :accounts, :email
+    remove_index :accounts, name: "index_accounts_on_email"
```
shaunxp20, line 3: Hmm, won't this be okay since the migration is scoped to Rails 5.1? Or do we need to update the version to 5.2 so it's consistent w/ the new syntax?

stenington: I don't know, I made this change because the build broke on CI without it. I can dig a little deeper if this seems wrong to you.

Gemfile
stenington: Oh, good catch. I'll check. I also realized I need to back out another patch that was put in as well.

db/migrate/20191008152233_remove_deleted_account_email_index.rb
stenington: It seems like Rails 4 would allow the old syntax here, but Rails 5 only claims to handle it if exactly one index on that table and column exist, which is not the case by the time this migration is running. So it seems like this never should have worked in 5? 

https://apidock.com/rails/v4.2.9/ActiveRecord/ConnectionAdapters/SchemaStatements/remove_index
vs
https://apidock.com/rails/v5.2.3/ActiveRecord/ConnectionAdapters/SchemaStatements/remove_index

shaunxp20: Huh, that's weird, but I guess if CI is good w/ it and everything is working, seems like it's okay.

Issue #2509, opened 2020-05-06
Environment variable cleanup [20,20,1]


stenington:
As groundwork for #2502, we should work through the various environment variables found in our .env files, staging, and production, and do two things:
* eliminate old config the app doesn't rely on, and
* identify required config that would be a candidate for an initializer as described in #2502

stenington:
I find something like ~46 environment variables that I think are unused, and if we clean up Heroku add-ons that we have in staging but aren't actually using, we can probably remove a few more. That seems like a good start on this one, I think?

stenington:
I would recommend messing with config and add-ons after this season ends, though. If it were to mess something up badly, we shouldn't do it at a time where we may want to move fixes through to production relatively quickly. I'm adding the off season tag to that end.

stenington:
This doc is a work in progress, but contains my notes on comparing the environments so far: https://docs.google.com/spreadsheets/d/1yfIXVJkWLDrvcthtWHPfHhxVOFtrO-lOMQb81YtPABc/edit?usp=sharing

stenington:
I removed sendgrid and bonsai add-ons in staging, because we don't have them in production and I believe they were unused.

stenington:
All config marked as unused has been dropped in staging. I have a snapshot of staging config prior to doing that in case we need to restore anything. I did the same thing locally and all tests passed.

stenington:
I don't think there are any QA steps on this one per se, but I think I'll wait a few days and see if anything major happens on QA before dropping the same config in production.

As a reminder, the idea was to do this when platform usage winds down, and to rely in large part on "exploratory QA" for crucial stages of the next season to ferret out any issues this may have introduced ahead of time.

stenington:
For the second part
> identify required config that would be a candidate for an initializer as described in #2502

out of 174 initial config variables, 87 would be appropriate to require in an initializer. For a few of them, I'd argue we should adjust code not to require them, but [this filtered view](https://docs.google.com/spreadsheets/d/1yfIXVJkWLDrvcthtWHPfHhxVOFtrO-lOMQb81YtPABc/edit#gid=0&fvid=1161495751) shows the ones I'd currently put in an initializer.

stenington:
Moving this one, because I've adjusted staging config and I think giving this one some time to sit is probably smart. I already realized `DEPLOY_TASKS` was in fact in use and needed to be put back.

[...]

PR #2603, opened 2020-08-12
#2509 remove totally unused config from circleci config/env spec

stenington:
As I was looking at the config that can be removed from production, I saw some bits that are totally unused but still referenced in our Circle CI config and/or the env spec. This clears those out since they're completely irrelevant at this point.


shaunxp20 (approved):
👍

Issue #2600, opened 2020-08-07
Update Postgres from v10 to v11 [13,8,1.63]


hellafitz:
Part of Epic #2599 and associated Theme to upgrade admin user search.

Deliverable: For both upkeep and to enable more advanced search on the admin, upgrade Postgres from v10 where we are to v11. 

Be sure to make an announcement in Slack for when the site is going to be down for this upgrade.

hellafitz:
@stenington do you think this should this be a spike first, or can we proceed directly?

hellafitz:
Decision: proceed directly. 

Steps notes (@stenington to edit/update with link) : 
- Use Heroku toolbelt to provision a new database of the version you want as a follower. Then wait for it to catch up. issue commands to shut down the old one and turn the follower into the main database. (There is some downtime; should be done within a day). 
- May need a new postgres version on CI to make sure version matches

stenington:
Heroku docs, we should be able to do the `pg:upgrade` option. https://devcenter.heroku.com/articles/upgrading-heroku-postgres-databases

stenington:
Nice, thanks for the notes @shaunxp20. I can take a crack at updating CI and staging tomorrow.

stenington:
Can't go to v12, `postgresql-client-12` doesn't seem to be available yet.

stenington:
Staging is now on postgres 11. `HEROKU_POSTGRESQL_GRAY_URL` is the v10 database in case we need that back. Otherwise it can be deprovisioned after this passes QA I think.

stenington:
Apparently the update method doesn't work to upgrade versions, so I'm following the copy steps for production as well.

PR #2611, opened 2020-08-24
#2600 use postgres 11 on CI

stenington:
Let's see how tests do in postgres 12


stenington:
`postgresql-client-12` isn't stable yet, so backing off to 11

PR #2614, opened 2020-08-27
Pre-filled student consent forms

stenington:
This is a proof of concept PR for making partially pre-filled parental consents available for the student to download using `pdfkit`.


stenington:
Not included yet:

* instructions on installing `wkhtmltopdf` in your development environment
* tests on the new endpoint
    * tests for content pre-fill on html endpoint
    * test for successful pdf response from pdf endpoint
* whatever will be needed to make this work on CI

And I don't yet know whether this will run on Heroku, although I do have the `wkhtmltopdf-heroku` gem included for staging and production.

stenington:
Things to know:

* `wkhtmltopdf` may be slow in production, that would be something for us to monitor (rather than over-engineering up front) and may need rework after we see a season.
* Pixel-perfect rendering of the pdf consent form is unlikely, but I think we can get something good enough.
* Trying to render the online consent form shown to parents caused `wkhtmltopdf` to hang indefinitely on my local machine, but it seems able to successfully render a more tightly controlled page (i.e. the tool seems finicky).

stenington:
Here's an example of the pdf it produces: [downloadable_parental_consent.pdf](https://github.com/Iridescent-CM/technovation-app/files/5136975/downloadable_parental_consent.pdf)

hellafitz:
related idea: new admin page for managing consent forms and directly looking up students when looking at a paper form

hellafitz:
Next steps:
- Deliverable for this ticket: Implement this proof of concept, including confirming on Heroku that this process can actually generate a pdf
- Then, discuss w/ PT to identify and prioritize related set of changes to improve admin search

stenington:
Recap for myself next week:
* `wkhtmltopdf` wasn't working in tests locally when installed by hand, but worked when installed through the `wkhtmltopdf-binary-edge` gem
* the `wkhtmltopdf-binary-edge` gem didn't seem to work on CI, returning an html result when pdf was expected although I didn't see the actual error
* the test suite has been updated to dump the actual error in the scenario above now
* I'm unclear if `wkhtmltopdf-binary-edge` or `wkhtmltopdf-heroku` are getting in the way on CI, but it seems like `pdfkit` cannot find a manual install on CI without further config, which I have not yet tried
* I don't think Gemfile groups don't work like I thought they did

Issue #2443, opened 2020-03-25
Make dates configurable in Welcome Email to students [8,8,1]


rsgonzal:
Email subject line: Welcome to Technovation 2020!

Registration deadline did not change with the updated configuration dates. It's not that important for this season, but may want to consider pulling from config for next season.

![Screen Shot 2020-03-20 at 1 54 53 PM](https://user-images.githubusercontent.com/22801026/77517812-f91f7100-6e39-11ea-880d-bac63a6f57c4.png)

## QA STEPS

1. Set these configuration settings on QA:

```
DATES_REGISTRATION_OPENS_YEAR
DATES_REGISTRATION_OPENS_MONTH
DATES_REGISTRATION_OPENS_DAY
```

```
DATES_OFFICIAL_START_OF_SEASON_YEAR
DATES_OFFICIAL_START_OF_SEASON_MONTH
DATES_OFFICIAL_START_OF_SEASON_DAY
```

```
DATES_TEAM_REGISTRATION_DEADLINE_YEAR
DATES_TEAM_REGISTRATION_DEADLINE_MONTH
DATES_TEAM_REGISTRATION_DEADLINE_DAY
```

2. Register as a new student, and verify the dates in the welcome email are correct (based off the configuration settings in step 1)

hellafitz:
Deliverable: Make the first 3 bulleted dates pull from config (the 4th already does)

shaunxp20:
I made these dates dynamic via configuration settings. The new configuration settings are:
`DATES_OFFICIAL_START_OF_SEASON_YEAR`
`DATES_OFFICIAL_START_OF_SEASON_MONTH`
`DATES_OFFICIAL_START_OF_SEASON_DAY`

`DATES_TEAM_REGISTRATION_DEADLINE_YEAR`
`DATES_TEAM_REGISTRATION_DEADLINE_MONTH`
`DATES_TEAM_REGISTRATION_DEADLINE_DAY`

The "registration opens" date was already a configuration setting (`DATES_REGISTRATION_OPENS_X`), we just weren't using it in the welcome email.

I added the names of these new configuration settings on both QA and production. We'll just need to update the values of these settings before this goes live.

Also in the welcome email is a link to an info-graphic that I assume will need to be updated as well:
https://venngage.net/ps/bAAeKGLlsk/2019-technovation-important-dates-and-milestones

🤔 😃

shaunxp20:
[...]

PR #2625, opened 2020-08-28
#2443 Remove registration_opens setting from the registration job

shaunxp20:
As per our discussion, removing this code!

#2443

### DIFF
--- a/app/jobs/register_to_current_season_job.rb
+++ b/app/jobs/register_to_current_season_job.rb
@@ -2,7 +2,7 @@ class RegisterToCurrentSeasonJob < ActiveJob::Base
   queue_as :default
 
   def perform(record)
-    return false if perform_not_needed?(record)
+    return false if already_registered_for_season?(record)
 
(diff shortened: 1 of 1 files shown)


app/jobs/register_to_current_season_job.rb
```diff
@@ -14,14 +14,9 @@ def perform(record)
...
-    if record.respond_to?(:season_registered_at)
-      record.seasons.include?(Season.current.year) &&
-        record.season_registered_at &&
-          record.season_registered_at.to_date >= ImportantDates.registration_opens
-    else
-      record.seasons.include?(Season.current.year)
-    end
+
+  def already_registered_for_season?(record)
+    record.seasons.include?(Season.current.year)
```
shaunxp20, line 19: After removing `registration_opens`, it looks like this method can be refactored down to one line; let me know if this looks right to you.

stenington: Yep, looks right to me!

Issue #2622, opened 2020-08-28
Pre-fill student consent forms with lookup info  [13,8,1.63]


stenington:
Currently students can download a blank consent form from their dashboard in order to get consent from a parent who does not have an email address, or is having trouble receiving the consent email. The problem with this system is that the parent doesn't always provide the necessary information or the correct information (on the form or in the email they send) to look up the student account. 

The idea in this ticket is to provide the student with a parental consent form that has the key lookup information pre-filled. The key look up info is probably their name and email address as it has been entered in the platform. 

![Screen Shot 2020-08-27 at 9 39 26 PM](https://user-images.githubusercontent.com/55319/91511663-35a23d80-e8ae-11ea-9b32-11cb5242737b.png)

#2614 began as a proof of concept for using `pdfkit` to generate the pre-filled pdf. We don't want to use the same method we use for certificates, as the underlying tool is no longer supported and it has been a little difficult for us to work with in the past.

Next steps:
- Deliverable for this ticket: Implement this proof of concept, including confirming on Heroku that this process can actually generate a pdf
- Then (outside of this ticket's scope), discuss w/ PT to identify and prioritize related set of changes to improve admin search

Reference: [student consent form process map on Invision](https://projects.invisionapp.com/freehand/document/k82pbNbfy)

related idea: new admin page for managing consent forms and directly looking up students when looking at a paper form

stenington:
## QA steps:

1. Log in as a student who hasn't gotten parent consent yet
2. Click the "Download the permission form" link
3. Confirm that your name and email have been pre-filled
4. Go back and enter your parent name and email, click "Send the form"
5. Click the "Download the permission form" link again, and now it should include your parent's name pre-filled

kate-technovation:
All steps worked as expected--would it be pushing it to hope that after step 4, the parent email address would get added to the PDF too? 

Either way, thank you so much!!

Screenshot of Student dashboard:
<img width="999" alt="Screen Shot 2020-09-02 at 5 04 13 PM" src="https://user-images.githubusercontent.com/57409606/92036896-ef7e2b80-ed3e-11ea-9159-044cf0d6c147.png">

Screenshot of pre-filled parent info:
<img width="970" alt="Screen Shot 2020-09-02 at 5 06 34 PM" src="https://user-images.githubusercontent.com/57409606/92036937-fdcc4780-ed3e-11ea-9fb7-614d695d29e8.png">

rsgonzal:
This looks great!

Issue #2617, opened 2020-08-27
Shut down TechnovationChallenge.org GSuite 9/3-4 [20,8,2.5]


hellafitz:
Close down the standalone TechnovationChallenge.org GSuite account and then add the domain to our primary GSuite account for IridescentLearning.org (/Technovation.org)

stenington:
Post in #dev-team when starting

stenington:
Done around 10:30am ET on 9/3, now waiting for the domain to free up within Google's systems to be added as a domain alias. It currently claims the domain is being used, and support docs say it can take up to 24 hours for that to change.

stenington:
i was able to add technovationchallenge.org as a domain alias in our iridescentlearning.org account, so now it’s a waiting game for that change to propagate. up to 72 hrs, i think the docs said.

stenington:
I believe emails are going through on the new domain alias today. @rsgonzal Can you test and confirm?

rsgonzal:
tested & confirmed!

PR #2628, opened 2020-08-31
#2622 pre-filled consent

stenington:
This is a cleaner version of PR #2617 
It should work on CI, don't yet know if it works on heroku.


stenington:
A note on the new libraries:
* `pdfkit` uses a utility called `wkhtmltopdf` under the hood to turn an html page into a pdf
* `pdf-reader` is used in tests to consume the new .pdf endpoint and parse it to check for the pre-filled values
* `wkhtmltopdf-heroku` is supposed to be a good way to make `wkhtmltopdf` available in your heroku environment

To install `wkhtmltopdf` locally in my development environment, `brew cask install wkhtmltopdf` worked great for me.

The new `app/views/student/downloadable_parental_consents/show.html.erb` contains inline styles because the utility didn't seem able to generate a pdf from a page with the normal app layout, it would just hang. I'm actually not sure whether or not these styles could be moved to an external stylesheet, but doing it this way was a good reminder to myself to keep things simple.

I'm not sure whether `PARENT_CONSENT_PDF` would be unused after this. We talked a little bit about maybe making it available directly as a backup, so it may depend on QA feedback as to whether or not that config should be removed.

shaunxp20 (approved):
Looks like we're adding a lot of dependencies, but looks like it will get the job done as well. Changes look okay to me, didn't test it though.

Issue #2620, opened 2020-08-27
Split user search fields in Admin [20,13,1.54]


hellafitz:
from epic #2599

- split the Name or Email fields apart into First Name and Last Name (searching results from any user type unless admin selects something specifically), Email
- split off the Parent Name (combined first and last)
- fields should search using exact match (which we'll confirm w/ PT and CET and Ops feedback after)

stenington:
In implementing this, I noticed that changing this for admins will change it for chapter ambassadors too.

stenington:
@hellafitz @rsgonzal I realized while working on this that the same code backs the chapter ambassador and admin participants pages. Would you want this change to happen for both types of user?

stenington:
There is a way for me to do this only for admins, which is what I will default to for now since that's how the ticket is written. But if chapter ambassadors should also get these new search fields, that's pretty straight forward to do I think.

rsgonzal:
I don't see anything wrong with chapter ambassadors getting the same changes. I think it would probably be helpful for them as well?

stenington:
## QA steps:

1. log in as an admin
2. make sure you can use each of the new search fields, and get the results you would expect
3. log in as a chapter ambassador
4. make sure the search fields work the same over there

kate-technovation:
Wow--thank you so much, Team!! I really, really appreciate this upgrade.

1. Done
2. Tried searching in First Name field for "North". Found exact results:
<img width="780" alt="Screen Shot 2020-09-08 at 6 59 45 PM" src="https://user-images.githubusercontent.com/57409606/92535555-79c60400-f205-11ea-9f62-acf3cccc318c.png">

Tried searching in Last Name field for "Carlsberg". Found exact results:
<img width="653" alt="Screen Shot 2020-09-08 at 7 00 26 PM" src="https://user-images.githubusercontent.com/57409606/92535591-906c5b00-f205-11ea-9817-65448fa0802c.png">

Tried searching in Email field for "ophelie+ca1@technovation.org". Found exact results:
<img width="630" alt="Screen Shot 2020-09-08 at 7 01 56 PM" src="https://user-images.githubusercontent.com/57409606/92535677-c578ad80-f205-11ea-816c-1cabdfafa530.png">

Added Parent or Guardian Name filter and searched on parent first name "Mom" and then parent last name "Tea" in the Parent or Guardian Name field. Found exact results both times:
[...]

PR #2629, opened 2020-08-31
#2620 split out admin search

stenington:
On the participants page, instead of "Name or Email" you can now search separately by First name, Last name, Email, Parent or guardian name, Parent or guardian email.


app/data_grids/accounts_grid.rb
```diff
@@ -407,16 +407,77 @@ class AccountsGrid
 
   filter :name_email,
     header: "Name or Email",
```
shaunxp20, line 409: Just double checking, were we going to split/remove the "Name or Email" field, or split/keep this field?

shaunxp20 (approved):
Didn't test it, but the additions looks okay to me! 👍

app/data_grids/accounts_grid.rb
stenington: And all the new fields are only shown to admins.

stenington:
Changed this back so both admins and chapter ambassadors have the same fields.

shaunxp20 (approved):

PR #2633, opened 2020-09-02
Make release stop pushing all branches

stenington:
@shaunxp20 does this look right? I think this would just push the right branches and we could take the branch clean up step out of the deployment process.


shaunxp20 (approved):
Looks fine to me. I think a `--dry-run` could confirm it's what we want, and/or just testing on the next deploy as you mentioned earlier. 👍

Issue #2606, opened 2020-08-17
Stop Mentors-who-converted-from-students from receiving student emails [8,13,0.62,]


rsgonzal:
When a student is converted to a mentor, you would expect they no longer need permission from their parents to participate. However, they are still receiving an email to sign the form. They also received the student welcome email. This was reported by Ophelie after logging in as a converted mentor on Staging. 

Open Qs: Is the platform treating them like a student in any other way?

## QA STEPS
* Ensure the welcome email and parental consent emails aren't being sent out to students who have been converted to mentors, one way to do this:
* As an admin, find a student who has participated in a previous season (but not the current one)
* Update this student's email address to your own, then convert this student to be a mentor
* Log in as this student/mentor, you shouldn't get the student emails

hellafitz:
@mdancho @ophelieh any idea how many account go from student to mentor each year (roughly)?

ophelieh:
Hi! I would say between 15 to 30, not including accounts that had been converted in previous seasons.

rsgonzal:
1. Stop converted mentors from receiving student emails
2. Stop sending out parent consent for these users

stenington:
Deliverable here is stopping the sending of the two emails mentioned above: student welcome and parental consent. 

The open question mentioned up top is still an open question, but should be explored elsewhere.

shaunxp20:
The fix I made for this: basically if an account has both a student and mentor profile, we'll just onboard them as a mentor, and none of the student onboarding code (which includes these two emails) will get run.

A more robust fix would be something like adding the `seasons` field to the profile tables (student, mentor, judge, chapter ambassador), then accounts could have multiple profiles and we could know which profile types participated in which season.

I also looked briefly at the other places that reference `student_profile` and it's too hard/time consuming to go through all of them to see how it could affect the student-mentor combo situation. Instead, we could reach out to mentors who also have a student account and see if they ran into any other issues, or just leave it be until someone notices something, or create another ticket to do more of a deep dive on this. 🤷 😄

shaunxp20:
This one is on QA, and some QA steps have been added as well. 😄 👍 🌞

ophelieh:
Hi all,

When I log in as a student who's now a mentor, my profile looks right and I did not receive the student emails - so this works. I received no email at all. Do mentors not receive a copy of the consent they sign? 

This is what I noticed - not related to emails though. If there is another ticket where to out this, let me know. 
[...]

PR #2634, opened 2020-09-03
#2606 Update conditions to #prepare_student_for_current_season

shaunxp20:
This is more of a band-aid fix, basically if an account has both a student profile and a mentor profile, we'll just treat them as a mentor, and not run `#prepare_student_for_current_season`, which includes the two emails we don't want to send out.

#2606


stenington (approved):
Yeah, seems reasonable. 👍

Issue #2613, opened 2020-08-26
Remove option to choose years participated in from RA registration 


rsgonzal:
The year chosen does not show up on the RAs profile and is not used by staff for any reason. This question only applies to a few RAs anyway. Team wants to remove.

![screencapture-technovation-qa-herokuapp-chapter-ambassador-signup-2020-08-25-17_19_55](https://user-images.githubusercontent.com/22801026/91240671-7ad03f00-e6f7-11ea-98a9-326e1e714059.png)
![Screen Shot 2020-08-25 at 5 20 24 PM](https://user-images.githubusercontent.com/22801026/91240679-7f94f300-e6f7-11ea-8991-01f3bf7632e9.png)

Remove Question: IN WHICH YEAR DID YOU BECOME A CHAPTER AMBASSADOR?

## QA STEPS
* Ensure "IN WHICH YEAR DID YOU BECOME A CHAPTER AMBASSADOR?" isn't showing up for chapter ambassadors in the profile view, or during sign up

shaunxp20:
Just double checking, do we want to remove the existing data in the database for this as well? 🤔 🚶 🌞 

@rsgonzal

rsgonzal:
yes! we can remove the data for that question.

shaunxp20:
This one is also on the QA site, and QA steps have been added as well! 😺 👍

ophelieh:
Tested as a new Chapter ambassador - worked as expected. Thanks! :)

ksrebecca:
+1, nice.

shaunxp20:
Thanks @ophelieh and @ksrebecca! 🎉 🍰 🍪 👍 😄

PR #2637, opened 2020-09-03
#2613 Remove `ambassador_since_year` field

shaunxp20:
This will remove the field, "In which year did you become a chapter ambassador?", its related code, and the data in the database.

#2613


stenington (approved):
👍

PR #2638, opened 2020-09-03
This may help with the memory errors we're seeing sometimes on CI

stenington:
I'm seeing this sort of thing in our javascript tests:

```
Summary of all failing tests
 FAIL  spec/javascript/utilities/vuex-utils.spec.js
  ● Test suite failed to run

    ENOMEM: not enough memory, read

      at Runtime.readFile (node_modules/jest-runtime/build/index.js:1910:21)
```

One suggestion I saw was to limit jest workers. Another is to upgrade your CircleCI resource class, which may require a paid account. I haven't checked that option yet: https://discuss.circleci.com/t/unit-tests-enomem-not-enough-memory-read/35695


shaunxp20 (approved):
🤞 👍

Issue #2642, opened 2020-09-09
Mentor details page standardization and accept/decline options [13,13,1]


rsgonzal:
From the student dashboard, a student can click on two types of mentors
- ones that are on their team
- ones that are not on their team yet but requested

![Screen Shot 2020-09-09 at 9 26 49 AM](https://user-images.githubusercontent.com/22801026/92626552-e6c9b000-f27e-11ea-8a10-9bd1aa807089.png)

Changes:
- add the same back button to the mentor detail page of a mentor on the team
- allow the student to approve the mentor on the page from a mentor who requested to join team, by adding the buttons we see on this page 
- remove the hyperlink back to My Team invites:
![Screen Shot 2020-09-09 at 9 32 52 AM](https://user-images.githubusercontent.com/22801026/92626903-74a59b00-f27f-11ea-9b68-b914ddac775f.png)

## Recommended QA:

**Back link**
1. find (or create) a team with mentors who have been invited, and mentors who have requested to join
2. click through to the mentor profile of both, and make sure both have the desired link back to the dashboard

**Approve/decline controls**
1. find (or create) a team with 4 mentors requesting to join it
2. approve one from the dashboard
3. approve one from within the mentor profile,
4. decline one from the dashboard
5. decline one from within the mentor profile
6. for all of the above, make sure you were taken to the right page with the right message, and the mentor is/isn't on the team as appropriate

**Additional testing**
1. Maybe try searching for a mentor already on your team and viewing their profile, ensure it doesn't error out and displays correctly (e.g. doesn't show approve/decline controls any more)
2. Same for a mentor not on your team that hasn't been invited or requested to join

hellafitz:
Screenshots are from https://projects.invisionapp.com/freehand/document/1hk5y3kgu

hellafitz:
Open Q: 
- what should happen when the buttons are clicked? (e.g. show confirmation and take you back to My Team page.

rsgonzal:
A:
- pop up confrming
- green banner to say they were approved
- goes to mentor section of team pages
![Screen Shot 2020-09-09 at 12 28 19 PM](https://user-images.githubusercontent.com/22801026/92644385-1b963100-f298-11ea-9dfe-d2bf470fc75d.png)
![Screen Shot 2020-09-09 at 12 28 08 PM](https://user-images.githubusercontent.com/22801026/92644390-1df88b00-f298-11ea-9145-ed4c1a5d3371.png)

[...]

PR #2647, opened 2020-09-10
#2642 mentor deets

stenington:
This makes the changes requested in #2642 related to mentor invites and mentor join requests.


shaunxp20 (approved):
Looks like this will work! 👍

Issue #2641, opened 2020-09-09
Redirect page to team page after you ask to join a team [5,5,1]


rsgonzal:
After a students asks to join a team, they are taken to this page and it displays a small message at the top that they have sent a request.

![Screen Shot 2020-09-04 at 10 57 07 AM](https://user-images.githubusercontent.com/22801026/92625842-03b1b380-f27e-11ea-810b-cce08775fb9e.png)

Instead, they should be sent to this page where they can see their request
![Screen Shot 2020-09-04 at 10 58 03 AM](https://user-images.githubusercontent.com/22801026/92625862-0a402b00-f27e-11ea-9829-d96714846ed4.png)

## Suggested QA
1. sign in a student not on a team
2. request to join a team
3. confirm the redirect

(est. 5 min)

stenington:
I opened PR #2652 to explore this a little.

hellafitz:
@stenington what do you see as our next step here?

stenington:
Input from @shaunxp20 on the PR, I've asked a question there.

ophelieh:
Works as expected! :) Logged in as student Stella Noir, I asked to join Jellicle Cats and was redirected.

rsgonzal:
perfect!

kate-technovation:
1. Signed in as a student not on a team:
<img width="762" alt="Screen Shot 2020-10-02 at 6 24 55 PM" src="https://user-images.githubusercontent.com/57409606/94974438-94298f80-04dc-11eb-8fbc-3579d8607ec8.png">

2. Requested to join a team:
<img width="768" alt="Screen Shot 2020-10-02 at 6 25 27 PM" src="https://user-images.githubusercontent.com/57409606/94974460-a60b3280-04dc-11eb-8dd6-a564c7091a54.png">

3. Confirm the redirect! Hooray!
<img width="767" alt="Screen Shot 2020-10-02 at 6 25 42 PM" src="https://user-images.githubusercontent.com/57409606/94974471-af949a80-04dc-11eb-91c4-9e6da642c3d7.png">

PR #2652, opened 2020-09-10
#2641 redirect to find-team path within dashboard app

stenington:
This would solve #2641 I think, although something about it makes me uneasy. It feels like a potentially brittle coupling between the rails side and vue side of things. But I'm not sure I see a better way to achieve the goals of the ticket given the way things work.


stenington:
@shaunxp20 What does your gut tell you about this potential change? Smells fishy, or seems fine?

stenington:
FWIW, when I come back and look at this, it seems realistic to do it this way. Named route helpers are nice, but you still have to consider bookmarks or external links or whatever anyway.

hellafitz:
Shaun: seems fine to proceed

shaunxp20 (approved):

Issue #2486, opened 2020-04-23
Remove admin requests section [8,8,1]


stenington:
In the admin, there is a Requests section that appears to be unused. We should remove it.
<img width="960" alt="Screen Shot 2020-04-23 at 5 31 00 PM" src="https://user-images.githubusercontent.com/55319/80151535-39ccf000-8588-11ea-80f4-e492380ac0bf.png">

## QA Recommendations

1. log in as admin
2. confirm that the **Requests** menu link is gone

Note: I didn't find anywhere that chapter ambassadors could interact requests, although it did look like a feature for ambassadors to request regions.

hellafitz:
@stenington mind labeling this and the other new tickets you added in the icebox? Effort estimates are helpful too. We should have capacity for a number of them.

nickiredescent:
Can confirm that there is no more "Request" menu accessible via the admin dashboard.

ophelieh:
Logged in as admin, I also confirm that "Request" is gone from the menu.

kate-technovation:
Logged in as Admin. Can't see a Requests section anymore:
<img width="838" alt="Screen Shot 2020-09-18 at 1 40 54 PM" src="https://user-images.githubusercontent.com/57409606/93628631-d919d600-f9b4-11ea-8ee6-7a21061d5d96.png">

mdancho:
Confirming there is no request section on the admin menu.

PR #2654, opened 2020-09-14
#2486 Removing requests

stenington:
Removing the admin request section, and backing code.


shaunxp20 (approved):
Lots of red, looks good to me! 😁 👍

PR #2655, opened 2020-09-14
#2620 add back the legacy 'name or email' field for first iteration of this

stenington:
This adds back the original "Name or email" search field, so now the change is only additive. People who prefer the old-style search have it, and the more specific exact-match search fields can go to production for some feedback on real data.


shaunxp20 (approved):
LGTM 👍

Issue #2349, opened 2020-01-13
Change admin Content & Settings data initialization [8,20,0.4]


stenington:
Related to #2345, there are other character combos besides a single quote that will break the content and settings page. We could continue on the path of #2346 and try to robustly encode or escape all potentially hazardous characters, but my proposed solution is for the `AdminContentSettings` component to initialize with an AJAX call instead of from a Vue prop. 

## Recommended QA

1. log in as an admin
2. visit **Content & Settings** and make sure the page loads
3. change some settings and save
4. make sure your settings changes stick
5. try out some weird characters in dashboard notices: single quotes `'`, double quotes `"`, backslash then quote `\"`. Reference #2345 for ideas on character combos that used to break this page
6. save and confirm that the page isn't broken!

stenington:
While the data is being loaded:
<img width="1280" alt="Screen Shot 2020-09-16 at 3 13 02 PM" src="https://user-images.githubusercontent.com/55319/93381826-3c79fb80-f82f-11ea-99ca-460d0667f370.png">

If there's an error loading the data:
<img width="1280" alt="Screen Shot 2020-09-16 at 3 13 30 PM" src="https://user-images.githubusercontent.com/55319/93381825-3be16500-f82f-11ea-8bf0-5b937df0a0c4.png">

rsgonzal:
Looks great!

PR #2658, opened 2020-09-15
#2349 safer admin content & settings data initialization

stenington:
This does away with the content & settings app initializing from html attributes, which was the root of #2345 from way back when.


shaunxp20 (approved):
No blockers, looks like this'll work, approved!

Issue #2616, opened 2020-08-27
Add help text on Judging in Content & Settings [5,5,1]


rsgonzal:
Currently, the team does not understand the implications of the different options when judging is open.

Suggestion is to add help text for a little more clarity. I think the suggestions from the user docs would be helpful: https://docs.google.com/document/d/1bFN0yf2M0RrpdxBaq4tudNfhcvGetifY3v1gnHvkV20/edit?usp=sharing

![Screen Shot 2020-08-27 at 9 36 59 AM](https://user-images.githubusercontent.com/22801026/91470172-394fa900-e849-11ea-954e-16ea51d59250.png)

![Screen Shot 2020-09-09 at 1 18 21 PM](https://user-images.githubusercontent.com/22801026/92648897-08d32a80-f29f-11ea-93d7-f824597252f3.png)

## QA STEPS

View the new help text tooltips (for the judging options):
* verify the help text
* verify the document link/link-text

rsgonzal:
Let's also link the reference doc somewhere on this screen

shaunxp20:
Here are a couple of samples/mock-ups for this:

# Inline help text
<img width="1076" alt="Screen Shot 2020-09-14 at 2 35 43 PM" src="https://user-images.githubusercontent.com/58957909/93140732-784b7e80-f6d2-11ea-9fa4-b0075eb36a67.png">

-----

-----

# Help text below in a table
<img width="1076" alt="Screen Shot 2020-09-14 at 3 13 15 PM" src="https://user-images.githubusercontent.com/58957909/93140755-826d7d00-f6d2-11ea-9a63-8b6511809903.png">

-----

-----

# Help icons with tooltips
<img width="1076" alt="Screen Shot 2020-09-14 at 3 34 32 PM" src="https://user-images.githubusercontent.com/58957909/93140792-931df300-f6d2-11ea-91dc-db2b51adb612.png">


Not shown here, but there will also be a link to the doc for more information.

rsgonzal:
Ouh! Thanks for the options, shaun! I kind of like the third one. @nickiredescent what do you think will be most helpful?

nickiredescent:
I also like the third one. Looks good!

shaunxp20:
Cool, thanks for the feedback @rsgonzal @nickiredescent, I'll work on the third one today! 😄 👍 🌞

shaunxp20:
This one is on QA, basic QA steps have been added as well. 😄 😁

[...]

PR #2659, opened 2020-09-15
#2616 Add help text to judging options in admin

shaunxp20:
This will add help text via tooltips to the judging options in the Content & Settings area in the admin.

#2616


stenington:
The project already has v-tooltip as a dependency. Can that be used instead of tippy?

shaunxp20:
> The project already has v-tooltip as a dependency. Can that be used instead of tippy?

Good catch! I just pushed up a new commit that uses v-tooltip. 👍

stenington (approved):
Works for me!

PR #2672, opened 2020-09-17
Open the judging document in a new window

shaunxp20:
This will open the judging document/documentation in a new window.

#2616


stenington (approved):
👍

Issue #2643, opened 2020-09-09
Change text on button when searching for new team or mentor [13,3,4.34]


rsgonzal:
Currently, there are two steps to requesting to join a team or mentor

The first button says to send a request, but it is actually sending you to a page to view more details

Changes:
- Change text from "Connect with.." or "Ask to join.." to "View more details"

![Screen Shot 2020-09-09 at 9 42 40 AM](https://user-images.githubusercontent.com/22801026/92628165-40cb7500-f281-11ea-9b02-0b14b24f6b45.png)
![Screen Shot 2020-09-09 at 9 42 29 AM](https://user-images.githubusercontent.com/22801026/92628167-41640b80-f281-11ea-80fb-83d5e79cca68.png)

# QA Steps:

* Confirm button text in the locations shown in the screenshot

hellafitz:
This change should apply to both Student and Mentor user types. More context on the overall user flows is at https://projects.invisionapp.com/freehand/document/1hk5y3kgu

rsgonzal:
looks great!

stenington:
@rsgonzal Looks like a scenario was missed here in the initial ticket. Students searching for mentors see "Ask X to mentor your team" on the search results page. I believe students searching for teams and mentors searching for teams and mentors have the updated language.

stenington:
## QA recommendation for round 2

Log in as a student, check the button text when searching for a mentor.

If you want to be very thorough...
1. log in as a student
2. confirm button text when searching for a team
3. confirm button text when searching for a mentor
4. log in as a mentor
5. confirm button text when searching for teams
6. confirm button text when searching for other mentors

rsgonzal:
thanks for fixing, mike! looks good to me.

PR #2682, opened 2020-09-21
#2643 One more button text change

stenington:
Looks like we had handled students searching for teams, and mentors searching for other mentor or teams in the previous PR for #2643, but had missed the students searching for mentors scenario.


shaunxp20 (approved):
LGTM! Sorry for taking so long to approve, I was AFK then meetings. 😊

Issue #2222, opened 2019-09-23
Make age calculation cutoff configurable


stenington:
When registering, we asks users their birth date and then show them their current age and the age they will be during World Pitch, which is part of the official division calculation. This cutoff date is [hard-coded](https://github.com/Iridescent-CM/technovation-app/blob/22b6f221487f02fe890932d94690b40b6ed8afa2/app/javascript/registration/store/state.js#L16) into the frontend, but it would be nice if it were configurable. 

I believe we would have two options for sending that cutoff from backend config to the frontend: either create an API endpoint for it, or include it in the DOM-based data-passing seen [here](https://github.com/Iridescent-CM/technovation-app/blob/cb40810d898a72d3cc41291a2741b5570f94c193/app/views/application/_signup_wizard.en.html.erb#L6) for example. I think since we've started down the latter path we may need to stick with it, despite the extra complexity it introduces.

stenington:
This comes from #2199

stenington:
Also, passing data is easier than I thought with the `dotenv` js library. See [here](https://github.com/Iridescent-CM/technovation-app/blob/qa/app/javascript/utilities/utilities.js#L4) for example.

stenington:
After looking more closely, the situation is maybe a bit muddier than I thought. There's some logic around divisions and cutoffs that gets repeated between the front end and back end code, and ideally there would be one source of truth, I think. 

The naive solution here would be taking the date currently hard coded on the front end and making it another piece of date configuration on the back end. That way, updating that front end piece each year is a config change, not a code deployment. The big caveat there is that that config would only control the front end date, unless we also reworked the `Division` model on the back end to draw from it. But `Division` currently has it's own way of updating the cutoff year that I'm not sure I totally understand, so I'm not certain what impact that would have.

hellafitz:
@stenington I'm open to doing this one and will need a recommendation on approach - either through technical refinement or a spike. What's your preferred next step?

stenington:
Oh, the approach to #2689 actually addresses this, so it will get closed by that.

PR #2695, opened 2020-09-23
#2689 update division age cutoff to pull from config

### LINKED ISSUE #2689: Update age calculation on sign up and edit text [20,3,6.67]
Sign up page is calculating for 2020, needs to be updated for World Summit 2021

Change text to --> "Your age during World Summit"

Screen Shot 2020-09-23 at 9 43 45 AM

## QA Recommendation

1. start signing up
2. confirm that the age part shows the correct 2021 date

stenington:
This PR could close #2689 and #2222. It makes the division cutoff date configurable in the environment, and both the front end and back end pull from that to determine junior/senior division.

It does away with `Division.cutoff_year` which was doing a strange date comparison that I never fully understood. I think pulling a date from config that we update yearly better maps to how we currently do things. Down the road if we introduced full-fledged seasonality, I think this would be a piece of config on each individual season, which isn't too much a leap from where this PR would take us.


shaunxp20 (approved):
👍

Issue #2582, opened 2020-07-16
Re-set the data terms agreement on season rollover  [40,5,8]


hellafitz:
Make data terms agreement re-collectable for Students, Mentors, Judges, RAs.

Designing Season Rollover doc https://docs.google.com/document/d/1oBVG_RMi6SVQ8KlznqmdRT_FuNCgRCkBy8maM4tXeWA/edit

Open Qs:
- What's the trigger for re-collecting (automatic or manual)?
- Do we need a record of past consents (and what those terms were?)
- Separate controls by user type or some other criteria (e.g. agreements more than 2 years old)?

## Potential QA 

1. log in as a student, mentor, or judge
2. confirm that data use terms have been agreed to in the past
3. run the reset with the force option (or change the rollover day and run)
4. log in again/refresh, see that you're interrupted with date use terms agreement screen

hellafitz:
- 1) R: ideal: any new user gets it, then manual trigger for existing ones; every year is good standard
- 2) confer with Amy
- 3) nice-to-have factors: people in europe, adults vs students

hellafitz:
Baseline: we make it look like no one has ever consented to data terms before, the platform will then require everyone to complete terms.
Next steps: 
- Confer with Amy bc the record of past consents has potentially the most implications on our database structure (right now all we do is mark a timestamp of when they consented; new timestamps overwrite the old one; if we need history that's a significant change)
- Do we want agreement history? 
- Do we want admin controls? If yes, what level of detail should they have / what's the interface for triggering agreement (re)collection?

hellafitz:
@amyiridescent fyi we'll ask you about this on Friday

hellafitz:
decision: no admin control needed

stenington:
My recollection of the conversation with @amyiridescent was that 
* for a given season, we want the timestamp of when they agreed
* for past seasons, consulting backups is fine for the time being
* we don't need admin controls (as @hellafitz noted above)

So that means we have the technical capabilities to satisfy the requirements already if we just blank out `terms_agreed_at` when the season rolls over. We can decide to automate that, or have a developer do it by hand.

hellafitz:
Deliverables: 
- add wiping out terms_agreed_at to the season rollover task
- ensure that timestamps are still recorded when a user accepts the terms
- confirm whether or not we've documented what's involved in the season rollover task. If we haven't create a ticket to do that. If we have, link it here and add the data terms wipe to the doc.

[...]

PR #2698, opened 2020-09-24
#2582 update the reset_for_season task to blank out data use terms agreements

stenington:
`reset_for_season` will now blank out data use terms agreements when it runs. I also added a `rails reset_for_season[force]` option to force it to run on a day that isn't the season rollover day, for easier testing.

This also removes `Account.populate_terms_agreed_from_signup_attempt` which as best I can tell was a band-aid we needed at some point, but isn't needed now.


shaunxp20 (approved):
Looks good, I like the `force` option addition. 👍

Issue #2660, opened 2020-09-17
Allow students and mentors to cancel their team join requests [20,20,1]


rsgonzal:
As a student, I want to be able to cancel my request to join a team so that I can move on and find another team.

Problem:
- Girls would write to us if they requested to join a team, and that team did not accept their invitation; the girls requesting to join could not ask to join other teams until the original team requested, rejected them

![Screen Shot 2020-09-16 at 5 24 21 PM](https://user-images.githubusercontent.com/22801026/93405331-879d0a00-f841-11ea-9e6a-5447f09c06a8.png)

## Suggested QA

Things to test:
1. you can cancel your request as a student
2. you can cancel your request as a mentor
3. as a team member already on the team, if you follow the link you get emailed **after** the requestor has cancelled, you see the "Sorry, but..." page
4. the "Manage my team" button works on that page as a student
5. the "Manage my team" button works on that page as a mentor

Other things to maybe think about: @rsgonzal 
6. I haven't looked at what this looks like when team building gets turned off? Do we need any special accommodation there?
- Send a request and turn off team building
- Login as the student that sent the request and go to find my team (ideally would be told they can't participate anymore)
- Login as a student that received a request. Try to accept the request from the email (request should be expired or should not be able to add to team anymore)

7. Does join requests show up somewhere else I'm not thinking of? Can you track them in the admin or anything? Does that all still work appropriately?
- After doing 1-6 above, login as admin and go to the team page and review the join requests section - did everything get recorded?

rsgonzal:
Proposed solution:
1. Standard red button: "Cancel my request to join this team"
2. Standard pop up: "Are you sure you want to cancel this request?"
![Screen Shot 2020-09-16 at 5 54 00 PM](https://user-images.githubusercontent.com/22801026/93407566-60493b80-f847-11ea-994e-7a87ee9cceb5.png)
![Screen Shot 2020-09-16 at 5 54 13 PM](https://user-images.githubusercontent.com/22801026/93407586-69d2a380-f847-11ea-9949-36d0c5b69996.png)

@ophelieh @kate-technovation how does this look?

ophelieh:
It looks great. Thanks @rsgonzal!

rsgonzal:
- if they cancel, can they re-request from the same team (YES)
- if they cancel, what does the team see if they click on outdated request (Need to check email flow)

stenington:
[...]

Issue #2662, opened 2020-09-17
Bug: Student sends request to join a team but it doesn't stick


rsgonzal:
**Describe the bug**
Student requesting to join a team is told they sent the request, but when they go to "find your team", there is no request there. 

**To Reproduce**
1. Go to participants
2. search for students, name: north dakota
3. Login as student
4. Go to "find your team" --> find team
5. Try to join "newteam"
6. go back to "find your team"
7. See the request isn't sent

**Expected behavior**
The "find your team" should list "new team" under ask to join team section

**Screenshots**
![Screen Shot 2020-09-16 at 6 14 04 PM](https://user-images.githubusercontent.com/22801026/93408646-ecf4f900-f849-11ea-8740-3e06bec0b811.png)

![Screen Shot 2020-09-16 at 6 25 07 PM](https://user-images.githubusercontent.com/22801026/93408724-20378800-f84a-11ea-8f68-d4024f9ed748.png)

**Additional context**
It's possible they've denied my request already or I removed myself from the team previously. Probably just a staging issues, but since we know these requests stall out (#2607) might be worth checking out
This page tells me when others have denied my request, but no info for "newteam"
![Screen Shot 2020-09-16 at 6 29 21 PM](https://user-images.githubusercontent.com/22801026/93408918-92a86800-f84a-11ea-9b39-3e72206158ca.png)

rsgonzal:
@shaunxp20 this might be one you can explore

shaunxp20:
> Additional context
> It's possible they've denied my request already or I removed myself from the team previously.

In the admin I see that the "North Dakota" join request was deleted.

<img width="1067" alt="Screen Shot 2020-09-16 at 10 43 09 PM" src="https://user-images.githubusercontent.com/58957909/93421565-37ee2b00-f8a1-11ea-83fe-71a19545f984.png">

Locally, I hard-deleted this request and I was able to successfully make another join request. I can do this on QA too, if you want. 😄 

It seems as if a student can only have one join request per team, even if it has been deleted. 😕 😞 🤕 

We can definitely address this, we could talk about it with the dev team and/or program team to come up with a solution. Some potential solutions: 
* display an error message (instead of misleading a student into thinking another join request has been made)
* send another join request
* don't allow students to request to join teams where the join request has been deleted
* something else

rsgonzal:
If a request gets deleted, maybe it should not count towards that one request.

Need to investigate:
- Why and how do requests get deleted.

[...]

PR #2708, opened 2020-09-30
#2660 Add ability for users to cancel their team join requests

stenington:
For students and mentors to be able to cancel join requests.

Right now, the implementation is that the join requests gets `deleted!`, which sets a `deleted_at` timestamp. The system also uses that in certain contexts, and I'm not explicitly capturing that the user cancelled vs. one of the other contexts, because the ticket didn't require that.

### DIFF
--- a/app/controllers/student/join_requests_controller.rb
+++ b/app/controllers/student/join_requests_controller.rb
@@ -39,7 +39,11 @@ module Student
         redirect_to student_team_path(team),
           alert: t("views.team_member_invites.show.invites_disabled_by_judging")
       elsif
-        current_student.join_requests.find_or_create_by!(team: team)
+        if join_request = current_student.join_requests.find_by(team: team)
+          join_request.pending!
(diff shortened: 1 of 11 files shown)


app/controllers/student/join_requests_controller.rb
```diff
@@ -39,7 +39,11 @@ def create
         redirect_to student_team_path(team),
           alert: t("views.team_member_invites.show.invites_disabled_by_judging")
       elsif
-        current_student.join_requests.find_or_create_by!(team: team)
```
stenington, line 42: This change would change the behavior seen in #2662, but I don't see a resolution there of what should actually happen, so this is potentially wrong.

shaunxp20 (approved):
The description still says WIP, but it looks good to me. Regarding #2662, I think this is a fine way to address it, better than what we have now, imho, and we can always add another ticket to address it differently.

Issue #2694, opened 2020-09-23
Replace mentor Expertise with Skills & Interests [13,?,?]


hellafitz:
from #2670, change the expertise section to read Skills & Interests, and offer the following options in place of the current ones:
Coding
Experience with Java
Experience with Swift
Project Management
Business or entrepreneurship
Marketing
Design

note: replacing the current 'expertise' options may automatically display the new options as profile labels in mentor search results and profile views. That is expected and good. 

Open Q for @bagomer: do these options need to appear in any particular order or need to be randomized?

Out of scope but next step/s or Qs:
- what does this look like on the mentor profile? 
- how does a mentor edit this info (including for returning mentors who may have no responses here)?
- what shows on initial search results (vs the full profile)?

bagomer:
I think we are ok with the order as presented

shaunxp20:
This one is on QA, and some QA steps have been added as well. 😄 🌞 😀

mdancho:
Registered as a mentor and confirm the check-box header says Skills & interests in the registration process. It does not show experience with java and swift. 

<img width="978" alt="Screen Shot 2020-10-07 at 12 15 31 PM" src="https://user-images.githubusercontent.com/52216992/95378118-52ac3200-0898-11eb-8cc1-2af4538be511.png">

The new options do not show when returning to your basic profile tab on the mentor dashboard after registering.  

<img width="857" alt="Screen Shot 2020-10-07 at 12 24 14 PM" src="https://user-images.githubusercontent.com/52216992/95378126-5475f580-0898-11eb-9fa4-2be6688c4887.png">

shaunxp20:
> Registered as a mentor and confirm the check-box header says Skills & interests in the registration process. It does not show experience with java and swift.

Doh! Sorry about that Mackenzie! I forgot to run the script to add the new ones! 🤦 😊 I just ran it, so they should be there now. 😅 😀

mdancho:
Thanks Shaun! Just checked and can confirm that the changes have been made on the registration page, and profile page.

shaunxp20:
> Thanks Shaun! Just checked and can confirm that the changes have been made on the registration page, and profile page.

Awsome, thanks for testing this one! 🙌 🍰 🎉

rsgonzal:
only thing i see is a double tag on this user

![]()

rsgonzal:
[...]

PR #2711, opened 2020-09-30
#2694 Update Mentor Skills & Interests

shaunxp20:
Renaming Mentor Expertises to Skills & Interests; and removing/adding some, and ordering them.


stenington:
Can you fix [the bootstrap task](https://github.com/Iridescent-CM/technovation-app/blob/qa/lib/tasks/bootstrap.rake) to set up the correct ones as well?

shaunxp20:
> Can you fix [the bootstrap task](https://github.com/Iridescent-CM/technovation-app/blob/qa/lib/tasks/bootstrap.rake) to set up the correct ones as well?

Good catch, I just pushed up a commit to fix these as well. 👍

stenington (changes requested):
If I drop my database and use `bin/setup` to rebuild, I get the right expertise options but without any order set on them.

shaunxp20:
> If I drop my database and use `bin/setup` to rebuild, I get the right expertise options but without any order set on them.

I think this one is ready for another round, I removed the bootstrap task, so now all the Skills & Interests/Expertises will get created via migrations. I think it's right, I didn't drop my database to verify though.

stenington:
Well, it turns out this doesn't work. It looks like `bin/setup` runs `rails db:setup`, which in turn creates the database from `db/schema.rb` or `db/structure.sql`, but does **not** run all migrations. I didn't realize that's how `bin/setup` worked.

Digging a little deeper, [this article](https://codingitwrong.com/2016/07/29/how-to-seed-and-migrate-rails-data.html) (which in turn references [this Thoughtbot article](https://thoughtbot.com/blog/data-migrations-in-rails)) make a case for keeping data changes out of db migrations, and I see their point in the long run. 

I think using `db/seeds.rb` to set up development and test databases should work, and then whether you want to keep the production expertise set up in a migration or not is up to you for now. Going forward maybe we shouldn't rely on that pattern but if this needs to get out the door, I don't think there's anything wrong with doing it that way in the short term.

shaunxp20:
Good digging on this one! A lot of that makes sense to me.

Incorporating some of the ideas in those articles, and somewhat thinking out loud here for my next steps:
* only add one migration to add an `order` column (changing schema only)
* update (remove/add/order) expertises in a new rake task (which we'd need to run manually on QA/prod, and afterwards I can add a PR to remove this)
* update `db/seeds.rb` with the default expertises/ordering

Thoughts on this? 🤔

stenington:
I think that sounds great!

shaunxp20:
>* only add one migration to add an `order` column (changing schema only) 
>* update (remove/add/order) expertises in a new rake task (which we'd need to run manually on QA/prod, and afterwards I can add a PR to remove this)
>* update `db/seeds.rb` with the default expertises/ordering

Okay, I made these changes, it's ready for another pass when you get a moment. Also, I tested this w/ a QA database (by migrating, and running the new rake task), and also w/ a new database (via db:drop db:setup), and both are working as expected for me.

stenington:
Cool, I can confirm this sets up my dev expertises table correctly.

stenington (approved):

Issue #2656, opened 2020-09-15
Add users to configured segments on mailing list subscription [20,13,1.5]


hellafitz:
(Rebecca) With the auto Mailchimp list signups by account, would it be harder to add signups to a somewhat different place in Mailchimp - eg one single list but different segments?
- Maggie has specific ideas. Talk to her.
- Can we route people to different segments in one listserv? If not, we still need new lists for 2021

****

I met w/ Maggie today, here are the new requirements:
* Add all accounts to one Mailchimp list (Technovation Girls 2021 Season Participants)
* Add a tag (student, parent, mentor, or judge) to each account, which then can be used as a segment
* Clean up old *_LIST_ID configuration

****

## QA STEPS

* Sign up as a new student, mentor, and judge
* Verify each new account gets added to this list in Mailchimp: "Staging Technovation Girls 2021 Season Participants", and gets tagged appropriately: student gets a student tag, mentor gets a mentor tag, and a judge gets a judge tag
* For parents, sign a parental consent form, and they should end up on the same email list with a parent tag

**Note:** there's a new config setting for the master list: `MAILCHIMP_LIST_ID`, for QA it's set to the list "Staging Technovation Girls 2021 Season Participants", for production, we'll set it to the real one

stenington:
In general, Mailchimp has a very detailed and complete API so the questions are probably:

1. What exactly does the team want?
2. What parts should be configurable? Hard-coded?
3. Does the Mailchimp API support what's wanted? (Probably yes)
3. What will it take to update our code to use the appropriate portion of the Mailchimp API? 

[This](https://github.com/Iridescent-CM/technovation-app/blob/qa/app/technovation/mailchimp.rb) is the API wrapper that would probably need updating.

rsgonzal:
@mjaris you were identified as having specific ideas, wondering if you have more info for the first 2 Qs?

mjaris:
Hi @rsgonzal, Rebecca and I talked about having people be automatically routed into segments of one master 2021 list in mailchimp, instead of being added to separate lists.

**2021 Technovation Girls Season Participants**

- Students (segment that's automatically populated during registration process)
- Mentors (segment that's automatically populated during registration process)
- Parents (segment that's automatically populated during registration process)
- Judges (segment that's automatically populated during registration process)
- Ambassadors (segment that's manually updated/maintained by program team, beyond scope of this ticket)

stenington:
[...]

PR #2726, opened 2020-10-07
#2656 Add tags to Mailchimp API call

shaunxp20:
Just doing the simplest thing here (actually backed out a few changes to make it minimal) to get this working. I'll add another ticket to do some refactoring.

**Notes:**
There's a new ENV/config setting that needs to be added locally: `MAILCHIMP_LIST_ID`.
There's a test list setup in Mailchimp, "Staging Technovation Girls 2021 Season Participants" w/ the ID: `49795c1e0e`, if you wanted to test this locally.

#2656


stenington (approved):
I'm fine with this, with one question first. Did you get the strong sense that they're not going to want to split users out into different lists again? If that commitment was firm, this is good. If not, I don't see any harm in keeping the list IDs configured by account type, and just putting the same list ID into each for the time being.

Just food for thought. Feel free to merge this if you feel the requirement for one list was firm.

shaunxp20:
> Just food for thought. Feel free to merge this if you feel the requirement for one list was firm.

Cool, thanks for the feedback! I'll merge it in as is, one list requirement was firm, and if we need multiple lists again it'll be pretty simple to revert this to get the old functionality back. 👍

Issue #2601, opened 2020-08-07
Update TG emails for 20/21 season [20,20,1]


rsgonzal:
Emails need updating. First step is to gather copy from Program Team

## QA Recommendation

1. check each email in the [referenced doc](https://docs.google.com/document/d/1Fg5pw-ROi3ngKPm_Z-CmUbozXS1DLWX5uZ12IyIfb4k/edit#) for the updates requested
2. also invite a student **who already has an account** to your team, and look up a mentor and invite them to your team
3. confirm that both of those emails are still correct (they shared a template with the invitation for students without accounts)

rsgonzal:
Rough list of emails/copy: https://docs.google.com/document/d/1vWOEY7CA3O4ojCULyajA2_OXEqnUGmLzXhcbAQbwi6g/edit#heading=h.d9r2f2l02xjl

rsgonzal:
- Double emails send out to student after parent signs consent form

hellafitz:
Decision: need final email copy by Monday 9/7 EOD. Rebecca is point on PT. 
- @rsgonzal to doublecheck the list by pulling the emails again

hellafitz:
Email copy has been provided. @rsgonzal to organize it for easier implementation. Would be available by next week.

rsgonzal:
organized changes: https://docs.google.com/document/d/1Fg5pw-ROi3ngKPm_Z-CmUbozXS1DLWX5uZ12IyIfb4k/edit#

stenington:
Reiterating a comment from within the doc: we're planning to update the `HELP_EMAIL` setting as part of this, which will change things beyond just emails, but that's desired.

stenington:
We will also want to update `MENTOR_SLACK_SIGNUP_URL` as part of this.

stenington:
I have updated `HELP_EMAIL` and `MENTOR_SLACK_SIGNUP_URL` on staging only.

stenington:
@rsgonzal Can I change the help email and mentor slack on production now? That doesn't need to happen at the time of season rollover does it?

rsgonzal:
Yes! You can change it now. Thanks!

stenington:
Ok, these have been updated in production as well.

ksrebecca:
QA - looked at, left comments in the doc, confirmed the different emails!

ophelieh:
I am adding edits to the document provided in the QA description. 

I tested as:
- a new student
[...]

PR #2727, opened 2020-10-07
#2601 editing parental consent instructions if user is having problems

stenington:
Hopefully final changes requested from QA.


shaunxp20 (approved):
👍

Issue #2716, opened 2020-10-02
Rebrand: Replace World Pitch with World Summit all over TG [8,3,2.67]


rsgonzal:
We are still using 'World Pitch' in various places on the platform. Let's try to do a direct replacement with 'World Summit'

## QA Suggestion

We were saying "World Pitch" in the following locations:

* During the age verification step of registration
* If attempting to sign up as someone younger than 10, the error message shown on the profile selection part of registration
* The email sent declining a chapter ambassador (not sure if this gets sent any more, since we invite them)
* Twice in the parental consent email

hellafitz:
Estimated 5 places where this occurs in the codebase

rsgonzal:
This looks good. we don't decline ChAs anymore.

I didn't see the language in parent consent though

rsgonzal:
@stenington which email is it exactly?

rsgonzal:
actually i can check!

rsgonzal:
ah it's the actual consent form. Can confirm the changes

rsgonzal:
wow, the pdf gets updated too! magic

PR #2745, opened 2020-10-09
#2716 Update world pitch to world summit where it occurs

stenington:
Just a handful of occurrences needed changing from what I saw.


shaunxp20 (approved):
👍

Issue #2741, opened 2020-10-08
Bug: Consent link in Mentor Email says invalid token when clicked [8,8,1]


rsgonzal:
Number 6 in mentor welcome email takes you to a broken page whether logged in or not
![Screen Shot 2020-10-08 at 10 57 13 AM](https://user-images.githubusercontent.com/22801026/95496062-25718980-0955-11eb-92d0-43fba0a66e5d.png)

![Screen Shot 2020-10-08 at 10 26 34 AM](https://user-images.githubusercontent.com/22801026/95493567-78e1d880-0951-11eb-9aba-99f1dadab773.png)

![Screen Shot 2020-10-08 at 10 27 30 AM](https://user-images.githubusercontent.com/22801026/95493562-77181500-0951-11eb-9b11-49439bd3ed99.png)

link: `https://mandrillapp.com/track/click/30787762/technovation-qa.herokuapp.com?p=eyJzIjoiTW9nWV84QVJsb0hINVd1ZGlqb3RCWldXWmZZIiwidiI6MSwicCI6IntcInVcIjozMDc4Nzc2MixcInZcIjoxLFwidXJsXCI6XCJodHRwczpcXFwvXFxcL3RlY2hub3ZhdGlvbi1xYS5oZXJva3VhcHAuY29tXFxcL2NvbnNlbnRfd2FpdmVyc1xcXC9uZXc_bWFpbGVyX3Rva2VuPVBjVEZLQ3BpNEQ1djQxdUx4cEFmV3h6b1wiLFwiaWRcIjpcIjE1ODk0NjFjYzAyOTQ5ZmM5OWZjNTU3MTRmNDUzZjg2XCIsXCJ1cmxfaWRzXCI6W1wiODUyYjZkYmU5ZjhmZjAyMDI0N2RmYTZlM2UxZWFkMmY3OTE4NTg4NVwiXX0ifQ`

## QA suggestion

**Works when signed in**
1. sign up as a mentor
2. receive the welcome email
3. making sure you're **logged in** as the mentor, click the consent form link
4. sign the consent form

**Works when signed out**
1. sign up as a mentor
2. receive the welcome email
3. making sure to **log out** first, click the consent form link
4. make sure you're taken to the consent form, and automatically logged in
5. sign the consent form

(~10 min)

hellafitz:
Steps to Reproduce:
- sign up as a new mentor
- get a mentor welcome email 
- click step #6

[...]

PR #2746, opened 2020-10-09
#2741 fix signing consent waiver from welcome link

stenington:
There were two main problems with the mentor welcome email link to the consent waiver:
1. it didn't include the token with which to look up the consent, and
2. it included a mailer token that didn't actually work.

This should fix both things, such that following the welcome email link will log you in if necessary, and then take you to your consent that you can sign.


stenington:
This one was trickier than expected so I'm writing up an explanation of some pieces of it.

Firstly, consent waivers were backed by a generic consent waivers controller that knew nothing about profile types and knew nothing about authentication. This led to weird behavior like letting you see the consent waiver and the field to consent to it when you were logged out, and only complaining when you tried to actually sign and submit that form.

In order to put it behind the standard authentication checks (and incidentally to also allow `mailer_token` to auto-login when necessary) I made the old controller a concern, and then added mentor and judge controllers that inherit from that concern. That required some restructuring of the views as well to use `current_scope` and things like that to get user scoping right.

Then for testing, I added `capybara-email` as a dependency which lets you write specs against emails that have been sent pretty cleanly. The main change here though is that we have to specify that capybara launch the test server at a specific port, and make sure `HOST_DOMAIN` is configured accordingly, as well as setting `config.action_mailer.default_url_options` appropriately in the test environment. Otherwise the links in the email don't point at the test server and it all breaks. (Or it hits your dev server without you realizing and you get really confused for a good couple of hours.)

I also fixed a small bug in the participants data_grid that wasn't showing the status of judge consent waivers when it should be showing it.

That's what's going on in this PR.

shaunxp20:
Regarding this functionality:
> following the welcome email link will log you in if necessary

Sounds like it will automatically login a judge or mentor when clicking the the link in the email? If so, does the link expire after a certain period of time? And (I guess more important if the link doesn't expire) will the link/auto-login still work if the account's email address or password has changed? 🤔

stenington:
Good questions! The link automatically logs in a judge or mentor when they click, the link does not expire, and the auto-login still works even if they change email address or password.

The behavior is based on the pre-existing code around `mailer_token`s, and in fact the link was already including the mailer_token, but not in a way where it was working correctly. But I didn't add the auto-login functionality to the platform with this PR, just properly took advantage of an existing feature.

shaunxp20 (approved):
I think it's okay to move this forward to get the consent wavers working again, I also added a placeholder ticket so we don't forget about the security vulnerability: https://github.com/Iridescent-CM/technovation-app/issues/2759

Issue #2740, opened 2020-10-08
Send Mentors to mentor training after sign up [8,8,1]


rsgonzal:
Mentor Training is the first thing on the list, but mentors seem to be sent to the consent form first after signing up

![Screen Shot 2020-10-08 at 9 58 26 AM](https://user-images.githubusercontent.com/22801026/95490640-577eed80-094d-11eb-9237-8b048f9ceb6f.png)

## QA Suggestion

1. Sign up as a mentor
2. When you reach the dashboard, you should land on Mentor Training

Additional note:

@rsgonzal You're right that the dashboard remembers where you last were, and takes you back to that screen after the initial visit. One way you can specifically test where it will take you if it doesn't have a last location is to log in as a mentor from a new private browsing window. That window will not have the cookie containing your last location, so it will use the same logic that gets used when you have a brand new account. As you complete the Profile items, it should be taking you to the next unchecked item, although I think it would take you to the background check item before the personal summary item, the way it's currently coded.

hellafitz:
@bagomer what's your ideal order here?

mdancho:
I would have it go to the mentor training first. The training is also in the auto-emails as the first thing to complete.

rsgonzal:
yeah i signed up a few other times and it took me to consent. I'm having mackenzie try it now

rsgonzal:
mackenzie tried it and said the first thing to pop up was the training. So I think this is ok

PR #2749, opened 2020-10-12
#2740 Fix next onboarding step calculation for mentors

stenington:
The next onboarding step is sent across as part of the serializer, and hadn't been updated to account for a mentor profile change, I think. This should fix the problem, and adds tests for that part of the serializer.


shaunxp20 (approved):
Looks good 👍

Issue #2729, opened 2020-10-08
Add 'Personal Summary' to Mentor signup and include it for confirmation in Team Building [13,13,1]


hellafitz:
The Personal Summary (bio) appears in the Team Building section of a Mentor's workflow. 

We want to first introduce it during Mentor signup - between Skills & Interests and above 'How did you hear about Technovation?' The Personal Summary should still appear in Team Building but this time it should display any existing text response and allow that text to be edited. 

The Personal Summary should still appear as Bio on the Mentor Details page as accessed through search  results. Resolving the naming issue (personal summary vs bio) is out of scope here but deserves another ticket). Also important but out of scope, we should decide if the bio should be displayed in the abbreviated search results or only in profile details (upon click-through from initial search results display).

#### sign up as mentor
1. sign up as a mentor
2. you should not be able to complete sign up without providing 100+ characters of personal summary
3. make sure the personal summary has been saved when you finish profile creation

#### sign up as student
1. sign up as a student
2. it shouldn't require a personal summary
3. it should still work

(~10 mins)

rsgonzal:
![Screen Shot 2020-10-08 at 10 12 11 AM](https://user-images.githubusercontent.com/22801026/95491745-e6403a00-094e-11eb-9766-3f72eb43b395.png)

hellafitz:
Include "you can change this later" ![Screen Shot 2020-10-08 at 12 50 06 PM](https://user-images.githubusercontent.com/6475595/95506508-d6335500-0964-11eb-9fb7-35f0bf45c6a6.png)

screenshot for signup flow context:
![Screen Shot 2020-10-08 at 12 49 33 PM](https://user-images.githubusercontent.com/6475595/95506511-d6cbeb80-0964-11eb-8748-f8137ce16289.png)

stenington:
In order to do this one, we need to talk more about what "enforced on dash" means.

Right now, the mentor bio can either be blank, or it has to be more than 100 characters worth of text. All profile editing screens enforce this, and it holds together conceptually. If you put anything, you have to put the minimum required amount, and otherwise an empty bio is just ignored when it comes to validation. That is to say that editing some other part of your profile won't result in an error that your bio is too short, it understands that a completely empty bio is allowed.

[...]

PR #2753, opened 2020-10-16
#2729 collect bio from mentors during basic profile registration

stenington:
This PR gets the job done in terms of allowing mentors to fill out their personal summary up front, but I encountered a couple bits of code I'm a little unsure of along the way.


stenington:
I think this PR gets the job done, although there are a couple of things I don't like about it. I'll add comments at the parts where I think it's a bit sketchy.

app/javascript/registration/components/BasicProfile.vue
```diff
@@ -305,6 +330,16 @@ export default {
       },
     },
 
+    bio: {
+      get () {
+        return this.$store.state.registration.bio || ""
```
stenington, line 335: I added the `|| ""` here so that a length check done above wouldn't blow up, and I don't think the distinction between null and the empty string matters otherwise. But hopefully there's not some subtle bug lurking.

app/javascript/registration/components/BasicProfile.vue
```diff
@@ -355,6 +390,10 @@ export default {
       this.debouncedProfileUpdate({ expertiseIds: value })
     },
 
+    bio (value) {
```
stenington, line 393: I don't fully understand why each item has a corresponding entry in both `computed` and `watch`, that seemed odd to me but maybe I don't know enough about Vue.

app/javascript/registration/store/getters.js
```diff
@@ -55,7 +55,9 @@ export default {
                     state.schoolCompanyName &&
                       state.jobTitle &&
                         state.mentorType &&
-                          state.genderIdentity)
+                          state.genderIdentity &&
+                            state.bio &&
+                              state.bio.length >= 100)
```
stenington, line 60: I don't love duplicating validation requirements on the front end and the back end. I'm not sure how normal this is for Vue though? Might be completely normal.

app/javascript/registration/store/mutations.js
```diff
@@ -136,6 +137,10 @@ export default {
     state.mentorType = value
   },
 
+  bio (state, value) {
```
stenington, line 140: I felt like there was a lot of boilerplate code to this solution, not sure if that's just how Vue is or if it's a code smell.

app/models/student_profile.rb
```diff
@@ -186,6 +186,9 @@ def mentor_type=(*)
   def expertise_ids=(*)
     false
   end
+  def bio=(*)
```
stenington, line 189: Not a fan of this. It seems like student profiles explicitly ignore certain attributes so that you can pass in a superset of potential attributes that a profile might have without really knowing which profile you're dealing with. It works I guess, but smells fishy to me.

app/javascript/registration/components/BasicProfile.vue
shaunxp20: Not sure if Vue.js supports it, but you could also use `??` here to be more explicit about just covering `null` and `undefined`, instead of all of JavaScript's falsy values, just a fwiw comment.

app/javascript/registration/store/getters.js
shaunxp20: I think it's okay, it makes me think of a database constraint with a corresponding Rails validation, which can help save a trip to the database. I see this as validation that can save a trip to the server, seems okay to me. #fwiw

app/models/student_profile.rb
shaunxp20: Just thinking out loud, but using a decorator to make the interface to our *_profile duck types consistent might make this better. Nonetheless, good to know how this works currently. 👍

shaunxp20 (approved):
No blockers, looks like it'll get the job done.

Issue #2710, opened 2020-09-30
Pass additional info to MailChimp on signup [13,13,1]


hellafitz:
@ksrebecca inquired whether we could include birth year and country when importing student data to Mailchimp. This is useful, for example, when doing targeted email outreach to alumni (e.g. sharing an opp for girls 14 and under, or something just for girls in Kazakhstan)

Deliverables: 
Pass additional merge variables to MailChimp from the TG platform when a student signs up:
- Birth year (not the full birthdate which is PII)
- Country (PT acknowledges there will probably be messy data included, e.g. Spain vs ES)

## QA STEPS
* Sign up as a new student/mentor/judge, after signing up `BIRTHYEAR` and `COUNTRY` should be populated in Mailchimp
* As a parent, sign a parental consent form, `BIRTHYEAR` and `COUNTRY` shouldn't be filled out in Mailchimp (but `NAME` should be)

Also, would be good to test...
* Update a student/mentor/judge profile (name, DOB, country, email), and those changes should be reflected in Mailchimp

##### Add `ENABLE_MAILCHIMP` configuration setting
- [X] QA
- [x] Production

##### Remove `ENABLE_MAILING_LISTS` 
- [X] QA
- [x] Production

##### Add new merge fields - `BIRTHYEAR`, `COUNTRY`, `NAME`
- [X] QA
- [x] Production

hellafitz:
@bagomer do you need this info for mentors too?

bagomer:
Yes please, I can see Country being quite helpful; birth year might be helpful as well for fundraising etc.

stenington:
We can add country as a merge variable, but it looks like Mailchimp also understands locations in lat/long, so I think we might consider using that too.

hellafitz:
Updated Deliverables:
Pass additional merge variables to MailChimp from the TG platform when a student signs up:
- Birth year (not the full birthdate which is PII)
- Country (PT acknowledges there will probably be messy data included, e.g. Spain vs ES)
- Lat/Long [DISCUSS]

Pass additional merge variables to MailChimp from the TG platform when a mentor signs up:
- Country (PT acknowledges there will probably be messy data included, e.g. Spain vs ES)
- - Lat/Long [DISCUSS]

shaunxp20:
> We can add country as a merge variable, but it looks like Mailchimp also understands locations in lat/long, so I think we might consider using that too.

[...]

PR #2758, opened 2020-10-21
#2710 Add additional Mailchimp fields

shaunxp20:
A pretty big refactoring here, here are some of the highlights.

The interface to `subscribe` and  `update` in the Mailchimp service has some pretty big changes:
* to subscribe an account now, you just need to pass in the account (removing the need to pass in merge fields and location data), and an optional profile type that will be used as a tag
* to update an account now, you just need to pass in the account (instead of what would have been four arguments), and optionally `currently_subscribed_as` (open to renaming that), used if the email address has changed

A couple of other benefits from the refactoring, now merge fields (and location data) are in one spot (instead of being scattered around), and the Mailchimp error handling is in one spot (instead of being scattered around.

The jobs have also changed:
* we had jobs that called other jobs to subscribe/update an account, this been simplified now
* now there's just one job to subscribe an account, and one job to update an account
* I also created two new jobs to subscribe/update a parent (which lead to the new `FauxAccount` class (opening to renaming this too) since they don't have real accounts


stenington (approved):
Cool, this seems really good to me! 👍 I didn't like how the merge fields were spread throughout the codebase previously either, and jobs calling other jobs was a pain to trace through. This seems like a big improvement!

Issue #2747, opened 2020-10-12
Bug: students marked as onboarded that shouldn't be [13,5,2.6]


stenington:
**Describe the bug**
Our daily task to check for students who should be onboarded found 14701 records just before the season reset date. Looking at one of those records a couple of days later, they are marked as onboarded but shouldn't be. If I run the `check_onboarding` task however, it currently passes, I think because it is narrowing the scope to current users only (and there aren't many of those a day or two after the season rollover).

It's not clear to me if this problem will fix itself as girls return. Maybe something about getting registered to the new season will actually fix the underlying onboarding mismatch, I'm not sure. But regardless, I suspect that our `onboard_missed_students!` and `reset_for_season` tasks are in conflict on the season rollover day.

**To Reproduce**
Don't have steps to reproduce yet.

## QA

I think @shaunxp20 or I can QA this one. I would recommend setting up at least one of each type of account: 
* onboarded but shouldn't be
* not onboarded but should be

We can do that if necessary through the console, using `update_column` to update the `onboarded` attribute without firing hooks that would identify and fix the problem we want to temporarily introduce. 

Then we can run the new task by hand and check the output for the correct accounts, as well as confirming the `onboarded` attribute is correct on each user afterwards.

## DEPLOYMENT NOTES

- [x] After this code deploys, the scheduled job for onboard_missed_students! will need to be changed to run fix_student_onboarding! instead.

hellafitz:
- Check this ASAP: Do we have girls who are marked as fully onboarded to this season who shouldn't be?
- Investigate to get it more clearly refined
- Try to reproduce

stenington:
I believe the cached `.onboarded` attribute will be recalculated when a student logs in and gets registered to the new season, because `RegisterToCurrentSeasonJob` does a `profile.save` to fire commit hooks, and `onboarded` gets updated in an `after_commit`.

So even if a student's account is marked as onboarded for 2020, when they log in and get registered to 2021 that should be corrected.

That said, currently 2 accounts are marked onboarded but shouldn't be. So there's still some way for things to get out of sync. Maybe the daily check that marks students as onboarded if needed should work both ways, and mark students as not onboarded if appropriate?

stenington:
I followed the QA steps outlined and it looks correct to me. 

## Deployment note

After this code deploys, the scheduled job for `onboard_missed_students!` will need to be changed to run `fix_student_onboarding!` instead.

PR #2768, opened 2020-10-26
#2747 student onboarding discrepancies

stenington:
This PR adds a rake task that will fix both
* students marked as onboarded who shouldn't be, and
* students not marked as onboarded who should be

My proposal would be to merge this, then replace the `onboard_missed_students!` scheduled job with `fix_student_onboarding!`, and add another Papertrail alert to catch the "should not be onboarded" case and alert in Slack.

This would be a band-aid fix to let us get a sense of how often it happens, and to fix occurrences of the problem sooner rather than later.


shaunxp20 (approved):
Looks like another good one to have until we can get to the root of the problem, thanks for adding this!

Issue #2732, opened 2020-10-08
Bug: Double expertise tag on mentor [5,?,?]


rsgonzal:
From #2694 only thing i see is a double tag on this user

![Screen Shot 2020-10-07 at 5 10 23 PM](https://user-images.githubusercontent.com/22801026/95400767-581d7280-08c0-11eb-8c07-bf4091dc1b0e.png)

_Originally posted by @rsgonzal in https://github.com/Iridescent-CM/technovation-app/issue_comments/705257710_

rsgonzal:
From Shaun:

> Hmm, that's interesting. 🤔 😕

>I see the first Project Management entry added on/at: 2020-10-05 21:14:47.263362
And I see the second one added: 2020-10-05 21:14:48.413439

>And another weird thing. I also see Coding added: 2020-10-05 21:14:47.900735
And a blank Skill/Interest added: 2020-10-05 21:14:46.771004

>Seems like a bug (maybe related to double emails being sent out 🤔 ). I can try to find something in the code that could be causing it, repro steps would come in handy too if you can figure out the steps to re-create this. 😄 🤔 🐛 🥀

stenington:
It looks like in production there are 84 instances of the same expertise being assigned to the same profile more than once, with creation dates anywhere from 2018 through 2020. So it's not a new bug, however it's happening.

hellafitz:
- needs technical refinement (we want to figure out why this is happening so we understand the Interests tags better
- needs repro steps

stenington:
I don't have steps to reproduce, but I've traced through on a couple of the instances of this problem and it doesn't look like it's necessarily happening on sign up. For one user, their signup attempt does **not** have the duplicate expertise, and from edit dates I would assume the mentor edited their profile later on and ended up with the duplicates somehow. But in general, editing your profile doesn't create duplicates as far as I can tell.

My gut tells me it's some sort of race condition when maybe a person has a bad internet connection and tries submitting a change multiple times? I'm not sure.

But even so, I do think I have a fix that will ensure users only have one of any given expertise: PR #2771

PR #2771, opened 2020-10-26
#2732 prevent creation of duplicate mentor expertise

stenington:
I believe this PR both prevents the further creation of duplicate mentor expertise, and where we already have them it hides that duplication without interfering with the mentor's profile.


shaunxp20 (approved):
Looks like a nice stopgap until we can get to the root of the problem. 👍

PR #2778, opened 2020-10-29
#2710 Update Mailchimp list when DOB changes

shaunxp20:
Simple fix to update the Mailchimp list when DOB changes.


stenington (approved):
Makes sense to me!

Issue #2780, opened 2020-10-29
Remove technovation-staging [x,2,x]


shaunxp20:
From #2776, we want to remove `technovation-staging`.

- [ ] Remove references in the codebase
- [x] ~~Backup code/database~~ (Nothing to backup)
- [x] Remove Heroku app

PR #2781, opened 2020-10-30
#2780 Remove technovation-staging

shaunxp20:
Removing references to technovation-staging.


stenington (approved):
Yep, looks like this should be fine.

Issue #2737, opened 2020-10-08
Edit Confirmation message a parent sees after signing consent form [5,3,1.6]


rsgonzal:
Add a second sentence to green banner that says --> You will receive a confirmation email with a copy of the consent you signed. 
![Screen Shot 2020-10-08 at 9 17 47 AM](https://user-images.githubusercontent.com/22801026/95486013-56e35880-0947-11eb-9392-932d0a48b48c.png)

## QA

1. sign up as student or find student without parental consent
2. sign the consent form
3. confirm additional sentence on success message

(~10 mins)

rsgonzal:
Confirmation message looks good!
![Screen Shot 2020-10-28 at 12 22 44 PM](https://user-images.githubusercontent.com/22801026/97486312-81a64880-1918-11eb-89e7-2d18c4176f29.png)

ophelieh:
Not directly related, but it seems that the "Onboarded students" - "No, still onboarding" filter is not working properly as it showed me Francine Sinatra who's fully onboarded. Same for Diana Garcia. Also, that filter shows students who are now mentors (Mentor McTeacher). 
![Screen Shot 2020-10-29 at 1 03 50 PM](https://user-images.githubusercontent.com/32206421/97613857-327b1900-19e7-11eb-97fd-302a16a4e2f0.png)

Another note, as the parent for student Sydney White, I received the parental consent form twice. 

The confirmation message looks good although the text "STUDENT NAME has been consented by parent" is a little weird. Could we rephrase to "STUDENT NAME received parental consent on DATE."? 


<img width="1440" alt="Screen Shot 2020-10-29 at 1 08 14 PM" src="https://user-images.githubusercontent.com/32206421/97614331-d06ee380-19e7-11eb-8a83-2e2365cbaee3.png">

rsgonzal:
> Not directly related, but it seems that the "Onboarded students" - "No, still onboarding" filter is not working properly as it showed me Francine Sinatra who's fully onboarded. Same for Diana Garcia. Also, that filter shows students who are now mentors (Mentor McTeacher).

Yeah, we don't know why it's happening! But it's happening a lot in staging because of the data reset we did. Should not be as frequest on production. the quick fix is to just update the parent email, not even changing the address - just hitting the update button


> Another note, as the parent for student Sydney White, I received the parental consent form twice.

don't know why double emails are happening either. doesn't seem to be consistent

[...]

PR #2786, opened 2020-11-05
#2737 change text of parental consent message

stenington:
Change requested in QA.


stenington:
This is just a text change, so moving it through without code review.

Issue #2697, opened 2020-09-24
Make search visibility options more prominent for teams [20->40,20,1->2]


hellafitz:
The checkboxes for a team or a mentor to allow themselves to appear in search results are hard to find and easily misinterpreted. Make them easier to find and more visible as shown below.

![Screen Shot 2020-10-08 at 12 36 40 AM](https://user-images.githubusercontent.com/6475595/95428990-9d609500-08fe-11eb-9d4b-3aface3ac7e6.png)

(#2731 is the mentor version of this ticket, which is much less far along)

## Suggested QA

The changes here should just be rearranging the page, not changing functionality. To that end, there are two different avenues for testing: 
1. make sure the rearranged elements are correct in a variety of situations, and
2. make sure the necessary page functionality hasn't broken

For the first, I would suggest looking at the page in a number of different situations, like
* team has one member, no invites or join requests
* team has multiple members
* team has pending members
* team has pending invites
* team has pending requests
* various scenarios above, with team building off
* various scenarios above, with the first round of judging started

And then make sure page controls still work, like
* removing a member
* deleting an invite
* inviting a user
* approving a join request
* denying a join request

hellafitz:
NOTE: Updated screens are in profile fields invision board

- what do admins and ChAs see? A: base it on this updated content but preserve current additional functionality (e.g. links to student profiles only appear for them, not for students)
- do we want parts to disappear when team-building turns off (as it currently does). A: yes, follow current pattern and feel free to break off the 'do you want more teammates' section
- also powers a view of past teams for admins and ChAs. A: follow current pattern and check in if Qs

QA note: 
- do 'good enough' QA in the absence of detailed mockups
- create new tickets for feedback after 1 round

rsgonzal:
#2666 could be used for the change to include info on how many team members are allowed

stenington:
I have an invision board where I'm trying to understand the current patterns of behaviour based on different page "scenarios" (data or config situations where something changes on the page): https://projects.invisionapp.com/freehand/document/OyNBmSwOk

The bottom row is the current site, the top row is what I'm building. 

## Current site
* Simplest page, one member 
    * this is the baseline
* One of each thing (member, pending member, pending invitation, pending join request)
    * pending student section appears
    * invites sent section appears
    * requests from others section appears
* Five+ things
    * invite box and searchability checkbox are replaced with "5 students" message
* Team building off
    * invite box and searchability checkbox are replaced with "team building off" message
[...]

PR #2787, opened 2020-11-05
#2697 team students page QA changes

stenington:
This addresses three items from QA:
* fixing punctuation
* making avatars smaller (note this will adjust both the student and mentor tabs, but I assume that's desirable)
* removing the max 5 message from the top


shaunxp20 (approved):
👍

PR #2907, opened 2021-04-18
Revert "Allow projectPage urls in team share url validator"

Aboisier:
After talking with Amy Kim, it seems like this feature is not desirable, as there is not way to lock in a state when sharing with the projectPage url. The teams would be able to continue working on the app after submission. This PR reverts it.


shaunxp20:
We already have this one merged in/reverted, I'm going to close this PR, thanks again! 😄

PR #2908, opened 2021-04-19
conditional logic to display judge helper text for thunkable projects

viviancan:


shaunxp20 (approved):
Awesome! Looks good, I didn't manually test it, but it seems pretty straight forward! 👍

PR #2910, opened 2021-04-20
Change logic for mentor certificate to match students

tamtran-agilityio:


app/technovation/determine_certificates.rb
```diff
@@ -94,7 +98,15 @@ def gets_mentor_appreciation_certificate?
 
   def needed_mentor_appreciation_recipients
     @account.mentor_profile.teams.by_season(season).select { |team|
-      !@account.appreciation_certificates.by_season(season).for_team(team).any?
+      # A Mentor should only receive a certificate for a team they are on if that team:
+      # Has a submission
+      # The submission is at least 50% complete
```
shaunxp20, line 103: These comments aren't necessary, an alternative to having comments here would be to document these things in the test file, it would be good to write some new tests for this functionality too! 😄 The test file is here: `spec/technovation/determine_certificates_spec.rb`

app/technovation/determine_certificates.rb
```diff
@@ -94,7 +98,15 @@ def gets_mentor_appreciation_certificate?
...
+      team_submissions = team.team_submissions.filter { |submission|
```
shaunxp20, line 104: I'm pretty sure we want to use `submission` here and not `team_submissions`, from what I'm seeing the `submission` association will return the current season's submission vs `team_submissions` which looks like it will return all submissions (I think we only want to award a certificate based on the current season). 🤔 😃

app/technovation/determine_certificates.rb
```diff
@@ -94,7 +98,15 @@ def gets_mentor_appreciation_certificate?
...
+        submission.status == SUBMISSION_STATUS[:compete] ||
```
shaunxp20, line 105: Good catch on checking for complete here! I think we can just use `submission.complete?` though instead of using this constant. 😄

app/technovation/determine_certificates.rb
tamtran-agilityio: @shaunxp20 I have update all, you can help me check it, thanks!

spec/technovation/determine_certificates_spec.rb
```diff
@@ -65,18 +65,20 @@
 
   context "for mentor" do
     it "awards mentor appreciation" do
```
shaunxp20, line 67: For these specs, it would be nice to capture the new logic, maybe something along the lines of:

```
context "when the team's submission is more than 50% complete" do
    it "awards a mentor appreciation certificate" do
```

```
context "when the team's submission is less 50% complete" do
    it "does not award a mentor appreciation certificate" do
```

app/technovation/determine_certificates.rb
```diff
@@ -94,7 +94,8 @@ def gets_mentor_appreciation_certificate?
 
   def needed_mentor_appreciation_recipients
     @account.mentor_profile.teams.by_season(season).select { |team|
-      !@account.appreciation_certificates.by_season(season).for_team(team).any?
+      !@account.appreciation_certificates.by_season(season).for_team(team).any? && 
+      team.submission.present? && (team.submission.complete? || team.submission.qualifies_for_participation?)
```
shaunxp20, line 98: Looking at this again, unless I'm missing something, I think this would make more sense as `team.submission.percent_complete > 50`, it would be simpler and more explicit (and it would line up better w/ the new specs). 🤔 😄

spec/technovation/determine_certificates_spec.rb
rsgonzal: @tamtran-agilityio

app/technovation/determine_certificates.rb
rsgonzal: @tamtran-agilityio

PR #2913, opened 2021-04-20
Load correct submission when a judge has multiple submissions

shaunxp20:


app/technovation/find_eligible_submission_id.rb
```diff
@@ -12,7 +12,7 @@ def call(judge_profile, options = {})
       else
 
         id_for_finished_score(judge_profile, options) ||
-          id_for_score_in_progress(judge_profile) ||
+          id_for_score_in_progress(judge_profile, options) ||
```
shaunxp20, line 15: This is basically the fix - `options` has the `score_id` which `id_for_score_in_progress` needs in order to find that score/submission. I also wrote a test for this (not the best/most readable imho, but I was just following the pattern already there). 😃

viviancan:
looks good to me. Ran it locally and the "flicker" behavior was happening again, but we will revisit that. Besides that, the correct project loaded in!

PR #2929, opened 2021-04-22
Rake task to update Spain regions

shaunxp20:
I definitely need a second pair of eyes on this one! Updating production data makes me feel 😬 😬 😬  

Using the list here:
https://docs.google.com/spreadsheets/d/16qmnGy6aBrJyeLjF-mmoAr0eNU-CnwGtLwuk2BkJONo/edit

I created a (run once, then I'll delete it) rake task.

The abbreviations for the regions I used in the rake task are from here:
https://en.wikipedia.org/wiki/ISO_3166-2:ES

Things to check please:
* Setup a copy of production data locally
* Then, in the admin, under Participants, search for "Spain", and one of the regions in the spreadsheet
  * You should see that some of the accounts in the spreadsheet are missing
* Run the rake task - `bundle exec rake update_spain_location_data`
* Now all of the accounts in the spreadsheet _**should**_ be showing up for their corresponding region


viviancan:
I ran the rake task locally against my local copy of prod db.
Andalucia went from 25 -> 41
Madrid 266 -> 793
Valencia 74 -> 245

Do those numbers seem accurate?

shaunxp20:
> I ran the rake task locally against my local copy of prod db.
> Andalucia went from 25 -> 41
> Madrid 266 -> 793
> Valencia 74 -> 245
> 
> Do those numbers seem accurate?

Thank you for looking at this! 😃 

Yes! Those numbers look good to me!

According to the spreadsheet:
Andalucía: 37
Madrid: 794
Valencia: 245

For Andalucía, I think a few more is fine (new accounts could have been created since the spreadsheet was created). For Madrid, there's actually one account on the spreadsheet that doesn't have an account on our platform (row 301), so I think this number is good, and Valencia looks good too! 👍

Issue #2923, opened 2021-04-21
Mentors who already have judge profile need to show up in judge segment on mailchimp [8,13, .62]


rsgonzal:
1. get list of mentors who already have judge profile and add them to the judge segment

split from #2912

shaunxp20:
Just fyi, I added the script to update existing accounts to #2912.

Issue #2912, opened 2021-04-20
Add mentor/judge combo account to mailchimp judge segment [13, 8, 1.63]


rsgonzal:
Currently, only new judges are being added to the mailchimp mailing list.

Mentors can toggle to become judges. If they toggle to become judges, they should also be added to the judge mailing list.

Because we are using segments this year, I'm not sure if they can be in two segments. If they can't, we need to check if we can remove them from the mentor segment.

## QA STEPS/NOTES
* To QA, create a new mentor, then login as this mentor and click the "Switch to Judge Mode" button
* In Mailchimp verify this mentor shows up in both the mentor's list and the judge's list (tagged w/ both "mentor" and "judge")

**Notes:**
* Only when a mentor clicks the "Switch to Judge Mode" button for the **first** time will a judge profile will be created, which will also add them to the judge's email list (by tagging them as "judge"); also, they will **only** get added to the judge's email list if they are already in Mailchimp on the mentor's email list (i.e., this will not subscribe them if they're not already in Mailchimp)
* If you impersonate a mentor and click the "Switch to Judge Mode" button, it will **not** create the judge profile or add them to the judge's email list - there is some code that prevents this from happening when impersonating, I'm not sure about the history of why that's in there, but that's the way it currently works: https://github.com/Iridescent-CM/technovation-app/blob/4ec0e881624a8ac969d4787167f2a2a8d5e37a98/app/controllers/judge/dashboards_controller.rb#L18-L19

The accounts in the following dataclip will be to added to the judge email list, but only if they're already in Mailchimp:
https://data.heroku.com/dataclips/oqlfzluvmrvcbtfvgaxcgjosgzno

## DEPLOYMENT NOTES

After this has been deployed, run the command:
`$ heroku run bundle exec rake add_existing_mentor_judge_accounts_to_judge_email_list  --app technovation`

it should run for about 1.8k accounts (close to the number in the dataclip).

rsgonzal:
1. add mentors who create judge profile to judge segment
2. get list of mentors who already have judge profile and add them to the judge segment

PR #2933, opened 2021-04-27
Add mentor/judge accounts to judge email list

shaunxp20:
When a mentor becomes a judge (via the "Switch to Judge Mode" button), this will add them to the "judge" email list (by tagging them as a judge). The account will only get tagged as a judge when the judge profile is first created (when the button is clicked for the first time), and only if they are currently on the Mailchimp list.

I also added the script to add existing mentor/judge accounts to the judge email list

Tickets: #2912 and #2923

### DIFF
--- a/app/technovation/create_judge_profile.rb
+++ b/app/technovation/create_judge_profile.rb
@@ -1,28 +1,31 @@
 module CreateJudgeProfile
   def self.call(account)
-    if account.can_be_a_judge? and account.is_not_a_judge?
+    if account.can_be_a_judge? && account.is_not_a_judge?
       create_judge_profile(account)
+      AddProfileTypeToAccountOnEmailListJob.perform_later(profile_type: "judge", account_id: account.id)
+
       true
     else
       false
     end
   end
 
-  private
   def self.create_judge_profile(account)
     attrs = if account.chapter_ambassador_profile.present?
-              {
-                company_name: account.chapter_ambassador_profile
-                  .organization_company_name,
-                job_title: account.chapter_ambassador_profile.job_title,
-              }
-            elsif account.mentor_profile.present?
-              {
-                company_name: account.mentor_profile.school_company_name,
-                job_title: account.mentor_profile.job_title,
-              }
-            end
+      {
+        company_name: account.chapter_ambassador_profile
+          .organization_company_name,
+        job_title: account.chapter_ambassador_profile.job_title
+      }
+    elsif account.mentor_profile.present?
+      {
+        company_name: account.mentor_profile.school_company_name,
+        job_title: account.mentor_profile.job_title
+      }
+    end
 
     account.create_judge_profile!(attrs)
   end
+
+  private_class_method :create_judge_profile
 end
--- a/app/technovation/mailchimp.rb
+++ b/app/technovation/mailchimp.rb
@@ -23,6 +23,14 @@ module Mailchimp
       end
     end
 
+    def add_profile_type_to_account(profile_type:, account:)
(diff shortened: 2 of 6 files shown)


app/technovation/create_judge_profile.rb
```diff
@@ -1,28 +1,31 @@
...
       create_judge_profile(account)
+      AddProfileTypeToAccountOnEmailListJob.perform_later(profile_type: "judge", account_id: account.id)
+
       true
     else
       false
     end
   end
 
-  private
```
shaunxp20, line 11: Th changes below were b/c the StandardRB linter was complaining (about indentation, and "IneffectiveAccessModifier"):
https://docs.rubocop.org/rubocop/cops_lint.html#lintineffectiveaccessmodifier

app/technovation/create_judge_profile.rb
```diff
@@ -1,28 +1,31 @@
 module CreateJudgeProfile
   def self.call(account)
-    if account.can_be_a_judge? and account.is_not_a_judge?
+    if account.can_be_a_judge? && account.is_not_a_judge?
       create_judge_profile(account)
+      AddProfileTypeToAccountOnEmailListJob.perform_later(profile_type: "judge", account_id: account.id)
```
shaunxp20, line 5: When the "Switch to Judge Mode" button is clicked for the first time, it will create the judge's profile, and then I added this new bit for the email list stuff.

And just an fyi, if you wanted to test this locally, you need to change `perform_later` to `perform_now`, at least that's what I had to do. I'm not sure how the local jobs queue is hooked up, but I wonder if it's queuing it up on QA/staging. 🤷

app/technovation/mailchimp.rb
```diff
@@ -23,6 +23,14 @@ def subscribe(account:, profile_type: "")
       end
     end
 
+    def add_profile_type_to_account(profile_type:, account:)
+      handle("add profile type #{profile_type} to #{account.email} on list: #{list_id}") do
+        list.members(subscriber_hash(account.email)).tags.create(
```
shaunxp20, line 28: Mailchimp API doc on adding tags (this is the only way I found to do it):
https://mailchimp.com/developer/marketing/api/list-member-tags/add-or-remove-member-tags/

And our Gibbon wrapper for the Mailchimp API on adding tags:
https://github.com/amro/gibbon#tags

lib/tasks/add_existing_mentor_judge_accounts_to_judge_email_list.rake
```diff
@@ -0,0 +1,14 @@
+desc "Add existing mentor-judge combo accounts to the judge's email list"
+task add_existing_mentor_judge_accounts_to_judge_email_list: :environment do
+  account_ids = Account
+    .joins(:mentor_profile)
+    .joins(:judge_profile)
+    .where("'2021' = ANY(seasons)")
```
shaunxp20, line 6: The SQL this generates is pretty close the the dataclip (the only difference I'm seeing is that it just pulls non-deleted accounts):

> SELECT "accounts".* FROM "accounts" INNER JOIN "mentor_profiles" ON "mentor_profiles"."account_id" = "accounts"."id" INNER JOIN "judge_profiles" ON "judge_profiles"."deleted_at" IS NULL AND "judge_profiles"."account_id" = "accounts"."id" WHERE "accounts"."deleted_at" IS NULL AND ('2021' = ANY(seasons))

To check this, I just used `.to_sql`:
`Account.joins(:mentor_profile).joins(:judge_profile).where("'2021' = ANY(seasons)").to_sql`

viviancan (approved):

Issue #2921, opened 2021-04-21
Add judge recusal info to score details page [5,5,1]


rsgonzal:
![Screen Shot 2021-04-21 at 10.17.37 AM.png](https://images.zenhubusercontent.com/60527169cde2772242d0de95/0d12b623-06b0-4218-b79f-ae4a944b8567)

viviancan:
QA Steps
1. Go to scores on right side navbar 
2. Scroll all the way right to the last column > "View" column --> Select the first icon (hamburger icon) > read score details
3. Upon viewing the individual score page, you will see a "Recusal" section near the bottom of the page

**No Recusals Example**
 ![no-recusal-info.png]
(https://images.zenhubusercontent.com/606b0f4fd13d53dff3d14fa5/0a38a5ce-47bc-467d-8409-147a76aa31ec)

**Recusal Example**
![recusal-info.png](https://images.zenhubusercontent.com/606b0f4fd13d53dff3d14fa5/a580df9a-2a04-4e47-ba20-ad572a44a193)

ophelieh:
Works as expected! The only comment I would make is that if you are on the judge profile and they selected "other" - you cannot see the reason. This note is for future seasons.

mdancho:
Works for me! 
<img width="829" alt="Screen Shot 2021-05-04 at 9 17 40 AM" src="https://user-images.githubusercontent.com/52216992/117035900-b6290180-acb9-11eb-970d-cb86784d2d2f.png">

rsgonzal:
@ophelieh you actually just need to hover over the 'other' text and the reason will pop up! I can't seem to capture it in a screenshot for you though

ophelieh:
Oh awesome. Thanks @rsgonzal !

PR #2934, opened 2021-04-27
#2921 judge recusal information on score detail page

viviancan:
I kept team name and submission name as part of the table for now, to make QA easier to verify team name. I will take that out before final push to prod!


app/views/admin/participants/_score_details_judge_recused_scores.erb
```diff
@@ -0,0 +1,42 @@
...
+  <table class="width--100-percent">
+    <thead>
+    <tr>
+      <th>Team Name</th>
+      <th>Submission Name</th>
+      <th>Judge</th>
+      <th>Reason</th>
+      <th>Round</th>
+    </tr>
+    </thead>
```
shaunxp20, line 11: Seems like the `tr` should be indented, or is that typically not indented? Anywho, 😎 at using `thead` and `tbody`. 👍 😄

shaunxp20 (approved):
Looks good, straight forward to me! (Didn't manually QA it). Way to go! 🙌 🥳 🎆

app/views/admin/participants/_score_details_judge_recused_scores.erb
viviancan: fixed!

PR #2936, opened 2021-04-28
Remove recused scores from incomplete scores

shaunxp20:


app/models/submission_score.rb
```diff
@@ -152,7 +152,7 @@ class SubmissionScore < ActiveRecord::Base
   }
 
   scope :complete, -> { where("submission_scores.completed_at IS NOT NULL") }
-  scope :incomplete, -> { where("submission_scores.completed_at IS NULL") }
+  scope :incomplete, -> { where("submission_scores.completed_at IS NULL AND submission_scores.judge_recusal = FALSE") }
```
shaunxp20, line 155: I think this is all we need to remove recusals from incomplete scores. 🤔 The rest of the changes in this diff are for the "pending_*" counter caches, which I did a preliminary search for, and can't figure out where were even using them. 😕

viviancan:
looks good to me!

PR #2946, opened 2021-05-14
Opening video link in new tab starts a new submission

tamtran-agilityio:


viviancan:
Checked locally, and worked as expected! will merge to QA once checks have passed. Thanks @tamtran-agilityio !

PR #3040, opened 2021-09-03
Media Consent Form

shaunxp20:
This adds basic functionality for media consent forms - being able to sign one, and view a signed one.

### DIFF
--- a/app/controllers/media_consents_controller.rb
+++ b/app/controllers/media_consents_controller.rb
@@ -1,0 +1,50 @@
+class MediaConsentsController < ApplicationController
+  def show
+    @media_consent = find_media_consent
+
+    if @media_consent.blank?
+      redirect_to root_path, alert: t("controllers.media_consents.invalid")
+    end
+  end
+
+  def edit
+    @media_consent = find_media_consent
+
+    if @media_consent.blank?
+      redirect_to root_path, alert: t("controllers.media_consents.invalid")
+    elsif @media_consent.signed?
+      redirect_to action: :show, token: params.fetch(:token)
+    end
+  end
+
+  def update
+    @media_consent = find_media_consent
+
+    if @media_consent.update(media_consent_params.merge(signed_at: Time.current))
+      redirect_to(
+        {
+          action: :show,
+          token: params.fetch(:token)
+        },
+        success: t("controllers.media_consents.update.success")
+      )
+    else
+      render :edit
+    end
+  end
+
+  private
+
+  def find_media_consent
+    Account.find_by(consent_token: params.fetch(:token))
+      &.student_profile
+      &.media_consent
+  end
+
+  def media_consent_params
+    params.require(:media_consent).permit(
+      :electronic_signature,
+      :consent_provided
+    )
+  end
+end
--- a/app/jobs/register_to_current_season_job.rb
+++ b/app/jobs/register_to_current_season_job.rb
@@ -66,6 +66,8 @@ class RegisterToCurrentSeasonJob < ActiveJob::Base
       ParentMailer.consent_notice(profile.id).deliver_later
     end
 
+    profile.media_consents.create(season: Season.current.year)
+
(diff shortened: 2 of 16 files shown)


app/jobs/register_to_current_season_job.rb
```diff
@@ -66,6 +66,8 @@ def prepare_student_for_current_season(record)
       ParentMailer.consent_notice(profile.id).deliver_later
     end
 
+    profile.media_consents.create(season: Season.current.year)
```
shaunxp20, line 69: A new media consent form will be created when a student is registered to the current season.

app/controllers/media_consents_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+      &.media_consent
+  end
+
+  def media_consent_params
+    params.require(:media_consent).permit(
+      :electronic_signature,
+      :consent_provided
+    )
+  end
+end
```
shaunxp20, line 50: Btw, the code in here is functionally similar to what's in the  [ParentalConsentsController](https://github.com/Iridescent-CM/technovation-app/blob/c8bd0fba14640133e8f6d6218f78f446d7a873ab/app/controllers/parental_consents_controller.rb). (I just tweaked it a little and tried to clean it up a bit.)

An interesting piece is that there is a `consent_token` column on the `Accounts` table, we use it in the `ParentalConsentsController` and here to find the `account` and then ultimately the parental or media consent form.

Screen Shot 2021-09-03 at 3 49 57 PM

In this screenshot 👆 the token in the URL is the `consent_token` on the `Accounts` table.

app/views/parent_mailer/consent_notice.en.html.erb
```diff
@@ -27,21 +27,27 @@
...
       <tr>
         <td class="content-block" itemprop="handler" itemscope itemtype="http://schema.org/HttpActionHandler">
-          <%= link_to "Sign the Permission Form", @url, class: "btn-primary", itemprop: "url" %>
+          <%= link_to "Sign the Permission Form", @parental_consent_url, class: "btn-primary", itemprop: "url" %>
+        </td>
+      </tr>
+
+      <tr>
+        <td class="content-block" itemprop="handler" itemscope itemtype="http://schema.org/HttpActionHandler">
+          <%= link_to "Sign the Media Consent Form", @media_consent_url, class: "btn-primary", itemprop: "url" %>
```
shaunxp20, line 50: This is the parental mailer that I tweaked just slightly to add a link to the media consent form. We'll update it properly in another ticket. 👍

app/models/student_profile.rb
```diff
@@ -96,6 +96,9 @@ class StudentProfile < ActiveRecord::Base
     class_name: "ParentalConsent",
     dependent: :destroy
 
+  has_one :media_consent, -> { where season: Season.current.year }
+  has_many :media_consents, dependent: :destroy
```
shaunxp20, line 100: `student_profile.media_consent` will return the media consent for the current season and `student_profile.media_consents` will return all media consents (for that student). I was thinking as we build in "seasonality" we can use this pattern for other models/associations, where we can get the X thing for the current season by using the singular version and if we want to get all of them, we can use the plural version. #RandomThoughts

viviancan: I like that idea! 👍

config/locales/controllers.en.yml
```diff
@@ -81,6 +81,11 @@ en:
           confirm: "Are you sure you want to cancel this request?"
           success: "You have cancelled your request to join %{name}"
 
+    media_consents:
```
viviancan, line 84: I had never seen this file before! Nice to know it exists!

viviancan (commented):
Nice work on all the pieces of functionality! Looks good to me. I did not run it locally.

shaunxp20:
Thanks for reviewing this! I'm going to get this one merged in!! 🤗 🙌 🎉

Issue #3043, opened 2021-09-08
Create beginners division in divisions table


rsgonzal:
- add to the table
- if there are tests, add

viviancan:
Divisions table has relationships with accounts, teams, regional_pitch_events, divisions_regional_pitch_events. I think division names are found in `division.rb` line 5 
`enum name: [:senior, :junior, :none_assigned_yet]`

shaunxp20:
This one is on QA, no real QA is necessary for it though. I also ran the rake task (from the deployment notes), and it ran successfully.

PR #3047, opened 2021-09-14
Add "beginners" to divisions table

shaunxp20:
I added a rake task to add the beginners division to the divisions table.

#3043

### DIFF
--- a/app/models/division.rb
+++ b/app/models/division.rb
@@ -1,8 +1,12 @@
 class Division < ActiveRecord::Base
   SENIOR_DIVISION_AGE = 15
 
-  # If these change, you will need to update dataclips
-  enum name: [:senior, :junior, :none_assigned_yet]
+  enum name: {
+    beginner: 3,
+    junior: 1,
+    senior: 0,
+    none_assigned_yet: 2
+  }
 
   validates :name,
     uniqueness: { case_sensitive: false },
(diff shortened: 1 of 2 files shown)


app/models/division.rb
```diff
@@ -1,8 +1,12 @@
...
   SENIOR_DIVISION_AGE = 15
 
-  # If these change, you will need to update dataclips
-  enum name: [:senior, :junior, :none_assigned_yet]
+  enum name: {
+    beginner: 3,
+    junior: 1,
+    senior: 0,
+    none_assigned_yet: 2
+  }
```
shaunxp20, line 9: I thought making this a hash would help make it clearer that the keys/labels are being mapped to those numbers in the database; instead of knowing/inferring, for example that "senior" at position `0` in an array is the value `0` in the database.

Division Name | Value in Database (in `name` column)
--- | ---
Beginner | 3
Junior | 1
Senior | 0
None Assigned Yet | 2

Also, from what I was reading, instead of storing integers in the database, it's better to store an enum in the database. I guess something to keep in mind if we want to refactor this or if we're creating new enums.

https://naturaily.com/blog/ruby-on-rails-enum

viviancan: Great idea! Definitely way more clear this way!

viviancan (commented):
Everything looks great here 👍

PR #3056, opened 2021-09-23
Updated dynamic links on recent activity view which was causing ChA dashboard to break

viviancan:


shaunxp20 (approved):
Awesome! Great work and sleuthing on this one! I remember seeing some similar changes that the contractors made (using `send` in some of the links in the controllers). Thanks for taking care of this and the great notes too! 🥳 🎉 🙌

viviancan:
I realized I missed one of the commits 😅

shaunxp20:
> I realized I missed one of the commits 😅

👍 Good catch! 🙌

Issue #3052, opened 2021-09-20
BUG: Chapter Ambassador onboarding is broken


rsgonzal:
Not sure what this related to (maybe rails?) but when I try to write my program description and finish my onboarding as a Cha, I am given an error page on not able to move forward or do anything anymore.

breaks when i hit save below
![screencapture-technovation-qa-herokuapp-chapter-ambassador-dashboard-2021-09-20-16_31_37.png](https://images.zenhubusercontent.com/60527169cde2772242d0de95/6f0410b6-4ffa-4274-b91a-7912bde67df3)![Screen Shot 2021-09-20 at 4.46.32 PM.png](https://images.zenhubusercontent.com/60527169cde2772242d0de95/ea5f356b-5a85-4f21-bb0b-291dfcf0098a)

viviancan:
Resources

- https://github.com/rails/rails/issues/42157
- https://api.rubyonrails.org/classes/ActionView/Helpers/UrlHelper.html#method-i-link_to
- https://github.com/chaps-io/public_activity#activity-views

Original Error from airbrake
![Screen Shot 2021-09-22 at 8.44.36 PM.png](https://images.zenhubusercontent.com/606b0f4fd13d53dff3d14fa5/7d96bc7f-d989-4f5d-9d68-0d167d270b12)

viviancan:
![PublicActivityNotes.png](https://images.zenhubusercontent.com/606b0f4fd13d53dff3d14fa5/faa407d0-75dd-4014-b5f6-9567b3f1790c)

viviancan:
PR https://github.com/Iridescent-CM/technovation-app/pull/3056

PR #3060, opened 2021-09-29
Updated incorrect paths for ChA profiles in the Recent Activity tab

viviancan:
This was a bug fix based on previous ticket #3052. I missed 2 of the paths which caused an error on production for some of the ChA logins. I also updated 2 other routes, to ensure the dynamic links are going to correct view.


shaunxp20 (approved):
Only thing popping out to me was changing `_participant_path` to `_team_path`. Which I'm sure is fine, so, thumbsing this one up! 🥳 👍 🙌 Thanks for taking care of this! 🎉 🎊

PR #3062, opened 2021-10-04
WIP: Completed front end UI and basic JS logic for the new registration student flow

viviancan:
Most of the additions are the different components for the new registration flow. There is basic logic to "submit" the form, but this PR does not contain the backend logic to actually create a student account. This also contains the new additions of tailwind.css and vue formulate (the library for the form). 

@shaunxp20 The only thing I was really unsure of was the correct location for the specific assets used for new-registration. I tried placing them in different directories. I could not access them from the new-registration pack until I placed them in the same directory, but it does not seem right! 

There will be an additional PR that adds in the mentor flow as well!


shaunxp20:
> @shaunxp20 The only thing I was really unsure of was the correct location for the specific assets used for new-registration. I tried placing them in different directories. I could not access them from the new-registration pack until I placed them in the same directory, but it does not seem right!

Probably something to do with our asset pipeline configuration - it probably auto-loads things in certain directories (and needs some additional config to load things in other directories). #JustAGuess  🤷 🤔 😄

shaunxp20 (approved):
Overall, looks good to me! Only thing popping out to me is the logging statements and some commented out code - maybe for these things we can capture that in another ticket (if we don't have that already). 🙌  🎉 :shipit: 😁 

And something to ponder on/talk about - maybe we can introduce a linter/formatter. 📌 🤔 😁

PR #3077, opened 2021-10-27
Updated the additional tabs of the "my profile" student dashboard with the new rebrand

viviancan:
Updated the additional tabs of the "my profile" student dashboard with the new rebrand. I also had to update a few of the test specs / HTML ids for the tests to pass.

Last Thursday, 10/21, I was stuck on one additional failing test, but was having trouble replicating it locally. After further investigation, I found that one of the student registration tests was failing because the location details form was not rendered in the view. The form is in Vue. This test kept failing because the form was never rendered on the new layout page. I think that this form will have to be updated/removed anyway with the new registration flow. For now, though, it should be included to meet the current flow and test specs!


shaunxp20 (approved):
👍 

A couple things popped (static date, jquery?, pending links, and other quibblers), but no blockers. I say let's get it merged in and address things as we go!

:shipit: 👍 🥳

PR #3080, opened 2021-10-31
Acount creation for new registration flow

shaunxp20:


app/technovation/profile_creating.rb
```diff
@@ -15,21 +15,6 @@ def self.execute(*args)
...
-
-    profile.account.update({
-      email_confirmed_at: Time.current,
-      icon_path: icon_path,
-      division: Division.for(profile.account),
-    })
-
     controller.remove_cookie(CookieNames::SIGNUP_TOKEN)
 
     AttachSignupAttemptJob.perform_later(profile.account_id)
```
shaunxp20, line 20: Some notes about the registration flows and this file `ProfileCreating`...

The current/main registration flow (front door signups):
`app/controllers/registration/accounts_controller.rb` (the registration flow ends w/ this file)

Side door signups:
`app/controllers/admin/signups_controller.rb` (actually not sure how/where this is used)
`app/controllers/chapter_ambassador/signups_controller.rb`
`app/controllers/mentor/signups_controller.rb`
`app/controllers/judge/signups_controller.rb`
`app/controllers/student/signups_controller.rb`

Each of one these flows creates an account and profile in the controller, and then calls this file `ProfileCreating`.

The new registration flow creates an account and profile in the controller but does not call `ProfileCreating`. There was more in this file that we don't need than need. So, I moved some things out of here so the parts we need will still work w/ the new registration flow.

At the bottom of this file we call `SignIn`, which will happen in the new registrations controller.

Basically, the new registration flow diverges from the existing flow by bypassing `ProfileCreating`, but then it gets back on the same track as the existing flow by calling `SignIn`.

app/models/division.rb
```diff
@@ -21,48 +21,62 @@ def self.junior
...
-    ImportantDates.division_cutoff
+  def self.for_account(account)
+    case account.age_by_cutoff
+    when 8..12
+      beginner
+    when 13..15
+      junior
+    else
+      senior
+    end
```
shaunxp20, line 53: This is what I was trying to describe the other day. Originally I made this case statement end with (which I think reads better):
```
when 16..18
  senior
else
  none_assigned_yet
end
```

But you'll see the original functionality (before the beginner division) returns either junior or senior:
`account.age_by_cutoff < SENIOR_DIVISION_AGE ? junior : senior`

For historical purposes, we can't return `none_assigned_yet` (even though that's a valid division type) b/c the tests fail. And I don't have time to go down that rabbit hole to find out why, so the easiest thing to do was to just make this return either beginner, junior, or senior; and leave `none_assigned_yet` out of this.

I did have a thought to make it like this:
```
when 16..18
  senior
else
  senior
end
```

But I'm not sure if that makes it clearer or not. 😕

app/models/division.rb
```diff
@@ -21,48 +21,62 @@ def self.junior
...
   def self.for(record)
-    if !!record
-      division_for(record)
-    else
-      none_assigned_yet
-    end
-  end
+    return none_assigned_yet if record.blank?
 
-  private
```
shaunxp20, line 36: A random note, this private statement that was removed, it wasn't working as intended, it doesn't make these class methods private.

The standardrb linter picked it up:
`private` does not make singleton methods private. Use `private_class_method` or `private` inside a `class << self` block instead.

app/models/division.rb
```diff
@@ -1,5 +1,5 @@
 class Division < ActiveRecord::Base
-  SENIOR_DIVISION_AGE = 15
+  SENIOR_DIVISION_AGE = 16
```
shaunxp20, line 2: This could be named better imho, it's really the `MINIMUM_AGE_FOR_SENIOR_DIVISION`. Anywho, I kept it in here b/c it's used in other parts of the app and in the tests. 😕

app/models/account.rb
```diff
@@ -505,6 +505,12 @@ def chapter_ambassador
     end
   }
 
+  before_create do
+    self.email_confirmed_at = Time.current
+    self.icon_path = ActionController::Base.helpers.asset_path("placeholders/avatars/#{rand(1..24)}.svg")
+    self.division = Division.for_account(self)
+  end
```
shaunxp20, line 512: This is some stuff I moved out of `ProfileCreating` (b/c the new registration flow isn't using/calling `PorfileCreating`) so that way both the old and new registration flows will have this functionality.

app/javascript/new_registration/components/FormWrapper.vue
```diff
@@ -98,11 +98,22 @@ export default {
...
+          case 'student':
+          case 'parent':
+            window.location.href = '/student/dashboard'
+          case 'mentor':
+            window.location.href = '/mentor/dashboard'
+        }
+      }
+      catch(error) {
+        if(error.response) {
+          console.log(error.response.data);
```
shaunxp20, line 114: Random note - this is where we'll need to hook up the errors to Vue Formulate. I noticed we'll have to make some tweaks though b/c the keys in the error messages need to correspond to the form field names. And right now the form field names are camel cased and what's returned from Rails would be snake case.

app/controllers/new_registration_controller.rb
```diff
@@ -1,11 +1,84 @@
...
+  def parent_params
+    {
+      parent_guardian_name: registration_params[:parentName],
+      parent_guardian_email: registration_params[:accountEmail],
+      school_name: registration_params[:schoolName],
+      account_attributes: account_attributes
+    }
+  end
+
+  def mentor_params
```
shaunxp20, line 53: Another random note, to hook up mentor registrations, we just need to add the mentor params to `registration_params` which is a few lines below, and also add them here. And that should be it (for the happy/success path).

app/controllers/new_registration_controller.rb
```diff
@@ -1,11 +1,84 @@
...
+      profile = StudentProfile.new(parent_params)
+    when "mentor"
+      profile = MentorProfile.new(mentor_params)
+    end
+
+    if profile.save
+      case params[:profileType]
+      when "mentor"
+        RegistrationMailer.welcome_mentor(profile.account.id).deliver_later
+      end
```
shaunxp20, line 21: The code here was a late add (and moved here pretty much out of frustration and to keep things moving along). I had moved this to the `RegisterToCurrentSeasonJob` (seems like an appropriate place for this, since that's where the student one is), but the tests were not happy about it, I also tried adding it to an `after_create` hook on the mentor profile, but that also didn't work. So basically, I had to leave it in `ProfileCreating` for the existing/old flows, and then I had to add it here for the new flow. 😕

app/controllers/new_registration_controller.rb
```diff
@@ -1,11 +1,84 @@
...
+
+      SignIn.call(
```
shaunxp20, line 23: The new flow calling `SignIn` which will align/sync the new/old registration flows (if that makes sense).

I didn't touch the `SignIn` code at all. Something confusing/weird about reusing this code though - this does not actually do the redirect, that happens in the JS layer, after the axios call.

app/javascript/new_registration/components/FormWrapper.vue
viviancan: I added a link to this comment in #3082

viviancan (approved):
👍 Everything looks great, Shaun! Thanks for all of the comments and helpful context. I cloned the branch locally and was able to register a student locally. I know we still have to update for the mentor and account for the email issue for the parent path. Besides that, all looks good! I'm excited to see it on QA!

shaunxp20:
Thanks Vivian! 🙌 There's still some followup to do, but I'll get this one merged in to keep things moving along! 😁

PR #3095, opened 2021-11-05
Fix images

shaunxp20:
- Moved images from `app/javascripts/new_registration/assets/` to `app/assets/images`
- Added configuration to `webpack.yml`:
```
resolved_pahs:
    - app/assets/images`
```

- And then updated all the ``


app/javascript/new_registration/components/Navbar.vue
```diff
@@ -4,7 +4,7 @@
       <div class="flex justify-between">
           <div>
             <a href="#" class="flex items-center py-4 px-2">
-              <img id="tg-logo" :src="require('../assets/tg-girls-logo.png').default" alt="Technovation Girls Logo" />
+              <img id="tg-logo" :src="require('tg-girls-logo.png')" alt="Technovation Girls Logo" />
```
shaunxp20, line 7: One thing I'm kinda eh/meh/hmm on is that we're adding `app/assets/images` to `required_paths`, so the images get required w/o `images/` in the path, whereas if we just added `app/assets` we could do `:src="images/blah....`, but we only need the images from `app/assets`, so I thought it was more ideal just to include what we need, but it makes it less ideal (imho) for the `src`. 🤷 🤔 🤯

viviancan (approved):
LGTM!

Thanks for tackling this one @shaunxp20! 🎉

shaunxp20:
Awesome! No problem! I'll get this one merged in so we can have images working again on QA! #Excited! 😀

PR #3098, opened 2021-11-11
Display back-end errors in Vue Formulate

shaunxp20:


app/controllers/new_registration_controller.rb
```diff
@@ -27,7 +27,7 @@ def create
         redirect_to: "student_dashboard_path"
       )
     else
-      render json: {errors: profile.errors}, status: :unprocessable_entity
+      render json: {errors: ValidationErrorMessagesConverter.new(errors: profile.errors).call}, status: :unprocessable_entity
```
shaunxp20, line 30: I added this new ValidationErrorMessagesConverter service that basically converts the error keys/attributes into what Vue Formulate expects.

viviancan (approved):
Looks good to me! I think the error converter was a great solution for the issue we were having. Did not try locally, but will test on QA!

PR #3099, opened 2021-11-12
Welcome email for the beginners division

shaunxp20:


app/jobs/register_to_current_season_job.rb
```diff
@@ -55,7 +55,11 @@ def prepare_student_for_current_season(record)
 
     subscribe_to_newsletter(record, :student)
 
-    RegistrationMailer.welcome_student(record).deliver_later
+    if profile.division.junior? || profile.division.senior?
+      RegistrationMailer.welcome_student(record).deliver_later
+    elsif profile.division.beginner?
+      RegistrationMailer.welcome_beginner_student(record).deliver_later
```
shaunxp20, line 61: This is the new mailer I added, I copied it from `welcome_student` and thus this one became `welcome_beginner_student`, but now I'm wondering if it should be `welcome_parent` or `welcome_beginner_division`? Preference/thoughs? 🤔 😃

viviancan: This is tough! Because the `welcome_beginner_student` follows the current convention of `welcome_student`, but at the same time, the email is technically going to the parent, even though it is a student that will be participating. And the umbrella theme of it all is the `beginner_division`

I don't have a strong preference either way! I think that they are all explicit in their own ways! 👀

viviancan (approved):
LGTM! 🎉

PR #3105, opened 2021-11-17
Front-end validation and minor refactoring of registration flow

shaunxp20:


app/javascript/new_registration/components/FormWrapper.vue
```diff
@@ -9,67 +9,50 @@
   >
 
     <div v-show="step === 1">
-      <StepOne :form-values="formValues"/>
-      <NextButton @next="next()"/>
+      <StepOne :form-values="formValues" @next="next" />
```
shaunxp20, line 12: I moved all of the "Next" and "Back" buttons into each "Step" component, so that each step/page can control/disable its own  buttons. Each page will disable its "Next" button if there are validation errors on that page.

app/javascript/new_registration/components/FormWrapper.vue
```diff
@@ -9,67 +9,50 @@
...
     </div>
 
     <div v-show="step === 2">
-      <StepTwo :form-values="formValues"/>
```
shaunxp20, line 17: There is no more `StepTwo`, now there is just `MentorStepTwo` and `StudentStepTwo`.

I like how this makes it so that all the logic about which step to display is kept in `FormWrapper`, which I'm thinking of as the main/parent registration component; and I think it makes sense it maintains/knows this knowledge.

A random side note about the step names -  StepOne, StepTwo, etc., it makes me think a little about positional arguments vs keyword arguments. Positional arguments are more brittle b/c if you change the order it won't work whereas keyword arguments can be in any order. It's not quite the same with these components b/c the order could change and it would still work (but the step names would out of order, if that makes sense). Or adding a new step would be a little more work. Anywho, random rant about how these names are a little brittle I think. Could possibly rename them to remove the step numbers, like the first one could be ProfileSelection (or something). #RandomThoughts!

app/javascript/new_registration/components/MentorStepTwo.vue
```diff
@@ -1,122 +1,177 @@
...
+        <h2 class="registration-title">Mentor Information</h2>
+
+        <div class="formulate-input-wrapper name-group">
+          <FormulateInput
+            name="firstName"
+            type="text"
+            label="First Name"
+            placeholder="First Name"
+            validation="required"
+            validation-name="First name"
```
shaunxp20, line 16: I added this `validation-name` attribute to most of the required fields (to the fields w/ multiple words). It's basically the name for the validation message.

"First name is required" vs "FirstName is required" (without this `validation-name` attribute)

app/javascript/new_registration/components/MentorStepTwo.vue
```diff
@@ -1,122 +1,177 @@
...
-    <div class="mentor-information">
-      <h2 class="registration-title">Set your personal summary</h2>
-      <p class="text-left">Add a description of yourself to your profile to help students get to know you. You can change this later.</p>
-      <FormulateInput
-          name="mentorSummary"
-          type="textarea"
-          validation="required|min:100,length"
-          id="mentorSummary"
-      />
+    <ReferredBy />
```
shaunxp20, line 115: I added this `ReferredBy` component, that the student component also uses.

Btw, It looks like there are a lot changes in here (and in the student component), but most of it is just formatting, b/c I added/merged StepTwo into each of them.

app/javascript/new_registration/components/ReferredBy.vue
```diff
@@ -0,0 +1,35 @@
...
+        {value: 6, label: 'Web search'},
+        {value: 7, label: 'Teacher'},
+        {value: 8, label: 'Parent/family'},
+        {value: 9, label: 'Company email'},
+        {value: 11, label: 'Other'}
+      ]
+    }
+  }
+}
+</script>
```
shaunxp20, line 35: This is the new `ReferredBy` component. I like having this separate now b/c when we add the text field for "Other", (we'll have to change some back-end files, but) I think this should be the only front-end file we'll have to change.

app/javascript/new_registration/components/StudentStepTwo.vue
```diff
@@ -1,80 +1,135 @@
...
   name: "StudentStepTwo",
+  components: {
+    ContainerHeader,
+    ReferredBy,
+    PreviousButton,
+    NextButton
+  },
+  data () {
+    return {
+      hasValidationErrors: true
```
shaunxp20, line 119: **A brief of the new validation stuff**

1. Each step now has a `hasValidationErrors` data property, field thingum.

app/javascript/new_registration/components/StudentStepTwo.vue
```diff
@@ -1,80 +1,135 @@
...
-      <FormulateInput
-          v-else
-          name="parentEmail"
-          type="email"
-          label="Parent Email Address (Optional)"
-          placeholder="Parent Email address"
-      />
+    <div class="registration-btn-wrapper">
+      <PreviousButton @prev="$emit('prev')"/>
+      <NextButton @next="$emit('next')" :disabled="hasValidationErrors"/>
```
shaunxp20, line 98: **A brief of the new validation stuff**

2. Each "Next" button uses`hasValidationErrors` to determine if the button should be enabled or disabled.

app/javascript/new_registration/components/StudentStepTwo.vue
```diff
@@ -1,80 +1,135 @@
...
+    PreviousButton,
+    NextButton
+  },
+  data () {
+    return {
+      hasValidationErrors: true
+    }
+  },
+  methods: {
+    checkValidation() {
```
shaunxp20, line 123: **A brief of the new validation stuff**

3. Each step has it's own `checkValidation` method, this can be customized for each step.

app/javascript/new_registration/components/StudentStepTwo.vue
```diff
@@ -1,80 +1,135 @@
...
+      const validationErrors = document.querySelector('#step-two-student div[data-has-errors="true"]')
```
shaunxp20, line 124: **A brief of the new validation stuff**

4. Right now to determine if a step/page is valid/invalid, we look for a `div` with a `data-has-errors="true" attribute. Even if there aren't validation errors visible, they will still be in the HTML markup (I randomly discovered this when inspecting elements on the page). This is pretty brittle, but it works for now.

app/javascript/new_registration/components/StudentStepTwo.vue
```diff
@@ -1,80 +1,135 @@
...
+
+      if (validationErrors) {
+        this.hasValidationErrors = true
+      } else {
+        this.hasValidationErrors = false
+      }
```
shaunxp20, line 130: **A brief of the new validation stuff**

5. And then finally, to make the circle complete, we set the value mentioned back in step 1. 😅

app/javascript/new_registration/components/StudentStepTwo.vue
```diff
@@ -1,80 +1,135 @@
...
-      <h2 class="registration-title">Parent Information</h2>
+          <FormulateInput v-if="formValues.profileType === 'parent'"
+            name="parentEmail"
+            type="email"
+            label="Parent Email Address"
+            placeholder="Parent Email address"
+            validation="required|email"
+            validation-name="Email address"
+            @keyup="checkValidation"
+            @blur="checkValidation"
```
shaunxp20, line 81: **A brief of the new validation stuff**

2.5. Oops, I forgot a step, on each required field, we add these event listeners to check the step/page's validation state. These two events seem to be working - I tried listening to different events before and was running into very weird situations, where the Next button would be disabled, but there were no validation errors, or vice versa. I have been playing w/ these two (`keyup`, and `blur`) and I haven't seen any weirdness, but that's something to keep in mind, keep a look out for. I did both keyup and blur b/c for example the birthday, you could just use the mouse and not the keyboard, so keyup wouldn't fire, but you would "leave" the field so blur would fire.

app/javascript/new_registration/components/FormWrapper.vue
viviancan: Great point on the naming, Shaun! Love that idea, and I think doing a rename would be helpful for any future versions of this.

app/javascript/new_registration/components/MentorStepTwo.vue
viviancan: 👍

viviancan (approved):
LGTM! Excited to see the validation on QA!

shaunxp20:
> LGTM! Excited to see the validation on QA!

Awesome, thank you! It's headed to QA now! 😄

PR #3143, opened 2021-11-24
Updated footer in vue & ruby

rsgonzal:
Updating footers in location where i've found them

### DIFF
--- a/app/views/application_rebrand/_footer.en.html.erb
+++ b/app/views/application_rebrand/_footer.en.html.erb
@@ -9,19 +9,39 @@
 
     <section id="links" class="flex flex-wrap-reverse md:flex-nowrap justify-between my-12">
       <div id="icons" class="flex mx-auto md:mx-0 my-8 md: my-0 w-1/2 md:w-48 justify-between">
-        <%= image_tag "new_registration/fb.png", class:"social-icon"%>
-        <%= image_tag "new_registration/twitter.png", class:"social-icon"%>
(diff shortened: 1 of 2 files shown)


rsgonzal:
I may have been way off with there 😅

app/views/application_rebrand/_footer.en.html.erb
```diff
@@ -9,19 +9,39 @@
...
-        <a href="#">Contact</a>
-        <a href="#">Terms of Use</a>
-        <a href="#">Press</a>
-        <a href="#">Privacy Policy</a>
-        <a href="#">Donate</a>
+        <%= link_to "FAQ",
+            "https://iridescentsupport.zendesk.com/hc/en-us/categories/115000091348-Technovation"
+          %>
+        <%= link_to "Contact",
+            "mailto:support@technovation.org"
```
shaunxp20, line 31: I think this is in a config/ENV setting, anywho, if so, no worries, we can update it later! 😃

shaunxp20 (approved):
Amazing! Thanks for updating all of this! Looks good to me! Feel free to merge, I updated it, so it's up to date w/ QA. :shipit: 👍 😀

app/views/application_rebrand/_footer.en.html.erb
rsgonzal: it is! but i wasn't sure how to do that! i was also trying to make the year in the footer pull from config, but just couldn't figure it out.. 😅

PR #3153, opened 2021-11-25
Update division descriptions 

rsgonzal:
updating division ages and adding beginner division. I found this on the my team pages under division. not sure if used in other parts of the app

### DIFF
--- a/app/javascript/registration/components/AgeVerification.vue
+++ b/app/javascript/registration/components/AgeVerification.vue
@@ -50,16 +50,22 @@
         <h4>Team divisions explained:</h4>
 
         <dl>
+          <dt class="color--secondary">Beginner Division</dt>
+          <dd>
+          Team members are between 8-12 years old as of {{ cutoffDay }}.
+          </dd>
+        
           <dt class="color--secondary">Junior Division</dt>
           <dd>
-            Team members are between 10-14 years old as of {{ cutoffDay }}.
+            Team members are between 13-15 years old as of {{ cutoffDay }}.
           </dd>
 
           <dt class="color--secondary">Senior Division</dt>
           <dd>
-            Team members are between 15-18 years old as of {{ cutoffDay }}.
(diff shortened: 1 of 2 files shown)


app/javascript/registration/components/AgeVerification.vue
```diff
@@ -50,16 +50,22 @@
         <h4>Team divisions explained:</h4>
 
         <dl>
```
shaunxp20, line 52: Btw, I'm pretty sure this file/component is just for the old registration form/flow - maybe it's used somewhere else? 🤔

rsgonzal: ah, i thought this was the mentor version of the my team section 
Screen Shot 2021-11-25 at 10 21 53 AM
🤔

shaunxp20: Ah, looks like that screenshot is for the other file (`app/views/teams/_division.en.html.erb`) in this PR, in that file I see the "which makes" text in there. Anywho, I guess no harm in updating this file as well, maybe it is used somewhere else. 🤷 😁

rsgonzal: ohhhhh. i see. i assumed there would be two files because mentor & student one look different but i guess they're just using different styles.. 😅🤔😭

shaunxp20 (approved):
Just updated this branch to up-to-date w/ QA. Feel free to merge whenever! Thanks again for helping out! 🙌 🥇 🌞 

I realize we're still kinda ironing out the flow, but I'd say w/ these PRs, after CircleCI passes (and it's up-to-date w/ QA), just feel free to merge it. I feel like I'm a blocker on these sometimes, and I don't want to be! We used to use a tag for "no code review necessary" (or something like that), and I think these would fall under that label; and if something is off after it's merged, we can always update it later. #JustMyTwoCents! 😃✌️

rsgonzal:
sounds like a plan! hoping i don't break anything 😅

Issue #3208, opened 2021-12-22
Rename source code section and edit page


rsgonzal:
![Screen Shot 2021-12-21 at 11.01.57 PM.png](https://images.zenhubusercontent.com/60527169cde2772242d0de95/1e9de817-275c-4ff9-95f1-4f81038bb329)

rsgonzal:
Nick's comment from #3011

> AI Project Submissions
>AI submissions should be a zip file with the following included:
>
>- The dataset you used to train your model
>- This can be a screenshot, spreadsheet, or direct link to an image/sound folder, etc.
>- Picture(s) of prototypes
>- This can be cardboard models, drawings, sample devices, etc.
>- (For online inventions) A link with sample login information

Where on this page would this go ideally? Might be confusing to have all of this for AI Projects and nothing for Mobile Apps?

ophelieh:
Looks good! But I would edit the section name. See ticket #3011

rsgonzal:
Hey Ophelie! Which section are you referring to? The larger one that is "Code."?

If so, what would then name the "Source Code" one?

ophelieh:
Maybe we can just call that section "Technical" - kinda like it used to be in the rubric so it doesn't take away from the upload your technical work? 

Also for uploading Technical work will the view change if they choose mobile app vs AI project? If it stays the same, we need to change the text above the "Choose file" button in the second image.

rsgonzal:
for now it'll stay the same. I'll edit the text.

rsgonzal:
bullet point to be: Technical Additions like the title (instead of 'upload your source code' or 'upload your technical work' in the mockup)

rsgonzal:
@ophelieh @nickiredescent do you want to upload this line?

Help for submitting your source code is [here](https://iridescentsupport.zendesk.com/hc/en-us/articles/360019590314-How-do-I-submit-my-source-code-) or take a look at the [debugging curriculum unit](https://technovationchallenge.org/curriculum/coding-14-debugging-tips-in-app-inventor-and-thunkable/) for more resources.

ophelieh:
Looks great! 

Could you edit to read: "For help submitting your source code, check here (link) or take a look....

nickiredescent:
Agree with Ophelie

ophelieh:
I added some of my comments for this in ticket #3011. 

[...]

PR #3214, opened 2022-01-07
updating source code language to technical

rsgonzal:


rsgonzal:
My attempt at #3208, tried to do most text changes from source code to technical work/additions

did not change any technical pieces or actual code/paths

PR #3258, opened 2022-02-22
3216 top menu mobile bug - Updated css/html structure of top menu to account for mobile devices 

viviancan:


devkonrad (approved):
The changes looks fine to me. :)

app/javascript/stylesheets/student_profile.css
```diff
@@ -59,6 +59,10 @@ nav {
    .tab-content{
         width: 48rem;
     }
+
```
shaunxp20, line 62: A couple of random FWIW comments/notes...

`@media only screen and (min-width: 768px)` - I know this wasn't from this PR, but I looked into this and found Tailwind's `screen` function, which says, "The screen function allows you to create media queries that reference your breakpoints by name instead of duplicating their values in your own CSS." 

https://tailwindcss.com/docs/functions-and-directives#screen

And this SO post says you can just use `@screen md` for this, neat! I just learned this! 😃 

https://stackoverflow.com/a/55551805

I also see `.tw-hidden` and the newly added `.tw-mobile-show` scoped to this breakpoint. Random thought, I think they could be inlined as `md:block` and `md:hidden`, or the classes could be named to be more consistent, or renamed to reflect the `md` breakpoint.

Anywho, all just random food for thought! 🍪 😋 😁

app/views/application_rebrand/_navigation.html.erb
```diff
@@ -30,20 +31,24 @@
...
-                                    SeasonToggles.survey_link(
-                                      current_scope, "url",
-                                      format_url: true,
-                                      account: current_account
-                                    ),
-                                    target: "_blank"
-                        %>
-                        <% end %></li>
+                      <% if SeasonToggles.survey_link_available?(current_scope, current_account) %>
+                        <li>
```
shaunxp20, line 41: I like how this change prevents an empty `<li>`! 👍 😄

shaunxp20 (approved):
Wow, thanks for taking care of this! 🏆 I just left some random notes, but overall LGTM! You're really going above and beyond here! 😅 😲 😀

app/javascript/stylesheets/student_profile.css
viviancan: Thanks for the feedback, @shaunxp20!! I will take a look at updating based on your notes ⭐ 🎉

app/javascript/stylesheets/student_profile.css
```diff
@@ -16,8 +16,8 @@ nav {
     font-family: 'Poppins', sans-serif;
 }
 
-.tw-hidden{
-    display: none;
+.tw-sm-hidden{
```
viviancan, line 19: I decided to go with `.tw-sm-hidden` as the class name since tailwind is mobile-first. I thought it made sense since these items are hidden in the small breakpoint.

app/javascript/stylesheets/student_profile.css
```diff
@@ -42,26 +42,23 @@ nav {
...
 }
 
 #show-container{
     @apply w-full
 }
 
-@media only screen and (min-width: 768px){
-    .tw-hidden{
-        display: block;
+@screen md{
```
viviancan, line 52: Updated the media query here based on the tailwind screen directive

viviancan:
I originally created the `tw-sm-hidden` class instead of using inline `sm:hidden` `md:block` bc I could not get `sm:hidden` to hide the links appropriately. I'm not sure if it has to do tailwind's mobile-first approach, but that combination of inline styling would not work.

When using just `hidden` and then `md:block` the links would stay hidden because there is already a `hidden` class in `_base.scss` that would consistently override the tailwind styles. I did not want to mess with that class since it is probably used in multiple places throughout the application!

viviancan:
> Wow, thanks for taking care of this! 🏆 I just left some random notes, but overall LGTM! You're really going above and beyond here! 😅 😲 😀

I originally created the `tw-sm-hidden` class instead of using inline `sm:hidden` `md:block` bc I could not get `sm:hidden` to hide the links appropriately. I'm not sure if it has to do tailwind's mobile-first approach, but that combination of inline styling would not work.

When using just `hidden` and then `md:block` the links would stay hidden because there is already a `hidden` class in `_base.scss` that would consistently override the tailwind styles. I did not want to mess with that class since it is probably used in multiple places throughout the application! 

Thanks for the review and feedback, Shaun!

Issue #3104, opened 2021-11-17
Add text input field for "other" referral option


shaunxp20:
On the new registration form, for the "How did you hear about Technovation?" question, when someone selects "Other", a text field should be displayed so they can (optionally) input something.

dboyer:
Nothing that the "Other" field may be used by registrants to indicate a chapter name or ambassador. We should try to add this in soon after launch.

rsgonzal:
works! and shows up when we search - perfect!

![Screen Shot 2022-03-08 at 9.40.02 AM.png](https://images.zenhubusercontent.com/60527169cde2772242d0de95/f417d9b7-dfbb-487c-a3fa-02e83df909da)

PR #3259, opened 2022-02-22
Adds the text field for save the 'other' referral

devkonrad:
Adds the text field for save the 'other' referral.

Changes:
modified:  app/controllers/new_registration_controller.rb
modified:  app/javascript/new_registration/components/ReferredBy.vue

Refs: https://github.com/Iridescent-CM/technovation-app/issues/3104


app/javascript/new_registration/components/ReferredBy.vue
```diff
@@ -8,6 +8,16 @@
       input-class="mentorSelectClass"
       id="referredBy"
       placeholder="Select an option"
+      @change="changeOption"
```
shaunxp20, line 11: Instead of wiring this up manually via an event, it can be achieved through a binding:
https://vuejs.org/guide/essentials/forms.html#form-input-bindings

And here's some model binding info for the Vue Formulate library we're using:
https://vueformulate.com/guide/#model-binding
https://vueformulate.com/guide/inputs/#model-binding

`v-model="referralOption"` And then `referralOption` will contain the value selected. 👍 😄

app/javascript/new_registration/components/ReferredBy.vue
```diff
@@ -8,6 +8,16 @@
...
+    />
+
+    <FormulateInput
+      name="referredByOther"
+      id="referredByOther"
+      type="text"
+      label="Who referred you?"
+      placeholder="Referred by"
+      v-show="isOtherOption"
```
shaunxp20, line 20: After using a binding, then this could just be: `v-show="referralOption == 11"` - which I think this is okay in this case, because what the value 11 is equal to is just a few lines down, it's all contained in this component. 😄

app/javascript/new_registration/components/ReferredBy.vue
devkonrad: Great tip, @shaunxp20 ! The code will be more clean this way! Thanks!

app/javascript/new_registration/components/ReferredBy.vue
devkonrad: Cool! I Got it! Do more with less lines of code! :)

app/javascript/new_registration/components/ReferredBy.vue
```diff
@@ -8,6 +8,16 @@
...
+      v-model="referralOption"
+    />
+
+    <FormulateInput
+      name="referredByOther"
+      id="referredByOther"
+      type="text"
+      label="Who referred you?"
+      placeholder="Referred by"
+      v-show="referralOption == 11"
```
shaunxp20, line 20: Looking back, it was my sloppy code sample that used loose equality`==` and your function had strict equality `===` - Not sure it would make a difference here? 🤔 🧐 Probably more just good practice 🤔, nonetheless I like your way of using strict equality, FWIW! 😄

shaunxp20 (approved):
Looks good to me! Simple changes, straight forward, I like it! 🥳 👍 🙌

app/javascript/new_registration/components/ReferredBy.vue
devkonrad: Well... How it is a controlled scenario (the varibles are declared inside the component) I guess that we dont have to worry so much about datatype (by using === strict mode), but as a good practice i guess that its better to turn it to "strict mode" anyway. :)

Issue #3251, opened 2022-02-10
Apply new styles to location interrupt screen


rsgonzal:
As a student, I need to update my location in order to access the dashboard.

Currently, the box that shows up has a mix of old and new styles

![Screen Shot 2021-11-30 at 10.18.15 AM.png](https://images.zenhubusercontent.com/60527169cde2772242d0de95/9dc9dc46-91c5-4126-a05e-e68aa0a41e8f)

Task: Apply new styling conventions and fix the size of the input fields
![Screen Shot 2022-02-10 at 10.51.51 AM.png](https://images.zenhubusercontent.com/60527169cde2772242d0de95/38744d59-c8d9-4b0d-b997-6a9d6be748f4)

rsgonzal:
- [ ] Mentor or judge visits my profile and edits location
<img width="1165" alt="Screen Shot 2022-03-13 at 9 32 01 PM" src="https://user-images.githubusercontent.com/22801026/158105383-c20673bd-426e-4c21-9f51-228ff6d8d46c.png">

- [ ] Check team location as a student

<img width="844" alt="Screen Shot 2022-03-13 at 9 36 19 PM" src="https://user-images.githubusercontent.com/22801026/158105663-65b4f5ac-6350-44ca-98c6-c34821d560e7.png">

devkonrad:
> There are a few places where styles are broken, we can start with the last one in this list that affects students.
> 
> Let's chat tomorrow on the best way to approach mentor & judge views.
> 
> * [ ]  Mentor or judge visits my profile and edits location
> 
> <img alt="Screen Shot 2022-03-13 at 9 32 01 PM" width="1165" src="https://user-images.githubusercontent.com/22801026/158105383-c20673bd-426e-4c21-9f51-228ff6d8d46c.png">
> 
> * [ ]  Signing up as mentor or judge
> 
> <img alt="Screen Shot 2022-03-13 at 9 39 36 PM" width="1160" src="https://user-images.githubusercontent.com/22801026/158106026-4b8aadea-6337-4338-a2d5-8eed73aa300f.png">
> 
> * [ ]  Check team location as a mentor
> 
> <img alt="Screen Shot 2022-03-13 at 9 32 01 PM" width="1165" src="https://user-images.githubusercontent.com/22801026/158105779-db4cf3f3-b6e8-48b3-8dd0-c9c25cb05e06.png">
> 
> * [ ]  Check team location as a student
> 
[...]

PR #3263, opened 2022-02-28
Apply new styles to location interrupt screen

devkonrad:
Apply new styles to location interrupt screen

Changes:
modified: app/javascript/location/components/LocationForm.vue
modified: app/views/location_details/show.en.html.erb 

Refs: https://github.com/Iridescent-CM/technovation-app/issues/3251

### DIFF
--- a/app/javascript/location/components/LocationForm.vue
+++ b/app/javascript/location/components/LocationForm.vue
@@ -3,192 +3,191 @@
     class="panel panel--contains-bottom-bar panel--contains-top-bar"
     @submit.prevent="handleSubmit"
   >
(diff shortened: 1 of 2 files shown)


viviancan (commented):
This change looks good. The only thing I would say is that I'm not sure where else this interrupt screen is rendered, so there may be some inconsistent styles depending on the view. I think only the new tailwind CSS is only used in the student view, although it could have changed when I was out! :) @shaunxp20 @rsgonzal

viviancan (approved):
Looks good to me! We will need to verify the styles for the different user types on QA

PR #3277, opened 2022-03-07
Added judge registration to the new registration flow

viviancan:


app/javascript/new_registration/components/JudgeStepTwo.vue
```diff
@@ -0,0 +1,162 @@
...
+          placeholder="Job Title"
+          validation="required"
+          validation-name="Job title"
+          @keyup="checkValidation"
+          @blur="checkValidation"
+        />
+      </div>
+    </div>
+
+    <ReferredBy />
```
shaunxp20, line 94: Nothing more than a thinking out loud comment - I wonder if some of the other account level/common fields could be extracted into a SFC (single file component) like this, and then including them where they're needed would be as simple as this one. Maybe in the future! 😅

app/javascript/new_registration/components/JudgeStepTwo.vue
```diff
@@ -0,0 +1,162 @@
...
+        'Female',
+        'Male',
+        'Non-binary',
+        'Prefer not to say'
+      ],
+      hasValidationErrors: true
+    }
+  },
+  methods: {
+    checkValidation() {
```
shaunxp20, line 132: I remember adding this functionality to prevent going to the next page while there are validation errors, it was a bit rushed before registration opened. Anywho, looking at it now, it probably could be cleaned up/done better. Also, just a #RandomComment! 😬 😅

shaunxp20 (approved):
I really like how clean, simple and straight forward this is! Just makes sense! And having judges register via the new registration flow, AMAZING!!! Huge win! ⭐ 💯 🙌

devkonrad (approved):
The way that the feature are implemented looks good to be. It follow some patterns that were already established for the other kind of users. Great! :)

app/javascript/new_registration/components/JudgeStepTwo.vue
viviancan: I really like that idea! Would be great to extract a few of those pieces and be able to reuse where needed

Issue #3219, opened 2022-01-10
Remove mentor age any place it is publicly visible


dboyer:
We ask for a mentor's age as part of the registration process to verify they are over 18. Their age should not display anywhere publicly visible, however, including on their public profile or when they are invited to join a team.

- in the mentor search for mentors
![Screen Shot 2022-02-16 at 3.11.36 PM.png](https://images.zenhubusercontent.com/60527169cde2772242d0de95/e0328c2f-e70c-496c-a01f-57205c3937bf)

rsgonzal:
- in the mentor search for mentors

rsgonzal:
@devkonrad where should these changes be visible? I am still seeing the age on this page
![Screen Shot 2022-03-08 at 9.48.58 AM.png](https://images.zenhubusercontent.com/60527169cde2772242d0de95/d1cbb841-5a1c-44be-b601-50745bf5207c)

devkonrad:
> @devkonrad where should these changes be visible? I am still seeing the age on this page ![Screen Shot 2022-03-08 at 9.48.58 AM.png](https://camo.githubusercontent.com/4d1cafc2c97d54ac0f295a117d503fcfc112b08894df69e55afc776c14825bf1/68747470733a2f2f696d616765732e7a656e68756275736572636f6e74656e742e636f6d2f3630353237313639636465323737323234326430646539352f64316362623834312d356131632d343462652d623630312d353037343562663532303763)

Hi! I was changed the specific mentor view, but looks like the participants lis uses another tempplate. I juts fixed that and created a new PR. :)

devkonrad:
> > @devkonrad where should these changes be visible? I am still seeing the age on this page ![Screen Shot 2022-03-08 at 9.48.58 AM.png](https://camo.githubusercontent.com/4d1cafc2c97d54ac0f295a117d503fcfc112b08894df69e55afc776c14825bf1/68747470733a2f2f696d616765732e7a656e68756275736572636f6e74656e742e636f6d2f3630353237313639636465323737323234326430646539352f64316362623834312d356131632d343462652d623630312d353037343562663532303763)
> 
> Hi! I was changed the specific mentor view, but looks like the participants lis uses another tempplate. I just fixed that and created a new PR. :)

dboyer:
Confirmed this is working as intended on QA

PR #3281, opened 2022-03-09
Hide mentors age from participants view

devkonrad:
Looks like there another places that displays Mentors Age, this PR cover the 'participant view file' and hide the age info for mentors on that.

Changes:
modified: app/views/chapter_ambassador/participants/show.en.html.erb

Refs: https://github.com/Iridescent-CM/technovation-app/issues/3219

### DIFF
--- a/app/views/admin/participants/show.html.erb
+++ b/app/views/admin/participants/show.html.erb
@@ -43,7 +43,7 @@
             <dt><%= t('models.account.email') %></dt>
             <dd><%= mail_to @account.email %></dd>
 
-            <% if not @account.is_a_mentor? %>
+            <% if !@account.is_a_mentor? %>
               <dt><%= t('models.account.age') %></dt>
               <dd><%= @account.age %></dd>
             <% end %>
--- a/app/views/admin/participants/show.html.erb
+++ b/app/views/admin/participants/show.html.erb
@@ -43,8 +43,10 @@
             <dt><%= t('models.account.email') %></dt>
             <dd><%= mail_to @account.email %></dd>
 
-            <dt><%= t('models.account.age') %></dt>
-            <dd><%= @account.age %></dd>
+            <% if not @account.is_a_mentor? %>
+              <dt><%= t('models.account.age') %></dt>
+              <dd><%= @account.age %></dd>
+            <% end %>
 
             <% if @account.student? %>
               <dt><%= t('models.account.division') %></dt>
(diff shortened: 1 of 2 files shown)


app/views/admin/participants/show.html.erb
```diff
@@ -43,8 +43,10 @@
             <dt><%= t('models.account.email') %></dt>
             <dd><%= mail_to @account.email %></dd>
 
-            <dt><%= t('models.account.age') %></dt>
-            <dd><%= @account.age %></dd>
+            <% if not @account.is_a_mentor? %>
```
shaunxp20, line 46: A very minor thing here, but could you use `!` (without a space) instead of `not` here.

A lot of Ruby style guides don't recommend it:

> The and, or, and not keywords are banned. It's just not worth it. Always use &&, ||, and ! instead.

https://github.com/airbnb/ruby#conditional-keywords

Also, fwiw, I've been trying to slowly adhere the codebase to [Standard](https://github.com/testdouble/standard), which also doesn't allow `not`. 😅 😄

devkonrad: > A very minor thing here, but could you use `!` (without a space) instead of `not` here.
> 
> A lot of Ruby style guides don't recommend it:
> 
> > The and, or, and not keywords are banned. It's just not worth it. Always use &&, ||, and ! instead.
> 
> https://github.com/airbnb/ruby#conditional-keywords
> 
> Also, fwiw, I've been trying to slowly adhere the codebase to [Standard](https://github.com/testdouble/standard), which also doesn't allow `not`. sweat_smile smile

Thank you for the tip, Shaun! I will follow this for the next times. :)

shaunxp20 (approved):
The changes look fine to me! 👍 

A few things caught my eye though:
- One of the files changed was for the admin view - I'm not certain if we want to hide the age there
- I looked at the [screenshot in #3219](https://github.com/Iridescent-CM/technovation-app/issues/3219#issuecomment-1062043352) - and I think that might be a different page/file than what was changed in this PR 🤔 but maybe I'm wrong, either way we can move this one forward and get it onto QA and see if it's what we want 👍

dboyer:
The mentor team confirmed that the mentor's age should still be visible on the admin page which only Technovation admins can see.

devkonrad:
> The mentor team confirmed that the mentor's age should still be visible on the admin page which only Technovation admins can see.

Thanks @dboyer , i will adjust that!

PR #3284, opened 2022-03-10
Restrict rebranded invites to students only

shaunxp20:


app/controllers/concerns/team_member_invite_controller.rb
```diff
@@ -17,7 +17,7 @@ def show
         redirect_to: "#{current_scope}_team_member_invite_path"
       )
     else
-      if @invite.inviter_type == "StudentProfile" && @invite.status == "pending"
+      if current_scope == "student" && @invite.status == "pending"
```
shaunxp20, line 20: This logic was off a bit - a `StudentProfile` can invite a student or a mentor (that's why the mentors were seeing the rebranded invite) - I updated this to check which account type is currently logged in - if it's a student, then we'll show the rebranded invite; and mentors will fall through to use the old invite stuff.

viviancan (approved):
Great fix, Shaun! 🥳

app/controllers/concerns/team_member_invite_controller.rb
devkonrad: Thats make sense. :)

Issue #3065, opened 2021-10-05
Add column option for media consent on participant page


rsgonzal:
For both ChAs & Admin
![Screen Shot 2021-10-05 at 12.59.07 PM.png](https://images.zenhubusercontent.com/60527169cde2772242d0de95/3dd4b193-339d-4eb1-8e17-0fc0335a16ac)

PR #3286, opened 2022-03-11
Add media consent filter on participant search

devkonrad:
Add media consent filter on participant search

Changes:
modified: app/data_grids/accounts_grid.rb

Refs: https://github.com/Iridescent-CM/technovation-app/issues/3065

### DIFF
--- a/app/data_grids/accounts_grid.rb
+++ b/app/data_grids/accounts_grid.rb
@@ -169,9 +169,9 @@ class AccountsGrid
 
   column :media_consent do |account, grid|
     if account.student_profile.present? and account.student_profile.media_consent.present?
-      account.student_profile.media_consent.consent_provided ? "yes" : "no"
+      account.student_profile.media_consent.consent_provided ? "Yes" : "No"
     else
-      "no"
+      "-"
     end
   end
 
--- a/app/data_grids/accounts_grid.rb
+++ b/app/data_grids/accounts_grid.rb
@@ -167,6 +167,14 @@ class AccountsGrid
     end
   end
 
+  column :media_consent do |account, grid|
+    if account.student_profile.present? and account.student_profile.media_consent.present?
+      account.student_profile.media_consent.consent_provided ? "yes" : "no"
+    else
+      "no"
+    end
+  end
+
   column :consent_waiver do
     if mentor_profile.present? || judge_profile.present?
       consent_waiver.present? ? "Signed" : "Not signed"
--- a/app/data_grids/accounts_grid.rb
+++ b/app/data_grids/accounts_grid.rb
@@ -148,7 +148,7 @@ class AccountsGrid
   column :background_check, if: ->(g) {
     g.admin or Array(g.country)[0] == "US"
   } do
-    if country_code == "US" and mentor_profile.present?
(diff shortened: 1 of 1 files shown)


app/data_grids/accounts_grid.rb
```diff
@@ -167,6 +167,14 @@ class AccountsGrid
     end
   end
 
+  column :media_consent do |account, grid|
+    if account.student_profile.present? and account.student_profile.media_consent.present?
+      account.student_profile.media_consent.consent_provided ? "yes" : "no"
```
shaunxp20, line 172: Minor thing here, but looks like to be consistent with our other fields, this should be capitalized - Yes/No. 😄

app/data_grids/accounts_grid.rb
```diff
@@ -167,6 +167,14 @@ class AccountsGrid
...
+    else
+      "no"
```
shaunxp20, line 174: This would put a "no" for the "Media consent" column for non-students (like judges and mentors), but I don't think it should be "no", but rather a `-` or `N/A` - I see both used in this file, @rsgonzal @dboyer thoughts?

dboyer: Mackenzie confirmed that we do not ask mentors or judges to complete a media consent unless they are participating in an event we've organized. She does not track those consents in the platform, currently. She would be interested in tracking mentors/judge media consent on the platform in the future though. That doesn't need to happen now in the middle of the season though.

app/data_grids/accounts_grid.rb
devkonrad: > Minor thing here, but looks like to be consistent with our other fields, this should be capitalized - Yes/No. smile

Thank you, @shaunxp20 ! Thats make sense!

app/data_grids/accounts_grid.rb
devkonrad: For me, make sense that if it's not applicable just put a "-".

app/data_grids/accounts_grid.rb
```diff
@@ -167,6 +167,14 @@ class AccountsGrid
     end
   end
 
+  column :media_consent do |account, grid|
+    if account.student_profile.present? and account.student_profile.media_consent.present?
```
shaunxp20, line 171: Just minor thing about using `and` here, copying a snippet of my [comment from #3281](https://github.com/Iridescent-CM/technovation-app/pull/3281#discussion_r825198488) from a Ruby styleguide:

> The and, or, and not keywords are banned. It's just not worth it. Always use &&, ||, and ! instead.

devkonrad: > Just minor thing about using `and` here, copying a snippet of my [comment from #3281](https://github.com/Iridescent-CM/technovation-app/pull/3281#discussion_r825198488) from a Ruby styleguide:
> 
> > The and, or, and not keywords are banned. It's just not worth it. Always use &&, ||, and ! instead.

Thanks, Shaun! I will pay more atention in this details. :)

viviancan (approved):
Thanks for making the changes according to the review, Alex! Looks good to me

Issue #3272, opened 2022-03-02
Allow admin & Cha Chas to see media consent status


rsgonzal:
Add section to admin that says
- Media Consent on File
- Media Consent not on File

![Screen Shot 2021-10-05 at 12.56.00 PM.png](https://images.zenhubusercontent.com/60527169cde2772242d0de95/4c30e144-da33-475b-824d-6ebf7429becc)

devkonrad:
Hi @dboyer and @rsgonzal , good morning! Just pasting here a preview of the funcionalitty to create the media consent independent of the parental consent.

![Captura de tela de 2022-03-22 09-12-04](https://user-images.githubusercontent.com/99268176/159479569-23e1b68d-7e26-4cc8-a21b-46991f606ee1.png)

![Captura de tela de 2022-03-22 09-12-24](https://user-images.githubusercontent.com/99268176/159479728-fa3c8165-b035-4255-b48d-74ae677289ce.png)

![Captura de tela de 2022-03-22 09-12-57](https://user-images.githubusercontent.com/99268176/159479570-6d5a0ef8-b92f-4549-800d-dd7be08f48fb.png)

devkonrad:
> That looks right to me!

Thanks @dboyer! The PR feature is in review!  :)

shaunxp20:
For media consents marked as ON FILE, should an admin person able to specify whether media consent was given or not?

dboyer:
Yes, please. I doubt it will be used frequently as more likely people just won't return the form. But we do need to differentiate between "ON FILE - consent granted" and "ON FILE - consent not granted."

viviancan:
I pulled down the branch and tested functionality locally. I was able to successfully see all the different states of media consent in the admin profile. Ref image below. 

Will we need to temporarily "enable" registration to QA this? I ended up "turning on" registration locally in order to create a few fresh student accounts. @dboyer @shaunxp20 

![media-consent-qa.png](https://images.zenhubusercontent.com/606b0f4fd13d53dff3d14fa5/2c24f7fe-0ed2-4a13-9484-843f8f8eaac6)

shaunxp20:
> Will we need to temporarily "enable" registration to QA this? I ended up "turning on" registration locally in order to create a few fresh student accounts. @dboyer @shaunxp20

[...]

PR #3288, opened 2022-03-14
Display media content status

devkonrad:
Display media content status

Changes:
modified: app/controllers/admin/paper_media_consents_controller.rb
modified: config/routes.rb 

Refs: https://github.com/Iridescent-CM/technovation-app/issues/3272

### DIFF
--- a/app/views/admin/participants/show.html.erb
+++ b/app/views/admin/participants/show.html.erb
@@ -203,10 +203,27 @@
     <% if @account.student_profile %>
       <div class="panel">
         <h5>
-            <% if @account.student_profile&.media_consent&.consent_provided %>
-              Media Consent on file
+            <% media_consent = @account.student_profile.media_consent %>
+            <% if media_consent.consent_provided %>
+              Media Consent provided via <%= 
+                media_consent.electronic_signature === "ON FILE" ? 
+                "paper form" : 
+                "online form" 
+              %>
             <% else %>
-              No Media Consent on file
+              <h5>
+                Indicate that a media consent is on file
+              </h5>
+              <%= link_to "Create media consent",
+                admin_paper_media_consent_path(
+                  id: @account.student_profile.id
+                ),
+                class: "button button--secondary",
+                data: {
+                  method: :post,
+                  confirm: "You are marking #{@account.full_name} " +
+                           "as having a media consent on file.",
+                } %>
             <% end %>
         </h5>
       </div>
--- a/app/views/admin/participants/show.html.erb
+++ b/app/views/admin/participants/show.html.erb
@@ -204,7 +204,7 @@
       <div class="panel">
         <h5>
             <% media_consent = @account.student_profile.media_consent %>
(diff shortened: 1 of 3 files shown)


app/views/admin/participants/show.html.erb
```diff
@@ -200,6 +200,18 @@
       </div>
     <% end %>
 
+    <% if @account.student_profile %>
+      <div class="panel">
+        <h5>
+            <% if @account.student_profile&.media_consent&.consent_provided %>
```
shaunxp20, line 206: `@account.student_profile&.` - it doesn't look like the `&` is necessary here, on line 203, we're already checking for a `student_account`

And another minor thing, it looks like the indentation is off for this entire `if` block. 🧐 😄

app/views/admin/participants/show.html.erb
```diff
@@ -200,6 +200,18 @@
...
+              Media Consent on file
```
shaunxp20, line 207: It seems like this language might not quite be accurate - the conditional on line 206 checks for consent, not if it's "on file" or not. 🤔 😄

devkonrad: I agree. @rsgonzal , is the purpose of this feature is check if the consent is provided?

dboyer:
Update after discussion with the program team. Ideally, they would like to know both if media consent has been given and the format it was given in (i.e. via the online form or via a paper form that someone emailed). So there are five possible states:

Agree to media consent via online form
Agree to media consent via paper form
Do not agree to media consent via online form
Do not agree to media consent via paper form
Has not completed media consent form

@devkonrad I realize this has expanded from the original plan for the issue. Please feel free to check in with someone if you need more details about how this could be implemented.

devkonrad:
As i could see, currently we have two params that controls the media_consent status inside de platform (consent_provided boolean flag and electronic_signature text info), so just por confirmation about the 5 status thar @dboyer mentioned:

Agree to media consent via online form => consent_provided is TRUE and electronic_signature is "Guardian Name"    
Agree to media consent via paper form => consent_provided is TRUE and electronic_signature is "ON FILE"
Do not agree to media consent via online form  => consent_provided is FALSE and electronic_signature is "Guardian Name"
Do not agree to media consent via paper form => consent_provided is FALSE and electronic_signature is "On File"
Has not completed media consent form  => The is no Media Content for the Student

Is that correct?

viviancan:
hey alex! Just did a quick QA locally! I think some of the wording/states are missing. Roxana will provide updated screens/details a little later. Thanks!

viviancan (changes requested):
hey alex! Just did a quick QA locally! I think some of the wording/states are missing. Roxana will provide updated screens/details a little later. Thanks!

devkonrad:
> hey alex! Just did a quick QA locally! I think some of the wording/states are missing. Roxana will provide updated screens/details a little later. Thanks!

Hi, Vivian. Thanks for the feedback. I'll wait for @rsgonzal screen.

PR #3304, opened 2022-03-22
Added green check mark circles for judge dashboard onboarding 

viviancan:


devkonrad (approved):
Looks good to me, Vivian! :)

app/views/judge/dashboards/rebrand/_side_nav_content.html.erb
```diff
@@ -1,5 +1,4 @@
-<div class="tw-tab-content">
-<!--  <a class="tw-active active-tab" id="profile-tab" href="#profile-tab-content">Consent Waiver</a>-->
+<div class="tw-tab-content list-none">
   <% if current_judge.consent_signed? %>
     <li>
```
shaunxp20, line 3: Some random notes/thoughts:

- Interesting this `<li>` isn't in an `<ul>` or `<ol>`, but I assume it works okay, so I guess it's okay! 😅 
- The conditional could be inside of the `<li>` then it wouldn't need repeating on line 11, though I'm guessing this was just a copy/paste of what we had before 🤔 
- And also interesting the checkmark circle on line 26 isn't inside of a `<li>` 🤔 

Anywho, doesn't seem like any of these things should hold this one up, just some #RandomThoughts! 😄

shaunxp20 (approved):
No blockers, looks good to me! :shipit: 🚢 🚀 🍾 🥳

app/views/judge/dashboards/rebrand/_side_nav_content.html.erb
viviancan: Great catch on the `<li>`s! Yes, it was a copy and paste/refactor from what we had before. I removed them because this side content is not actually a list so it really didn't make sense here. Thanks for the feedback Shaun 👍 🚀

Issue #3285, opened 2022-03-11
Remove survey answers from judge onboarding


rsgonzal:
Currently, survey answers are needed to complete onboarding as a judge. We want to take these away. These are irrelevant and not needed. Will introduce updated reasons and judge skills later![Screen Shot 2022-03-10 at 4.10.07 PM.png](https://images.zenhubusercontent.com/60527169cde2772242d0de95/420bc653-be66-4a90-9e56-a808f78e454d)

viviancan:
The survey answers were removed from the judge profile view, however there are still some backend components that need to be updated so that they are no longer required for onboarding to be complete.

rsgonzal:
wooo! i signed up and didn't see anything about survey answers and i was able to judge

PR #3308, opened 2022-03-23
Remove answers from judge onboarding

devkonrad:
Changes:
modified: app/models/judge_profile.rb
modified: app/views/judge/dashboards/_onboarding.en.html.erb
modified: app/views/judge/profiles/_menu.en.html.erb

Refs: https://github.com/Iridescent-CM/technovation-app/issues/3285

### DIFF
--- a/app/models/judge_profile.rb
+++ b/app/models/judge_profile.rb
@@ -226,7 +226,6 @@ class JudgeProfile < ActiveRecord::Base
     account &&
       account.email_confirmed? &&
         consent_signed? &&
-          training_completed? &&
-            survey_completed?
+          training_completed?
   end
 end
--- a/app/views/judge/dashboards/_onboarding.en.html.erb
+++ b/app/views/judge/dashboards/_onboarding.en.html.erb
@@ -42,21 +42,6 @@
     </li>
   <% end %>
 
-  <% if current_judge.survey_completed? %>
-    <li>
-      <%= web_icon(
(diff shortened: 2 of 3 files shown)


devkonrad:
> Changes: modified: app/models/judge_profile.rb modified: app/views/judge/dashboards/_onboarding.en.html.erb modified: app/views/judge/profiles/_menu.en.html.erb
> 
> Refs: #3285

@shaunxp20 and @viviancan, maybe we need to remove the route for Judge Answers and its controller. What you think about it?

viviancan:
> > Changes: modified: app/models/judge_profile.rb modified: app/views/judge/dashboards/_onboarding.en.html.erb modified: app/views/judge/profiles/_menu.en.html.erb
> > Refs: #3285
> 
> @shaunxp20 and @viviancan, maybe we need to remove the route for Judge Answers and its controller. What you think about it?

I agree that it should be removed, however with the deadline being so close, my concern is that more "sub issues" would come from it and we wouldn't have enough time to finish everything. Maybe it could be completed once all the higher priority tickets are completed!

viviancan:
Everything looks good to me. The only thing is that we are not using `app/views/judge/dashboards/_onboarding.en.html.erb` or `app/views/judge/profiles/_menu.en.html.erb` for the judge views. I think it's ok that you removed the survey answers from those files but eventually, they will be deleted!

Issue #3302, opened 2022-03-22
Update the judge edit profile with new rebranding


viviancan:
The judge edit your profile screen needs to be rebranded!
Judge -> "My Profile" > Basic Information > Change your basic profile details

viviancan:
![Screen Shot 2022-03-22 at 2.02.33 PM.png](https://images.zenhubusercontent.com/606b0f4fd13d53dff3d14fa5/6ef74c7e-51e4-4c67-8f18-726b031a8216)

dboyer:
Confirming the new location interrupt screen is now available on QA and working as expected

dboyer:
confirmed the profile editing is looking good on QA

PR #3312, opened 2022-03-23
[deprecated] Adjusting the profile form to fit the new brand

devkonrad:
Changes:
modified: app/views/profiles/_form.html.erb
modified: app/views/signups/_form.html.erb
modified: app/javascript/location/components/LocationForm.vue

Refs: https://github.com/Iridescent-CM/technovation-app/issues/3302

### DIFF
--- a/app/views/signups/_form.html.erb
+++ b/app/views/signups/_form.html.erb
@@ -2,19 +2,43 @@
   url ||= url_for(signup)
   title ||= t("views.signups.new.title.#{signup.scope_name}")
   admin ||= false
+  new_brand = ["student", "judge"].include? signup.scope_name
+  main_class = new_brand ? 
+    "container mx-auto flex flex-col lg:flex-row justify-around gap-6" : 
+    "grid grid--justify-space-around"
+  grid_class = new_brand ? 
+    "tab-content tw-active" : 
+    "grid__col-sm-8 panel" 
 %>
 
-<div class="grid grid--justify-space-around">
-  <div class="grid__col-sm-8 panel">
-    <h1>
-      <%= title %>
-      <% unless signup.persisted? %>
-        <%= t("views.signups.new.roles.#{signup.scope_name}") %>
-      <% end %>
-    </h1>
+<div class="<%= main_class %>">
+  <div class="<%= grid_class %>">
+    <% if !new_brand %>
+      <h1>
+        <%= title %>
+        <% unless signup.persisted? %>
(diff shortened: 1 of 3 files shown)


app/views/signups/_form.html.erb
```diff
@@ -2,20 +2,44 @@
   url ||= url_for(signup)
   title ||= t("views.signups.new.title.#{signup.scope_name}")
   admin ||= false
+  new_brand = ["student", "judge"].include? signup.scope_name
```
shaunxp20, line 5: Hmm, this is an interesting approach - I don't think we have any pages (or many if we do) where we mix both the new brand and the old brand. I'm not sure this is a good long term solution for working towards our rebrand efforts b/c eventually we're going to have to update this. 🤔 @viviancan Thoughts? Since you've done most of the rebranding work. 🤔

app/views/signups/_form.html.erb
```diff
@@ -81,6 +105,12 @@
           t('views.signups.form.create_account'),
           class: 'button' %>
       <% end %>
+      
+      <% if new_brand %>
+          </div>
+        </div>
+      <% end %>
```
shaunxp20, line 112: Hmm, I see it also creates some awkward conditions for the markup (which might make this page harder to read/work on in the future). 😕

app/views/signups/_form.html.erb
devkonrad: > Hmm, this is an interesting approach - I don't think we have any pages (or many if we do) where we mix both the new brand and the old brand. I'm not sure this is a good long term solution for working towards our rebrand efforts b/c eventually we're going to have to update this. thinking @viviancan Thoughts? Since you've done most of the rebranding work. thinking

Its a good point, i guess that maybe in a near future we can use a flag to separate old and new brand and just switch between specific partials.

app/views/signups/_form.html.erb
devkonrad: > Hmm, I see it also creates some awkward conditions for the markup (which might make this page harder to read/work on in the future). confused

Its true, i guess that we'll need to separete the templates for new brand in a near future.

devkonrad: > Hmm, I see it also creates some awkward conditions for the markup (which might make this page harder to read/work on in the future). confused

Its true, i guess that we'll need to separete the templates for new brand in a near future.

PR #3315, opened 2022-03-23
cleanup judge dashboard profile

viviancan:


app/views/judge/navigation/rebrand/_global_nav_links.html.erb
```diff
@@ -3,4 +3,10 @@
                "http://www.technovationchallenge.org/judging-rubric/",
                target: "_blank",
                class: "tw-sm-hidden"
-%>
\ No newline at end of file
+%>
+
+<% if current_judge.can_switch_to_mentor? %>
```
shaunxp20, line 8: Hmm, this conditional popped out to me - just wondering if this will work (not throw an error) for other profile types? Or if we need to check if the current account is a judge, then do this check? 🤔 😅

shaunxp20: Ah, just following up w/ my own comment, I see `can_switch_to_mentor?` is at the account level, so this should work fine. But I think it would be more accurate as `current_account.can_switch_to_mentor?` b/c chapter ambassadors can be mentors, even though (chapter ambassadors aren't using this layout); more of a technicality I  suppose. Anywho, looks like it will work fine as is! 👍

shaunxp20: Or hmm, following up again! 😅  I wonder if `current_judge` won't be defined for another profile type... like viewing this as a student or mentor... I will check this out and report back! 😄

shaunxp20: Oh, I see now! This file is limited to just judges! I see that in the directory path now! 🤦  😅 Okay, this should be fine! Nothing to see here! 😊 😊 😁

app/views/judge/navigation/rebrand/_dropdown_nav.html.erb
```diff
@@ -1,59 +1,55 @@
...
-  %>
-</li>
-<% if current_judge.can_switch_to_mentor? %>
+<ul>
+  <li class="md:hidden">
+    <%= link_to "Judging Rubric",
+                "http://www.technovationchallenge.org/judging-rubric/",
+                target: "_blank" %>
+  </li>
+  <li class="md:hidden"><a href="https://www.technovation.org/about/">About</a></li>
```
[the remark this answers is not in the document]
shaunxp20: Oh, darn it, just meant to leave this as a "comment". Anywho, def not blockers, just some minor things, so I'll mark this one approved! 👍

shaunxp20 (approved):
Nice cleanup work here! Looks good to me! 👍 🎉 🥳

app/views/judge/navigation/rebrand/_dropdown_nav.html.erb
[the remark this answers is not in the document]
viviancan: Thanks for the feedback, Shaun! I picked an approach and updated appropriately :)

app/views/judge/navigation/rebrand/_global_nav_links.html.erb
viviancan: 👍

PR #3316, opened 2022-03-24
Groundwork for new judging questions

shaunxp20:
This will lay the groundwork so that the judging questions are divided up by season and division, which means the questions for each season and division will live in separate files. This is the best I could come up with w/o doing some major rework - this PR adds new functionality and only adds two new lines of code to our existing code/flow, so it feels a little safer to me vs a drastic overhaul (maybe in the future!).

There is a new service I added called `JudgeQuestionsForSubmissionScore`, where you give it `submission_score` and it will return the questions for that submission factoring in its season and division. This should work for any submission in the 2020, 2021, or soon the 2022 season. I think we could go back further to get the questions for season before 2020 by going through the git history. 

I tried to leave some comments about the order of the changes, I'm happy to give a walk-though of the code if anyone wants too!

Also, I'm open to naming changes in this PR too, not sure I picked the best ones! 😅


app/services/judging/twenty_one/junior_questions.rb
```diff
@@ -0,0 +1,181 @@
+module Judging
+  module TwentyOne
+    class JuniorQuestions
```
shaunxp20, line 3: 2. This is the new way for judging questions - each season has the questions divided up by division. Which will hopefully make it easier to see/update questions for each division for each season.

Notice that the 'score` field has been removed; and the other minor change is that I reordered the fields, but basically this will return the same questions sans scores.

app/models/questions.rb
```diff
@@ -113,196 +113,7 @@ def as_json(*args, &block)
...
-      Question.new(
-        section: "overall",
-        field: :overall_2,
-        idx: 2,
-        text: "How well do you think our solution is thought out and can succeed?",
-        worth: 5,
-        score: submission_score.overall_2
-      )
-    ]
+    JudgeQuestionsForSubmissionScore.new(submission_score).call
```
shaunxp20, line 116: 3. Back in this file again, after removing the questions from this file, this is the new way of how we're going to get the questions (that will have the scores included too).

app/services/judge_questions_for_submission_score.rb
```diff
@@ -0,0 +1,50 @@
+class JudgeQuestionsForSubmissionScore
+  def initialize(submission_score)
+    @submission_score = submission_score
+    @season = submission_score.seasons.last
+    @division = submission_score.team_division_name
+    @questions = "Judging::#{season_module_name}::#{questions_class_name}".constantize.new.call
```
shaunxp20, line 6: 4. This is how the new questions are called, which will result in something like, `Judging:TwentyTwo:BeginnerQuestions.new.call`.

app/services/judge_questions_for_submission_score.rb
```diff
@@ -0,0 +1,50 @@
...
+    when "junior"
+      "JuniorQuestions"
+    when "senior"
+      "SeniorQuestions"
+    end
+  end
+
+  def questions_with_scores_populated
+    questions.map do |question|
+      question.score = submission_score.instance_eval(question.field.to_s)
```
shaunxp20, line 45: 5. `questions_with_scores_populated` this is where the scores will get populated, specifically this line^, which will set the `score` field to the` submission_score` field, this line will produce a result like the example from 1. something like - `score: submission_score.ideation_1`. In a nutshell, I made the questions standalone, and this service (JugeQuestionsForSubmissionScore) will attach the scores (which will retain the old functionality).

app/models/question.rb
```diff
@@ -1,5 +1,6 @@
 class Question
   attr_reader :section, :idx, :text, :worth, :score, :field
+  attr_writer :score
```
shaunxp20, line 3: 6. This line is needed for step 5.

app/services/judging/twenty_twenty/junior_questions.rb
```diff
@@ -0,0 +1,128 @@
+module Judging
+  module TwentyTwenty
+    class JuniorQuestions
```
shaunxp20, line 3: (This comment got outdated, just adding it back to have the full context in the diff view.)

2. This is the new way for judging questions - each season has the questions divided up by division. Which will hopefully make it easier to see/update questions for each division for each season.

Notice that the 'score` field has been removed; and the other minor change is that I reordered the fields, but basically this will return the same questions sans scores.

app/services/judge_questions_for_submission_score.rb
shaunxp20: Following up w/ this, you'll see the season are now more specific - `TwentyTwentyTwo` vs `TwentyTwo`, hopefully that makes it a bit clearer. 🤷 😅

PR #3318, opened 2022-03-24
3287 registration referred by bug

viviancan:
There were 2 issues with this bug!
First `v-model="referralOption"` and `referralOption: 0` was setting the initial value of the select to Friend based on vue formulate docs (https://vueformulate.com/guide/inputs/#setting-an-initial-value). Changing the value of referralOption  to `null` allowed for the referredBy param to be set to null.

The second issue was in the registration controller where the `referred_by` attribute was being set `registration_params[:referredBy].to_i`. When a null/nil value came through as part of the registration request `nil.to_i` would result in the value of `0` which is equal to a Friend.


shaunxp20 (approved):
Great work on this bug Vivian! 💯 🥇 🙌 Looks good to me! Thanks for taking care of this! 🎉

PR #3321, opened 2022-03-25
Adjusting the profile form to fit the new brand

devkonrad:
Creates a "rebranded?" that allows to simple switch between old and new profiles style.

Changes:
modified: app/views/location_details/show.en.html.erb
modified: app/views/profiles/_form.html.erb
modified: app/views/signups/_form.html.erb
modified: app/views/signups/_form_content.html.erb
modified: app/views/signups/rebranded/_form.html.erb

Refs: https://github.com/Iridescent-CM/technovation-app/issues/3302
PR Discussion: https://github.com/Iridescent-CM/technovation-app/pull/3312

### DIFF
--- a/app/models/admin_profile.rb
+++ b/app/models/admin_profile.rb
@@ -15,6 +15,10 @@ class AdminProfile < ActiveRecord::Base
     "admin"
   end
 
+  def rebranded?
+    false
+  end
+
   def method_missing(method_name, *args)
     begin
       account.public_send(method_name, *args)
--- a/app/models/chapter_ambassador_profile.rb
+++ b/app/models/chapter_ambassador_profile.rb
@@ -110,6 +110,10 @@ class ChapterAmbassadorProfile < ActiveRecord::Base
     secondary_regions
   end
 
+  def rebranded?
+    false
+  end
+  
   def onboarded?
     account.email_confirmed? and
       approved? and
--- a/app/models/judge_profile.rb
+++ b/app/models/judge_profile.rb
@@ -222,6 +222,10 @@ class JudgeProfile < ActiveRecord::Base
     "judge"
   end
 
+  def rebranded?
+    true
+  end
+
   def can_be_marked_onboarded?
     account &&
       account.email_confirmed? &&
--- a/app/models/mentor_profile.rb
+++ b/app/models/mentor_profile.rb
@@ -345,6 +345,10 @@ class MentorProfile < ActiveRecord::Base
     teams.include?(team)
   end
 
+  def rebranded?
+    false
+  end
+
   def onboarding?
     not onboarded?
   end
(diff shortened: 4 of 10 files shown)


viviancan (commented):
Hey Alex! Thanks for taking a look at this. I haven't had a chance to pull down locally yet. This does look like a logical approach to the rebranding work. The only concern I have is how we would move forward with this since we have used a different approach for quite a bit of the other rebranding pieces. 

Tagging @shaunxp20 and @dboyer for these follow-up questions!
Since we have done a ton of rebranding already using the first approach, what would the rest of the rebranding work look like? Would we switch to this approach for the rest of the rebranding? Also what would the clean-up tasks look like when we have finished rebranding everything? Wanted to drop this comment in before the weekend! I think we will have time to finalize/deploy on Monday 3/28/22. Have a great weekend everyone 🌻 💯 🚀

devkonrad:
> Hey Alex! Thanks for taking a look at this. I haven't had a chance to pull down locally yet. This does look like a logical approach to the rebranding work. The only concern I have is how we would move forward with this since we have used a different approach for quite a bit of the other rebranding pieces.
> 
> Tagging @shaunxp20 and @dboyer for these follow-up questions! Since we have done a ton of rebranding already using the first approach, what would the rest of the rebranding work look like? Would we switch to this approach for the rest of the rebranding? Also what would the clean-up tasks look like when we have finished rebranding everything? Wanted to drop this comment in before the weekend! I think we will have time to finalize/deploy on Monday 3/28/22. Have a great weekend everyone sunflower 100 rocket



> Hey Alex! Thanks for taking a look at this. I haven't had a chance to pull down locally yet. This does look like a logical approach to the rebranding work. The only concern I have is how we would move forward with this since we have used a different approach for quite a bit of the other rebranding pieces.
> 
> Tagging @shaunxp20 and @dboyer for these follow-up questions! Since we have done a ton of rebranding already using the first approach, what would the rest of the rebranding work look like? Would we switch to this approach for the rest of the rebranding? Also what would the clean-up tasks look like when we have finished rebranding everything? Wanted to drop this comment in before the weekend! I think we will have time to finalize/deploy on Monday 3/28/22. Have a great weekend everyone sunflower 100 rocket

Hi, Vivian! Thank you for the feedback. Good points! The purpose of this approach is afford a simple way to switch between the current and the new layout, also its a intermediary path for the fully rebrand (when we will need to remove the switch flag and use the content of the rebrand folder as the pattern). I guess that nothing wich have be done until now need to be rewrited, therefore the two ways are temporary and will be merged at the end of the rebrand, but its just my current thought about the case, maybe there are some other fatctors that i am not seeing. Have a great weekend! :)

viviancan (approved):
Pulled down and tested locally! Looks good to me. Thanks for this Alex!

Also following up from the standup this morning 3/28/22,  we will merge this to QA but continue using old approach for rebrand work.

Issue #328, opened 2016-05-10
Prevent judges to assign themselves to events outside of their region


millafalconi:
## Context

Currently, all judges can see all events of the current season.

We should show only events of the judge's region.
We want this behavior only during the quarterfinals.

PR #3325, opened 2022-03-29
Hides mentor age on public profile

devkonrad:
Avoid to display mentors age when its accessed from mentors show profile.

Changes:
modified: app/views/profiles/_privileged_show.html.erb

Refs: https://github.com/Iridescent-CM/technovation-app/issues/3219
Discussion: https://github.com/Iridescent-CM/technovation-app/pull/328

### DIFF
--- a/app/views/profiles/_privileged_show.html.erb
+++ b/app/views/profiles/_privileged_show.html.erb
@@ -41,8 +41,10 @@
           <dt><%= t('models.account.email') %></dt>
           <dd><%= mail_to account.email %></dd>
 
-          <dt><%= t('models.account.age') %></dt>
-          <dd><%= account.age %></dd>
+          <% if !account.mentor_profile.present? %>
+            <dt><%= t('models.account.age') %></dt>
+            <dd><%= account.age %></dd>
+          <% end %>
 
           <% unless account.student_profile.present? %>
             <dt><%= t('models.account.gender') %></dt>
--- a/app/views/judge/profiles/rebrand/_basic.en.html.erb
+++ b/app/views/judge/profiles/rebrand/_basic.en.html.erb
@@ -7,7 +7,7 @@
       <dt>Gender identity</dt>
       <dd><%= current_judge.gender || "Not specified" %></dd>
 
-      <dt>Age</dt>
+      <dt>Birthdate</dt>
       <dd><%= current_judge.date_of_birth.strftime("%B %e, %Y") %></dd>
 
       <dt>Company name</dt>


app/views/profiles/_privileged_show.html.erb
```diff
@@ -41,8 +41,10 @@
           <dt><%= t('models.account.email') %></dt>
           <dd><%= mail_to account.email %></dd>
 
-          <dt><%= t('models.account.age') %></dt>
-          <dd><%= account.age %></dd>
+          <% if !account.mentor_profile.present? %>
```
shaunxp20, line 44: Hmm, just curious if you have tested this with an account that has both a mentor and judge profile? Maybe it's okay, but just from looking at this code, I'm wondering if it would hide the age for judges for these cases? 🤔 😄

devkonrad: Thank you, for the tip @shaunxp20 ! I just tested with and account with mentor profile. I will add a Judge profile and test it. :)

viviancan: Following up on this!
I logged in as a mentor and then switched to a judge profile (for the same account) and did not see my age on either.

I did, however, catch a typo in `app/views/judge/profiles/rebrand/_basic.en.html.erb` line 10 says `age` and should be `Birthday` or `Date of Birth` or should this value be removed completely? @dboyer 
Screen Shot 2022-04-11 at 8 32 37 PM

dboyer:
We do not need to show age on the judge profile. It can be removed. 

The only place we have to have mentor and judge age is as a column we can add on the admin pages as shown in this screenshot. If it's available there, we're fine. 

Screen Shot 2022-04-12 at 9 52 49 AM

viviancan (changes requested):
Thanks for the work on this one, Alex! Please make the heading change on the Judge profile from `Age` to `Birthdate` as referenced in this comment 
https://github.com/Iridescent-CM/technovation-app/pull/3325#issuecomment-1096807479
Ty!!

devkonrad:
> Thanks for the work on this one, Alex! Please make the heading change on the Judge profile from `Age` to `Birthdate` as referenced in this comment [#3325 (comment)](https://github.com/Iridescent-CM/technovation-app/pull/3325#issuecomment-1096807479) Ty!!

Changes done! :)

app/views/profiles/_privileged_show.html.erb
shaunxp20: > Following up on this! I logged in as a mentor and then switched to a judge profile (for the same account) and did not see my age on either.

Ah, I guess this makes sense - the requirements were to just to hide the age for mentors, but we're hiding it for judges as well - it doesn't make sense to hide it for one but not the other since since they can just toggle between the two profile types.

shaunxp20 (approved):
Thanks Alex! 👍 🙌

PR #3328, opened 2022-03-29
Fixing column size location screen inline

devkonrad:
Fixing column size location screen inline

For some reason, the current vue css style  notation causes no effect on QA environment, alhough the same works fine locally. May an alterantive can be use the inline html notation until discover whats goes wrong here.


shaun-technovation:
Closing this PR, I created a new label `Unmerged PR` so it's easy to find/reference this PR in the future if we need to. 👍

PR #3337, opened 2022-04-05
Update judge questions

shaunxp20:
This PR will:
- add the 2022 judging questions
- add the new judging question fields to the database
- allow the new judging question fields to be updated in the controller


spec/features/judge/scoring_submissions_spec.rb
```diff
@@ -1,6 +1,6 @@
 require "rails_helper"
 
-feature "scoring submissions", js: true do
+xfeature "scoring submissions", js: true do
```
shaunxp20, line 3: I commented out this feature spec because it very heavily relies on the front-end/markup - I thought it made more sense to get the front-end in place and then fix this spec accordingly.

Issue #3297, opened 2022-03-16
Show new submission fields in submission details for admin and ChAChAs


rsgonzal:
- learning journey
- Submission type
- user adoption plan for junior 
- Additional Questions

![Screen Shot 2022-01-24 at 9.51.18 AM.png](https://images.zenhubusercontent.com/60527169cde2772242d0de95/322d2bed-89ed-494e-9cfb-870468c3f176)

devkonrad:
Hi @shaunxp20 and @viviancan , I am working on this issue and i have some doubts:
The information for:
- submission_type 
- learning_journey
I can get it from Team Submission table, but i couldnt find:
- user adoption plan for junior (I found business plan, is that info?)
- Additional Questions (Where can i get these?)

Thank you!

dboyer:
@devkonrad The additional questions are optional information teams can provide when submitting an application. There are three questions:

- Does your submission use artificial intelligence?
- Does your submission help solve climate change?
- Does your submission help solve a health problem?

Each question is a yes/no answer. If the user selects "yes," they have the option to fill in an open text field with information answering "Please explain how."

The answers should be saved in whatever file the other fields are saved connected to the submitted application. Do you not see those fields there?

In terms of user adoption plan, people in the junior division add that info. The senior division adds the business plan.

devkonrad:
Thank you, @dboyer ! Its more clear :)!

devkonrad:
> If needed, the full lists of questions asked in submissions are [available here](https://docs.google.com/spreadsheets/d/1xscD_VtTQrv0ub_jC21qWLogZMVn_3vnvrUK_18-LKE/edit#gid=0). I've limited access so you'll need to be logged in with your virtualmind account

Thank you very much, @Deb!

devkonrad:
> > If needed, the full lists of questions asked in submissions are [available here](https://docs.google.com/spreadsheets/d/1xscD_VtTQrv0ub_jC21qWLogZMVn_3vnvrUK_18-LKE/edit#gid=0). I've limited access so you'll need to be logged in with your virtualmind account
> 
> Thank you very much, @dboyer !

devkonrad:
Pull Request created for this issue: https://github.com/Iridescent-CM/technovation-app/pull/3338

shaunxp20:
> The additional questions are optional information teams can provide when submitting an application. There are three questions:
> 
>     * Does your submission use artificial intelligence?
> 
[...]

PR #3338, opened 2022-04-07
Adds new submission fields to admin details

devkonrad:
Adds aditional fields to submission detail on admin interface.

Changes:
modified: app/views/admin/team_submissions/show.html.erb

Refs: https://github.com/Iridescent-CM/technovation-app/issues/3297

### DIFF
--- a/app/models/team_submission.rb
+++ b/app/models/team_submission.rb
@@ -577,17 +577,16 @@ class TeamSubmission < ActiveRecord::Base
   def video_url_root(video_type)
     VideoUrl.new(video_link_for(video_type)).root
   end
-
-  def uses_ai?
-    ai == true
-  end
-
-  def solves_climate_change?
-    climate_change == true
-  end
-
-  def solves_health_problem?
-    game == true
+  
+  def question_response(question)
+    question_key = question == "health_problem" ? :game : question.to_sym 
+    if self[question_key].nil?
+      "-"
+    elsif self[question_key]
+      "Yes"
+    else
+      "No"
+    end
   end
 
   private
--- a/app/models/team_submission.rb
+++ b/app/models/team_submission.rb
@@ -578,6 +578,18 @@ class TeamSubmission < ActiveRecord::Base
     VideoUrl.new(video_link_for(video_type)).root
   end
 
+  def uses_ai?
+    ai == true
+  end
+
+  def solves_climate_change?
+    climate_change == true
+  end
+
+  def solves_health_problem?
+    game == true
+  end
+
   private
 
   def team_name_and_app_name
--- a/app/views/admin/team_submissions/show.html.erb
+++ b/app/views/admin/team_submissions/show.html.erb
@@ -118,6 +118,20 @@
         </p>
       <% end %>
 
+      <% if @team_submission.junior_division? %>
+        <p>
+          User Adoption Plan:
+          
+          <% if @team_submission.business_plan_url %>
+            <%= link_to @team_submission.business_plan_identifier,
+              @team_submission.business_plan_url,
+              target: :_blank %>
+          <% else %>
+            <span>blank</span>
+          <% end %>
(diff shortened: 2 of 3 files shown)


app/models/team_submission.rb
```diff
@@ -575,6 +575,18 @@ def video_url_root(video_type)
...
+  def uses_ai?
+    ai == true
+  end
+
+  def solves_climate_change?
+    climate_change == true
+  end
+
+  def solves_health_problem?
+    game == true
```
devkonrad, line 587: As i comented the other day, the flag that controls the "solves health problem" is called "game"...

app/models/team_submission.rb
```diff
@@ -575,6 +575,18 @@ def video_url_root(video_type)
     VideoUrl.new(video_link_for(video_type)).root
   end
 
+  def uses_ai?
+    ai == true
+  end
```
shaunxp20, line 580: This method and the one below aren't necessary. `uses_ai?` vs just using `ai?` or `solves_climiate_change?` vs `climate_change?` doesn't provide a lot of extra value, in my opinion. 😄

app/models/team_submission.rb
```diff
@@ -575,6 +575,18 @@ def video_url_root(video_type)
...
+
+  def solves_climate_change?
+    climate_change == true
+  end
+
+  def solves_health_problem?
```
shaunxp20, line 586: I like this, it's a nice wrapper method! 👍

app/views/admin/team_submissions/show.html.erb
```diff
@@ -162,6 +176,49 @@
         <% end %>
       </p>
 
+      <p>
+        Learning Journey:
+        <% if !@team_submission.learning_journey.nil? %>
```
shaunxp20, line 181: This might make more sense as `.present?` instead of checking "not nil".

app/views/admin/team_submissions/show.html.erb
```diff
@@ -162,6 +176,49 @@
...
+          <%= @team_submission.learning_journey %>
+        <% else %>
+          blank
+        <% end %>
+      </p>
+
+      <p>
+        Submission Type:
+        <% if !@team_submission.submission_type.nil? %>
```
shaunxp20, line 190: Hmm, I'm pretty sure `submission_type` is a required field, so every submission should have one. Are you seeing differently? I'm just off of the top of my head.

app/views/admin/team_submissions/show.html.erb
```diff
@@ -162,6 +176,49 @@
...
+          <%= @team_submission.submission_type %>
+        <% else %>
+          blank
+        <% end %>
+      </p>
+
+      <h5>Additional Questions</h5>
+      <p>
+        Does your submission use artificial intelligence? 
+        <%= @team_submission.uses_ai? ? "Yes" : "No"  %>
```
shaunxp20, line 200: Looks like a lot these ERB tags have an extra space at the end of the closing tag. 🔍 👀 😄

app/views/admin/team_submissions/show.html.erb
devkonrad: Well, i saw other cases on the codebase testing the submittion_type for blank. Sou i put this test here just for ensure about these case. Like below example:

Captura de tela de 2022-04-11 11-46-20

app/models/team_submission.rb
devkonrad: Good, catch! :) I aggre!

viviancan (approved):
Thanks for the work on this one @devkonrad! Feel free to merge into QA! 🎉 👍

devkonrad:
> Thanks for the work on this one @devkonrad! Feel free to merge into QA! tada +1

Thanks, @viviancan! I just merged it into QA!

Issue #3340, opened 2022-04-11
Team page will not open for mentor


dboyer:
A mentor is reporting that when they click on a certain team from their mentor dashboard they see an error page rather than the team details. I was able to reproduce this issue. Please contact me for the specific mentor and team names connected to the issue.

1. Go to https://my.technovationchallenge.org/mentor/dashboard#/submission
2. Click on the team https://my.technovationchallenge.org/mentor/teams/28740
3. Receive our standard error page
<img width="739" alt="Screen Shot 2022-04-11 at 11 22 05 AM" src="https://user-images.githubusercontent.com/516764/162772505-b5d7b170-8a9e-4480-809c-c3ff9e73379d.png">
4. Return to mentor dashboard and click on a different team https://my.technovationchallenge.org/mentor/teams/28745
5. Page loads correctly

I repeated the same process with the other mentor for the team and got the same error when loading the team's page, leading me to believe it's an issue with that particular team profile. The team name has an @ in it so I thought it might be an issue with a special character. I created a team with @ in the name on QA and was able to successfully open it from the mentor dashboard.

shaunxp20:
This was an interesting one. There was a pending invite for this team and an error was occurring when it tried to render the "delete" link. This link broke during the [Rails 6 upgrade](https://github.com/Iridescent-CM/technovation-app/commit/02a55fa3aadb0ae2baa74120b04c7e2db6cda9b5#diff-231db6a4ec3089799abdb58aa74da85709139633181ab4c8c89b7ea9d313153d).

I updated the link and it's working for me locally, so this fix should work for this team as well.

PR #3341, opened 2022-04-11
Update "delete" pending invite links

shaunxp20:
The "delete" pending invite link was throwing an error:
`ActionView::Template::Error (undefined method `mentor_invites_path' for #<ActionView::Base:0x00000000079478>`

Because the route it's trying to use doesn't exist, the "delete" link was incorrectly updated during the Rails 6 upgrade.

The update in this PR will resolve the links to use one of these two existing routes:
- student_team_member_invite_path
- mentor_team_member_invite_path

Refs: #3340

### DIFF
--- a/app/views/invites/_unregistered_invitee.en.html.erb
+++ b/app/views/invites/_unregistered_invitee.en.html.erb
@@ -11,7 +11,10 @@
       <small>Unregistered</small><br />
 
       <%= link_to "delete this invitation",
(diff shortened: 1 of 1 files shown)


devkonrad (approved):
Great! Good solution, shaun!

viviancan (approved):
Great catch, Shaun! 👍

Issue #3339, opened 2022-04-08
Media consent not displaying for chapter ambassadors


dboyer:
We have a "chapter ambassador" user group who should be able to see the media consent status for users in their area. Currently, they can add media consent as a column on the participants admin page. 

<img width="666" alt="Screen Shot 2022-04-08 at 3 34 07 PM" src="https://user-images.githubusercontent.com/516764/162512138-95a79a09-34fa-4c70-bcc2-6107ff9a1029.png">

However, if a chapter ambassador clicks through to the participant's individual page, the media consent section is not visible. Chapter ambassadors should be able to view their students' media consent. They **should not** be able to edit it.

<img width="1182" alt="Screen Shot 2022-04-08 at 3 40 28 PM" src="https://user-images.githubusercontent.com/516764/162515410-78eddbed-17fe-435a-af0c-60800584f31c.png">

dboyer:
As a reference, chapter ambassadors should be able to view these statuses on the page for each user. They SHOULD NOT be able to set the status of the media consent, however.

<img width="671" alt="Screen Shot 2022-04-15 at 9 26 32 AM" src="https://user-images.githubusercontent.com/516764/163576286-25357f25-10a4-414e-8238-a06d332a9ad7.png">

devkonrad:
> As a reference, chapter ambassadors should be able to view these statuses on the page for each user. They SHOULD NOT be able to set the status of the media consent, however.
> 
> <img alt="Screen Shot 2022-04-15 at 9 26 32 AM" width="671" src="https://user-images.githubusercontent.com/516764/163576286-25357f25-10a4-414e-8238-a06d332a9ad7.png">

Thats correct, Deb! They are allowed just to see the media consent status.
I just publishe a Pull Request that implements this feature: https://github.com/Iridescent-CM/technovation-app/pull/3347

dboyer:
Confirmed this is displaying as expected on staging.

PR #3347, opened 2022-04-15
Allow ambassadors do verify student media consent

devkonrad:
Allow ambassadors only check the student media consent status, but NOT EDIT it.

Changes:
modified: app/models/student_profile.rb
modified: app/views/admin/participants/_student_debugging.html.erb
modified: app/views/admin/participants/_student_media_consent.html.erb
modified: app/views/chapter_ambassador/participants/show.en.html.erb

Refs: https://github.com/Iridescent-CM/technovation-app/issues/3339

### DIFF
--- a/app/views/admin/participants/_student_media_consent.html.erb
+++ b/app/views/admin/participants/_student_media_consent.html.erb
@@ -1,0 +1,47 @@
+<% if profile %>
+    <div class="panel">
+        <% media_consent = profile.media_consent %>
+        <% if media_consent&.consent_provided.nil? %>
+            <h5>Has not completed media consent form</h5>
+            <% if admin  %>
+                <%= link_to "GRANT permission",
+                admin_paper_media_consent_path(
+                    id: profile.id,
+                    consent: true
+                ),
+                class: "button button--secondary",
+                data: {
+                    method: :post,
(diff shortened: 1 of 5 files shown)


app/views/admin/participants/_student_media_consent.html.erb
```diff
@@ -0,0 +1,47 @@
...
+            </h5>
+            <% else %>
+            <h5>
+                Did NOT AGREE to media consent via 
+                <%= media_consent.on_file? ? "paper form" : "online form" %>
+            </h5>
+            <% end %>
+        <% end %>
+    </div>
+<% end %>
```
devkonrad, line 47: In order to avoid code duplication, I created this partial which is shared between admin and ambassador profiles.

viviancan: I think the indentations are off in this file!

devkonrad: Thanks, vivian! I'll verify it! :)

viviancan (approved):
Please verify the indentations, but besides that, pulled down locally and was able to verify the expected functionality for both admin and Cha! 👍 Thx Alex

PR #3348, opened 2022-04-18
Add new submission type constants

shaunxp20:
This will create two constants:
- `MOBILE_APP_SUBMISSION_TYPE`
- `AI_PROJECT_SUBMISSION_TYPE`

This will be instead of using the magic strings "Mobile App" and "AI Project". It will also keep the `submission_type` field for the judging questions consistent.

### DIFF
--- a/app/models/team_submission.rb
+++ b/app/models/team_submission.rb
@@ -4,9 +4,12 @@ class TeamSubmission < ActiveRecord::Base
   MAX_SCREENSHOTS_ALLOWED = 6
   PARTICIPATION_MINIMUM_PERCENT = 50
 
+  MOBILE_APP_SUBMISSION_TYPE = "Mobile App"
+  AI_PROJECT_SUBMISSION_TYPE = "AI Project"
+
   SUBMISSION_TYPES_ENUM = {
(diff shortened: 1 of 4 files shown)


app/models/team_submission.rb
```diff
@@ -4,9 +4,12 @@ class TeamSubmission < ActiveRecord::Base
...
   PARTICIPATION_MINIMUM_PERCENT = 50
 
+  MOBILE_APP_SUBMISSION_TYPE = "Mobile App"
+  AI_PROJECT_SUBMISSION_TYPE = "AI Project"
+
   SUBMISSION_TYPES_ENUM = {
-    "Mobile App" => 0,
-    "AI Project" => 1
+    MOBILE_APP_SUBMISSION_TYPE => 0,
+    AI_PROJECT_SUBMISSION_TYPE => 1
```
viviancan, line 12: I really like this change! I think it will go a long way :)

viviancan (approved):
The changes look good to me. I think it will really help long-term with the judging flow and just basic submission usage in general. Can't wait to see all of this come together!

PR #3349, opened 2022-04-18
Filter Demo questions based on submission type

shaunxp20:
There are six possible questions for the Demo section - three are for an AI Project and three are for a Mobile App project.

Prior to this update the back-end would have returned all six questions to the front-end.

With this update the back-end will only return the necessary questions based on whether the submission type is an AI Project or a Mobile App project. This will make things simpler for the front-end.

### DIFF
--- a/app/services/judge_questions_for_submission_score.rb
+++ b/app/services/judge_questions_for_submission_score.rb
@@ -1,6 +1,7 @@
 class JudgeQuestionsForSubmissionScore
   def initialize(submission_score)
     @submission_score = submission_score
+    @submission_type = submission_score.team_submission_submission_type
     @season = submission_score.seasons.last
     @division = submission_score.team_division_name
     @questions = "Judging::#{season_module_name}::#{questions_class_name}".constantize.new.call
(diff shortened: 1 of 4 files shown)


app/services/judge_questions_for_submission_score.rb
```diff
@@ -43,10 +44,16 @@ def questions_class_name
...
   def questions_with_scores_populated
-    questions.map do |question|
+    filtered_questions.map do |question|
       question.score = submission_score.instance_eval(question.field.to_s)
 
       question
     end
   end
+
+  def filtered_questions
```
shaunxp20, line 54: A lot of the changes from above are just getting `submission_type` setup/accessible. The method below is where the core functionality change is - removing questions that don't match w/ the submission's `submission_type`.

spec/services/judge_questions_for_submission_score_spec.rb
```diff
@@ -5,12 +5,14 @@
     JudgeQuestionsForSubmissionScore.new(submission_score).call
   }
   let(:submission_score) { instance_double(SubmissionScore) }
+  let(:submission_type) { "Mobile App" }
```
viviancan, line 8: Nice testing coverage! 👍

viviancan (approved):
I like this change, more of the heavy lifting in the backend! It seems like it will allow the front end changes to be much more streamlined.

Issue #3350, opened 2022-04-18
Running list of judging notes


shaunxp20:

viviancan:
"Project Name & Description" is getting clipped in sidebar - VC
PR: [#3354](https://github.com/Iridescent-CM/technovation-app/pull/3354)

viviancan:
Styles are now being rendered on QA appropriately!

viviancan:
Re: "Title" issue in Question Section
I think it has to do with state management. Initially the value is undefined and then after a few seconds there a value and the title is rendered properly.

![Screen Shot 2022-04-21 at 9.09.03 AM.png](https://images.zenhubusercontent.com/606b0f4fd13d53dff3d14fa5/1a03a7e8-bbbe-4eb4-b7b9-6eea317af858)

viviancan:
There is a weird bug on the review score page with the exclamation visual that is shown to signify there is something wrong with the section score. If I go through the score stepper like normal by clicking the next button or by even clicking the section link in the score stepper, the exclamation visual displays as expected. However, if I refresh the page the exclamation visual will display on every section, even if there is not anything missing from that section. Then if I click back in the section and go back to the score stepper, the exclamation visual displays as expected. See the video attached. Note** the exclamation always stays on the first section because those comments are not currently saving.

https://images.zenhubusercontent.com/29868115/24967e26-55aa-4dc9-aa1b-5bcc7fd1b672/review_score_exclamation_bug2.mov

dboyer:
Nothing that this task is not completed yet on QA:

- Update text on Thank you screen to include the time "...make edits as you need through June 1 at 5pm (Pacific Daylight Time)."

viviancan:
> Nothing that this task is not completed yet on QA:
> 
> * Update text on Thank you screen to include the time "...make edits as you need through June 1 at 5pm (Pacific Daylight Time)."

I will take care of this, this morning!

PR #3353, opened 2022-04-19
Hooking up the juding score stepper

shaunxp20:
This will add very basic/rough functionality to get the judging screens working. There is still a bit of work to do (I added some more notes/todos to #3350), but this will at least give us the basic wiring up functionality and keep us moving forward.

### DIFF
--- a/app/views/judge/scores/new.en.html.erb
+++ b/app/views/judge/scores/new.en.html.erb
@@ -11,8 +11,20 @@
   >
   </div>
 
-  <div v-else class="flex flex-col lg:flex-row lg:gap-x-4 tw-wrapper">
-    <div id="judge-scores-stepper"></div>
-    <router-view class="w-full"></router-view>
-  </div>
+  <template v-else>
+    <% provide :judge_score_stepper do %>
+      <div class="flex flex-col lg:flex-row lg:gap-x-4 tw-wrapper">
+        <div id="judge-scores-stepper"></div>
+
+        <div class="w-full">
+        <% end %>
+
+        <router-view></router-view>
+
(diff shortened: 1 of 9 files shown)


app/views/judge/scores/new.en.html.erb
```diff
@@ -11,8 +11,20 @@
...
+        <div class="w-full">
+        <% end %>
+
+        <router-view></router-view>
+
+        </div>
+      </div>
+    </div>
+  </template>
+</div>
```
viviancan, line 28: I think there are a few extra `<div>`s here!

shaunxp20: Yes! That is correct and intended! 😊😊😅  

This is what I had to do to make the layout display correctly! 😅😅😅

If I remove one/all of them the footer won't display correctly. I'm not sure why, we might have some `<div>` tags misaligned somewhere. I'll add this to #3350 for followup work to figure this out. 😬😁😅

devkonrad (approved):
I guess that the changes make sense and take care of great part of the  issue https://github.com/Iridescent-CM/technovation-app/issues/3350

app/views/judge/scores/new.en.html.erb
viviancan: This formatting issue was added to #3350 ! We will revisit this in the next step!

viviancan (approved):
Woo! thanks for merging all the work and taking care of the score stepper! Very excited to see everything come together! 🎉 

For anyone reading this in the future, we have a few follow up items to tackle which have been added to #3350

PR #3356, opened 2022-04-20
Update prev/next button text

shaunxp20:
This will allow the text for the prev/next buttons to be customized.

Two new props were added:
- `prevButtonText`
- `nextButtonText`

These props can be passed to `GenericJudgingContainer` which will pass them to `QuestionScore` where they will be used.


viviancan (approved):
Great update here, Shaun!

Issue #3196, opened 2021-12-15
Add fields to more columns in "submission" section of admin


rsgonzal:
Both ChA ChAs and admin see this
New Sections:
- Learning Journey
- Submission Type (ai/ mobile)
- Change development platform just to that for mobile, would be blank for AI
- thunkable project url
![Screen Shot 2021-12-14 at 6.16.41 PM.png](https://images.zenhubusercontent.com/60527169cde2772242d0de95/bcd7e083-fdd9-4309-b740-a053e3b23283)

nickiredescent:
Checking that "User Adoption Plan" (the Junior "business plan") is somehow included here.

rsgonzal:
@nickiredescent where did the thunkable link show up before? i'm not seeing it

nickiredescent:
woah I don't know lol. [Here's the data pull from last year ](https://docs.google.com/spreadsheets/d/1IMZGck2RozlrPsZwHSu_BEr5OFmhbC9xyIjbTMIkNps/edit#gid=1048604029&range=K1)
Master Tab, Column K. It's called "thunkable_project_irl" but I don't know where it came from cause I don't see it on the platform either.

dboyer:
@nickiredescent Can you email that google doc with the data pull to Rox so we don't have a link to personal data on Github?

nickiredescent:
ah right, sorry forgot about that. Emailing now

devkonrad:
**Hi, a created a Pull Request for this issue:** https://github.com/Iridescent-CM/technovation-app/pull/3334

shaunxp20:
From the requirements:
> * Change development platform just to that for mobile, would be blank for AI

I'm interpreting this as for submission type "mobile app" then display whatever was selected for the development platform, and if it's an "AI" submission type don't display anything (b/c development platform doesn't apply to AI). Is that correct? 😅

dboyer:
Yes, that is how I interpret it as well.

viviancan:
Following up here -
Pulled this branch and was able to see the new fields

<img width="1426" alt="Screen Shot 2022-04-15 at 4 27 45 PM" src="https://user-images.githubusercontent.com/29210380/163634487-62cca7b5-36fd-4266-8831-93ca836eb5cc.png">

viviancan:
> Checking that "User Adoption Plan" (the Junior "business plan") is somehow included here.

Is this the same as 'business plan url'

dboyer:
[...]

PR #3357, opened 2022-04-20
Separates the "business plan" from "adoption plan" url on search

devkonrad:
This fix will separate business plan url from adoption plan url on Team Submission searchs.

Refs: https://github.com/Iridescent-CM/technovation-app/issues/3196

### DIFF
--- a/app/data_grids/submissions_grid.rb
+++ b/app/data_grids/submissions_grid.rb
@@ -33,7 +33,15 @@ class SubmissionsGrid
   column :app_inventor_app_name
   column :app_inventor_gmail
   column :source_code_url
-  column :business_plan_url
+  
+  column :business_plan_url do
+    !team.junior? ? self.business_plan_url : "-"
+  end
+
+  column :adoption_plan_url do
+    team.junior? ? self.business_plan_url : "-"
+  end 
+
   column :pitch_presentation_url
 
   column :team_name,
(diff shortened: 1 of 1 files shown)


app/data_grids/submissions_grid.rb
```diff
@@ -33,7 +33,15 @@ class SubmissionsGrid
   column :app_inventor_app_name
   column :app_inventor_gmail
   column :source_code_url
-  column :business_plan_url
+  
+  column :business_plan_url do
+    !team.junior? ? self.business_plan_url : "-"
```
shaunxp20, line 38: I don't think this logic is correct - the beginners division isn't "junior" (and beginners don't have a business plan), I think we want this to be `team.senior?` since only senior teams will have a business plan. 😄

devkonrad: > I don't think this logic is correct - the beginners division isn't "junior" (and beginners don't have a business plan), I think we want this to be `team.senior?` since only senior teams will have a business plan. smile

Great, @shaunxp20 ! I will correct it!

viviancan (approved):
👍 Looks good!

viviancan:
Confirming that I could see both Business Plan Url and Adoption plan URL for senior/junior teams and the values were listed as expected

PR #3358, opened 2022-04-20
Update styling for judging buttons

shaunxp20:
Just some basic styling updates to make the buttons look a bit better.

The main thing I changed was wrapping the buttons in a flex container (with `justify-content: space-between;`).


viviancan (approved):
👍 Looks good! Flexbox is so awesome!

PR #3362, opened 2022-04-21
Allow additional CSS in application rebrand layout

shaunxp20:
The fix/hack I made yesterday to get the judging styles working on QA was to just include the judging CSS on every page that uses the application rebrand layout.

This change is a more proper fix and will only include the judging CSS for judging pages (basically making it like it was before).

### DIFF
--- a/app/views/layouts/application_rebrand.html.erb
+++ b/app/views/layouts/application_rebrand.html.erb
@@ -21,7 +21,7 @@
 
     <%#= stylesheet_packs_with_chunks_tag "application" %>
 
-    <%#= yield :css %>
+    <%= yield :additional_css %>
 
     <%= javascript_include_tag determine_manifest,
       'data-turbolinks-track' => true %>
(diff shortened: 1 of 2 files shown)


app/views/layouts/application_rebrand.html.erb
```diff
@@ -21,15 +21,14 @@
 
     <%#= stylesheet_packs_with_chunks_tag "application" %>
 
-    <%#= yield :css %>
+    <%= yield :additional_css %>
```
shaunxp20, line 24: This was how the judging CSS used to be included (via `yield :css`), I enabled this and renamed it to `:additional_css` because I'm not sure what else could be potentially included by enabling this, and this way we know only the judging CSS will use `:additional_css`, seems safer. 😅

viviancan (approved):
I like this change! Definitely feels more intentional with the additional CSS :) 👍 

Thanks for figuring this one out!

PR #3363, opened 2022-04-21
3350 overview section update and start score btn

viviancan:
Updating Overview.vue component + all associated wiring up to ProjectDetails for better understanding

### DIFF
--- a/app/javascript/judge/scores/sections/ProjectDetails.vue
+++ b/app/javascript/judge/scores/sections/ProjectDetails.vue
@@ -2,7 +2,7 @@
     <div>
       <GenericJudgingContainer
           heading="Project Name & Description"
-          section="overview"
+          section="project_details"
           next-section="ideation"
           next-button-text="Learning Journey"
       >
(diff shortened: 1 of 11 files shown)


app/javascript/judge/scores/sections/ProjectDetails.vue
```diff
@@ -2,7 +2,7 @@
     <div>
       <GenericJudgingContainer
           heading="Project Name & Description"
-          section="overview"
+          section="project_details"
```
viviancan, line 5: The only way I could get the question comment section to render properly was by using snake case as section name!

viviancan:
Updating the naming due to issue where the "title" was undefined on the question component due to mismatched naming. This is now working as expected, however, I am unsure if using snake case in this way is best practice.

shaunxp20:
Maybe this is okay, but I am wondering if the new project details comment gets saved to the database. 🤔 

Also, we should rename the database fields, either in this PR or as a task in #3350, other than that, LGTM!

shaunxp20 (approved):
LGTM! Nice work! Makes it a lot clearer now! There might be some follow on work, but we can address that in another PR too. 👍 🙌 🎉

PR #3373, opened 2022-04-26
3350- Review-Score page rebrand

viviancan:
Files changed
- `app/javascript/judge/scores/sections/ReviewScore.vue`
- `app/javascript/tailwind.config.js`

Refs
#3350 

Notes
- Added additional variants to the tailwind config in order to access the child utilities provided by tailwind. Docs [here](app/javascript/tailwind.config.js). 
- Rebrand is complete, however, there is a weird bug with the way the state/getters is configured, which is causing inconsistent displays of the exclamation visual which indicates there is an error with the score. There are more details + a video of the bug added in #3350

### DIFF
--- a/app/javascript/judge/scores/sections/ReviewScore.vue
+++ b/app/javascript/judge/scores/sections/ReviewScore.vue
@@ -1,76 +1,53 @@
 <template>
-  <div class="grid">
-    <div class="grid__col-12 grid__col--bleed-y">
-      <h2>Review score</h2>
-    </div>
-
-    <team-info />
-
-    <div class="grid__col-9 grid__col--bleed-y">
-      <h1 class="border--b-thin-primary">{{ submission.name }}</h1>
-
-      <div class="app-description" v-html="submission.description"></div>
-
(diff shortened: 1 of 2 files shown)


app/javascript/judge/scores/sections/ReviewScore.vue
```diff
@@ -106,18 +86,30 @@ export default {
       'problemInSection'
     ]),
 
-    sectionNames () {
-      let names = [
-        'overview',
-        'ideation',
-        'pitch',
-        'demo',
+    sectionDetails () {
```
viviancan, line 89: I came up with this approach as a potential solution for the section names vs. display names on this page. It seems that this issue has popped up in many areas (here, the comment titles, button text to name a few), so maybe we need to rethink a more reusable approach :)

viviancan: Disregard ^^^ This already exists in `app/javascript/judge/scores/store/getters.js`... 😆 Thanks @shaunxp20 !

Issue #3234, opened 2022-01-19
New submission fields to be edited by admin


rsgonzal:
Add new submission fields to the admin

- learning journey
- Submission type
- Additional Questions
- images (this was never included here for some reason not sure if it's too complicated and it was on purpose or oversight)

![screencapture-technovation-qa-herokuapp-admin-team-submissions-no-name-yet-team-technovation-edit-2022-01-24-09_51_27.png](https://images.zenhubusercontent.com/60527169cde2772242d0de95/e1d5bd83-2a3f-46d9-a434-ef23c2b76a63)

dboyer:
Will not add image editing/upload in admin

devkonrad:
Just posted a PR for this one: https://github.com/Iridescent-CM/technovation-app/pull/3369

dboyer:
- [ ]  While editing a junior division app, it says "user adoption plan" when viewing the submission details but "business plan" when editing the details. It would be helpful if both said "user adoption plan" for junior division and "business plan" for senior division.

<img width="308" alt="Screen Shot 2022-04-25 at 8 10 31 PM" src="https://user-images.githubusercontent.com/516764/165194147-b1ba2525-f8c0-49f9-b0d5-8cff8009d771.png">

<img width="404" alt="Screen Shot 2022-04-25 at 8 10 06 PM" src="https://user-images.githubusercontent.com/516764/165194119-ec816a0c-157f-4206-88e0-e3821ec9de78.png">

devkonrad:
> * [ ]   While editing a junior division app, it says "user adoption plan" when viewing the submission details but "business plan" when editing the details. It would be helpful if both said "user adoption plan" for junior division and "business plan" for senior division.
> 
> <img alt="Screen Shot 2022-04-25 at 8 10 31 PM" width="308" src="https://user-images.githubusercontent.com/516764/165194147-b1ba2525-f8c0-49f9-b0d5-8cff8009d771.png">
> 
> <img alt="Screen Shot 2022-04-25 at 8 10 06 PM" width="404" src="https://user-images.githubusercontent.com/516764/165194119-ec816a0c-157f-4206-88e0-e3821ec9de78.png">

Sure! I will take care of it! :)

devkonrad:
> * [ ]   While editing a junior division app, it says "user adoption plan" when viewing the submission details but "business plan" when editing the details. It would be helpful if both said "user adoption plan" for junior division and "business plan" for senior division.
> 
[...]

PR #3377, opened 2022-04-26
Adjust plan label for junior for admin on edit submission

devkonrad:
Adjust plan label for junior for admin on edit submission

Changes:
modified: app/views/admin/team_submissions/edit.html.erb

Refs: https://github.com/Iridescent-CM/technovation-app/issues/3234

### DIFF
--- a/app/models/team_submission.rb
+++ b/app/models/team_submission.rb
@@ -543,14 +543,6 @@ class TeamSubmission < ActiveRecord::Base
     team_division_name == "senior"
   end
 
-  def plan_type
-    if self.senior_division?
-      :business_plan
-    elsif self.junior_division?
-      :user_adoption_plan
-    end
-  end
-
   def development_platform_text
     if submission_type == MOBILE_APP_SUBMISSION_TYPE && development_platform != "Other"
       "#{submission_type} - #{development_platform}"
--- a/app/models/team_submission.rb
+++ b/app/models/team_submission.rb
@@ -543,6 +543,14 @@ class TeamSubmission < ActiveRecord::Base
     team_division_name == "senior"
   end
 
+  def plan_type
+    if self.senior_division?
+      :business_plan
+    elsif self.junior_division?
+      :user_adoption_plan
+    end
+  end
+
   def development_platform_text
     if submission_type == MOBILE_APP_SUBMISSION_TYPE && development_platform != "Other"
       "#{submission_type} - #{development_platform}"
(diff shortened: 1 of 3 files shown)


app/views/admin/team_submissions/edit.html.erb
```diff
@@ -25,18 +25,21 @@
...
         
-        <div class="grid__cell--padding-md-y">
-          <h6>Entrepreneurship</h6>
-          <%= f.label :business_plan %>
-          <% if @team_submission.business_plan_url %>
-            <%= link_to @team_submission.business_plan_identifier,
-              @team_submission.business_plan_url,
-              target: :_blank %>
-          <% end %>
+        <% if !@team_submission.plan_type.nil? %>
```
shaunxp20, line 28: This looks like if a senior or junior team doesn't have a business plan (`nil?`), this will prevent an admin from adding one b/c this section wouldn't be displayed - it might be better to check if it's a senior or junior team instead. 😄

shaunxp20: Oh wait, I see, I think this is okay b/c `plan_type` doesn't actually return the plan, it returns a symbol based on whether it's a sr/jr team. 🤯 For this reason, I think it would less confusing to explicitly check for division here! Just IMHO! 😅 😄

app/models/team_submission.rb
```diff
@@ -520,6 +520,14 @@ def senior_division?
     team_division_name == "senior"
   end
 
+  def plan_type
```
shaunxp20, line 523: Hmm, I'm thinking this name is a little too vague, I'm not sure `@team_submission.plan_type` is the best way to convey that it's the entrepreneurship (business/adoption) plan. Also, I see it just returns a symbol for a form label, it doesn't actually return the business/adoption plan - this might be better as a helper to help distinguish that it's just for the view layer, or just inline it w/  ternary expression. 😄

devkonrad: Thanks @shaunxp20 ! I will place the plan name in a helper. :)

viviancan:
@shaunxp20 @dboyer I think this PR can be closed. I checked the admin portal and the other fields are there to be edited and the upload label displays as expected for senior vs. junior submissions

dboyer:
I think you're right @viviancan. I'm seeing user adoption plan for juniors, business plan for seniors, and neither for beginners. so I don't think this is needed.

viviancan:
Closing this as this is no longer an issue in the admin portal

Issue #3247, opened 2022-01-28
Add keyword search function to admin search pages


dboyer:
Mentors indicate their school or company name using an open text field. This makes searching for all the mentors or judges from a certain organization challenging as they may have entered slightly different names (i.e. Google, Google Brazil, Google (retired), etc). 

It would be useful if admins could keyword search for mentors or judges and have all school or company names that include that keyword return in results. Currently, admins must select a company from a drop-down list when searching for mentors or judges.

viviancan:
@dboyer Verifying that this search should only return results for mentor and judge accounts?

dboyer:
Correct. Program team confirmed that this search should only apply to mentors and judges, not students.

PR #3381, opened 2022-04-28
Performs admin search for company/school name through substring

devkonrad:
Allow perform search for company name through substring

Changes:
modified: app/data_grids/accounts_grid.rb

Refs: https://github.com/Iridescent-CM/technovation-app/issues/3247


dboyer:
Confirmed with other teams that having only a keyword search is fine. We do not need to include the auto-complete drop-down search in a separate search field.

We also should not expand the school search to include students. It should remain as just mentors and judges.

devkonrad:
> Confirmed with other teams that having only a keyword search is fine. We do not need to include the auto-complete drop-down search in a separate search field.
> 
> We also should not expand the school search to include students. It should remain as just mentors and judges.

Great! Thank you Deb!

viviancan (approved):
LGTM! Works as expected 🎉

devkonrad:
> LGTM! Works as expected tada

Thanks for the review, @viviancan ! :)

PR #3384, opened 2022-04-29
Adjust low score thresholds

shaunxp20:
The low score thresholds are being updated in the PR.

The tests had to be updated, and that's like 90% of the changes for this PR. 😅


viviancan (approved):
LGTM! Thanks for tackling this, Shaun 💯 🥇

PR #3389, opened 2022-04-30
Add judge progress icons

shaunxp20:
I added a new component for the progress icons - `SectionProgressIcons`. These progress icons will get displayed on the score summary page (like they do now), and now they will get displayed in the score stepper in the sidebar.


viviancan (approved):
This is awesome 👍 Tried it locally and was able to see the updated score stepper! I think it would be great to get some additional eyes on it for the different divisions and edge cases. Thanks, Shaun 🎉

PR #3390, opened 2022-04-30
Auto-redirect to judge dashboard

shaunxp20:
After someone signs in, if certain criteria is met, they will get redirected to their judge dashboard.

This PR introduces a new service `JudgeDashboardRedirector` that contains the conditions for when someone will get automatically redirected to their judge dashboard. This new service will get called (to see if the conditions are met):
- when anyone logs in
- when an admin uses the impersonation functionality.

### DIFF
--- a/app/controllers/admin/participant_sessions_controller.rb
+++ b/app/controllers/admin/participant_sessions_controller.rb
@@ -5,9 +5,13 @@ module Admin
       participant.regenerate_session_token
       set_cookie(CookieNames::SESSION_TOKEN, participant.session_token)
 
-      redirect_to send(
-        "#{participant.scope_name.sub(/^\w+_r/, "r")}_dashboard_path"
-      )
+      if JudgeDashboardRedirector.new(account: participant).enabled?
+        redirect_to judge_dashboard_path
+      else
+        redirect_to send(
+          "#{participant.scope_name}_dashboard_path"
+        )
+      end
     end
 
     def destroy
--- a/app/technovation/sign_in.rb
+++ b/app/technovation/sign_in.rb
@@ -42,18 +42,19 @@ module SignIn
   def self.after_signin_path(signin, context)
     last_profile_used = context.remove_cookie(CookieNames::LAST_PROFILE_USED)
 
-    if last_profile_used == "regional_ambassador"
-      last_profile_used = "chapter_ambassador"
-    end
-
     if last_profile_used and
         not signin.public_send("#{last_profile_used}_profile").present?
       last_profile_used = nil
     end
 
-    (last_profile_used && "#{last_profile_used}_dashboard_path") or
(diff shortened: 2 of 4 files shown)


app/technovation/sign_in.rb
```diff
@@ -42,18 +42,19 @@ def self.call(signin, context, options = {})
   def self.after_signin_path(signin, context)
     last_profile_used = context.remove_cookie(CookieNames::LAST_PROFILE_USED)
 
-    if last_profile_used == "regional_ambassador"
-      last_profile_used = "chapter_ambassador"
-    end
```
shaunxp20, line 47: This was put in place back when we renamed regional ambassadors to chapter ambassadors, which was a couple years ago, so this isn't needed now. Side note, there's some other cleanup to do around this too. 😄

app/technovation/sign_in.rb
```diff
@@ -42,18 +42,19 @@ def self.call(signin, context, options = {})
...
-
     if last_profile_used and
         not signin.public_send("#{last_profile_used}_profile").present?
       last_profile_used = nil
     end
 
-    (last_profile_used && "#{last_profile_used}_dashboard_path") or
-      "#{signin.scope_name.sub(/^\w+_chapter_ambassador/, "chapter_ambassador")}_dashboard_path"
```
shaunxp20, line 55: Apparently we have a `pending_chapter_ambassador`, and so this snippet would redirect a `pending_chapter_ambassador` to the `chapter_ambassador_dashboard_path`. I made this more explicit below. 😅

app/controllers/admin/participant_sessions_controller.rb
```diff
@@ -5,9 +5,15 @@ def show
       participant.regenerate_session_token
       set_cookie(CookieNames::SESSION_TOKEN, participant.session_token)
 
-      redirect_to send(
-        "#{participant.scope_name.sub(/^\w+_r/, "r")}_dashboard_path"
```
shaunxp20, line 9: This is related to my comment below about pending chapter ambassadors, only this one is for `pending_regional_ambassadors` - the `r` in the regex if for the `r` in regional ambassador, if that wasn't obvious! 😅 Anywho, this obtuse regex would replace the `pending_r` (`\w+_r`) and replace it with `r`, to make this `regional_ambassador_dashboard_path`. Anywho, I made this more explicit below. 😃

shaunxp20: Just following up with this - apparently we have a route this will redirect a pending chapter ambassador to the chapter ambassador dashboard.

```
get "pending_chapter_ambassador/dashboard",
  to: "chapter_ambassador/dashboards#show"
```

So, we don't need to capture this "pending" scenario here. 😅 💦

viviancan (approved):
Nice changes @shaunxp20! Great coverage and cleanup. I tried this locally and it seemed to work as expected! 👍

PR #3394, opened 2022-05-03
Updated score entry to account for different score value tooltips bas…

viviancan:
…ed on division per the rubric


shaunxp20 (approved):
Awesome! Thanks for taking care of this @viviancan! 🙌 👍 🎉 😀

PR #3398, opened 2022-05-03
Update "Scores in Progress" header

shaunxp20:
Just making the "Scores in Progress" header dynamic here, since most people will just have one score in progress. 😃


viviancan (approved):
👍 Nice!

Issue #3400, opened 2022-05-04
Lingering message regarding data exports from admin pages


dboyer:
Several admin users are reporting a lingering message on the admin pages indicating that an export is in progress. Kicking off a new export will result in that CSV downloading but then the same message returns.

<img width="724" alt="Screen Shot 2022-05-04 at 11 03 03 AM" src="https://user-images.githubusercontent.com/516764/166710857-f575ef52-e22e-42ee-bf6f-3fdca380afad.png">

viviancan:
Update 5/4/22: 

[PR Here](https://github.com/Iridescent-CM/technovation-app/pull/3402)

Created this rake task to address the issue where queued export jobs in the admin profile are causing the "export" popup in the admin profile to be stuck in a sort of infinite loop. There is currently no way to "force end" these jobs as they do not show up in Sidekiq. We believe this issue is happening due to a previous deployment that introduced an export bug on production. While that issue has been resolved, there are a few pending jobs that somehow got out of sync. This rake task will address these issues.

viviancan:
Verifying the expected output on QA. Ready for prod

PR #3402, opened 2022-05-04
Added rake task to force update the status of queued jobs to failed by owner type

viviancan:
Created this rake task to address the issue where queued export jobs in the admin profile are causing the "export" popup in admin profile to be stuck in a sort of infinite loop.

Refs #3400

### DIFF
--- a/lib/tasks/jobs.rake
+++ b/lib/tasks/jobs.rake
@@ -6,7 +6,6 @@ namespace :jobs do
     owner_type = args[:owner_type]
     dry_run = args[:dry_run] != "run"
 
-
     puts "DRY RUN: #{dry_run ? 'on' : 'off'}"
     puts "Force failing queued jobs for #{owner_type} that were created before #{(Date.today - num_of_days.days).strftime("%B %d, %Y")}"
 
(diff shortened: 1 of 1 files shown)


lib/tasks/jobs.rake
```diff
@@ -0,0 +1,10 @@
+namespace :jobs do
+  desc "Force fail queued jobs between today and 1 month ago by owner type"
+  task :force_fail_queued_jobs, [:owner_type] => :environment do |t, args|
+    puts "Force failing queued jobs for #{args[:owner_type]}"
+    Job.where(created_at: (Date.today - 1.month)..Date.today).where(status: "queued").where(owner_type: args[:owner_type]).find_each do |job|
```
shaunxp20, line 5: Some random thoughts - instead of removing jobs from today, what about removing the pending ones that are 7 (or how many ever) days old/pending? Or passing in a list of specific job ids to mark as failed? Or creating dry run functionality to ensure the right jobs are targeted? 🤔 😃

viviancan: Nice thoughts, what about 5 days + dry run?

shaunxp20: It could even be an argument to pass in, but either way, sounds good to me! I like it! 👍 🎆

devkonrad: I liked the idea of ​​mark as fail, because it will allow us to have a history of the cases.

devkonrad (approved):
Looks good to me. I liked it! :)

shaunxp20 (approved):
I didn't try it, but looks good to me! Dry run functionality, more versatile, love it! 👏 🥳 🙌

PR #3405, opened 2022-05-05
Registration settings backend endpoint

shaunxp20:
This will create a new backend endpoint that will return registration settings, these settings will be used to turn on/off student and mentor registrations.

Judges aren't included yet since we don't have an admin setting to control judge registrations yet - this will be done later.


devkonrad (approved):
Perfect. Clean and completely readble! I liked it.

viviancan (approved):
LGTM! I didn't pull anything locally but looking forward to seeing all these pieces come together. I also really like the testing coverage 💯

PR #3406, opened 2022-05-06
Disable student and parent registrations

shaunxp20:


viviancan (approved):
Does the red helper text need to be updated to include students? 
` Registration for the 2022 season is now closed for mentors` 
Otherwise, looks good to me! Thx Shaun

shaunxp20:
> Does the red helper text need to be updated to include students? ` Registration for the 2022 season is now closed for mentors` Otherwise, looks good to me! Thx Shaun

Great catch, thanks for pointing this out! 🏆 🍾 💯 Just updated this! 🙌

Issue #3372, opened 2022-04-26
Add limits on development platform when editing submissions as an admin


dboyer:
Split off from #3234 as this is not as time sensitive. All tasks in this issue pertain to the admin page that supports editing submissions. Go to /admin/team_submissions#/ and then select a submission to view the page where admins can edit submitted fields.

- [ ] I selected Mobile App as my submission type and app inventor as my development platform. I saved and it forced me to enter something in the app inventor app name field, which is correct behavior. However, I then entered a URL in the Thunkable project URL field and was able to save. The view after saving did not display the Thunkable URL, which I'm guessing is because "app inventor" was selected as my development platform. It would be helpful if we could have the page return a warning:

- If app inventor selected as the development platform and an admin tries to add a Thunkable project URL, show an error on the Thunkable project URL field saying "Cannot add a Thunkable project URL when App Inventor selected as the development platform." Do not allow user to save the form with those edits
- If Thunkable selected as the development platform and an admin tries to add an App Inventor app name, show an error on the App Inventor name field saying "Cannot add an App Inventor app name when Thunkable selected as the development platform." Do not allow user to save the form with those edits.

<img width="426" alt="Screen Shot 2022-04-25 at 8 18 59 PM" src="https://user-images.githubusercontent.com/516764/165194816-c264aa6a-cf39-447c-9110-da24746fa6bc.png"> 

<img width="438" alt="Screen Shot 2022-04-25 at 8 19 49 PM" src="https://user-images.githubusercontent.com/516764/165194892-29ac30bb-96c1-4fa8-bca6-6a1f3c7797af.png">
 

- [ ] When editing the pitch video and demo video fields, it is possible to put text in the field rather than a URL. Require a link to a URL. 
<img width="440" alt="Screen Shot 2022-05-03 at 10 18 50 AM" src="https://user-images.githubusercontent.com/516764/166470981-f48e0b46-5d00-4361-ad22-0e17b9df590d.png">

devkonrad:
Created the PR https://github.com/Iridescent-CM/technovation-app/pull/3412 to deal with this demand.

viviancan:
` When editing the pitch video and demo video fields, it is possible to put text in the field rather than a URL. Require a link to a URL.`

@dboyer Following up one of the checkboxes - is the word "link" supposed to show instead of the video URL?

viviancan:
A few followup notes - 

- What is considered a "valid" video URL? 

[...]

PR #3412, opened 2022-05-09
Add data limit according to platform

devkonrad:
Ads data constraint according to platform, for URLs and other fields 

Changes:

Refs: https://github.com/Iridescent-CM/technovation-app/issues/3372

### DIFF
--- a/app/models/team_submission.rb
+++ b/app/models/team_submission.rb
@@ -47,18 +47,18 @@ class TeamSubmission < ActiveRecord::Base
     use: :scoped,
     scope: :deleted_at
 
-  before_validation -> {
-    return if thunkable_project_url.blank?
-    
-    self.thunkable_project_url = padronize_url(thunkable_project_url)
-    self.demo_video_link = padronize_url(demo_video_link)
-    self.pitch_video_link = padronize_url(pitch_video_link)
-  }
-
   before_validation :reset_development_platform_fields_for_ai_projects
   before_validation :reset_development_platform_fields_for_app_inventor
   before_validation :reset_development_platform_fields_for_thunkable
   before_validation :reset_development_platform_fields_for_other_platforms
+
+  before_validation -> {
+    return if thunkable_project_url.blank?
+    
+    self.thunkable_project_url = padronize_url(self.thunkable_project_url)
+    self.demo_video_link = padronize_url(self.demo_video_link)
+    self.pitch_video_link = padronize_url(self.pitch_video_link)
+  }
 
   before_commit -> {
     self.ai_description = "" if ai.blank?
--- a/app/models/team_submission.rb
+++ b/app/models/team_submission.rb
@@ -612,7 +612,7 @@ class TeamSubmission < ActiveRecord::Base
 
   private
 
-  def standardize_url( url )
+  def standardize_url(url)
     return if url.blank?
 
     if !url.match(/^https:\/\//)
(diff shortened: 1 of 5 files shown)


devkonrad:
As discussed with @dboyer the "App Inventor Name" need to raise an error when its filled and we select "thunkable" as platform.

devkonrad:
> Ads data constraint according to platform, for URLs and other fields
> 
> Changes: modified: app/models/team_submission.rb
> 
> Refs: #3372

**QA Test Steps:**
_Enter to edit team submission page (on admin)_
1 - When develop platform changed to thunkable:
	- If you try to fill de App Inventor Name you will get an error
	- If you try to save without Thunkable URL you will get an error
2 - When develop platform changed to thunkable:
	- If you try to save with a Thunkable URL you will get an error
3 - If you try to save invalid URLs value on Demo video Link you will get an error
4 - If you try to save invalid URLs value on Pitch video Link you will get an error

app/models/team_submission.rb
```diff
@@ -593,6 +612,20 @@ def question_response(question)
 
   private
 
+  def padronize_url( url )
```
viviancan, line 615: Hey Alex! Thanks for all the work on this one! 

It looks like this method name is in Portuguese (padronize). Could you update it to the English equivalent maybe something along the lines of `standardize_url` or something similar, in order to keep the consistency with English naming throughout the application?

viviancan (changes requested):
Hi Alex, I left a few followup notes on the issue #3372. I had a few issues with editing a submission.

devkonrad:
> Hi Alex, I left a few followup notes on the issue #3372. I had a few issues with editing a submission.

Hi, Vivian! I answer your question about the validation here https://github.com/Iridescent-CM/technovation-app/issues/3372 . I just remain the validation that was use, but we can impove that if needed. :)

app/validators/thunkable_share_url_validator.rb
```diff
@@ -1,6 +1,6 @@
 class ThunkableShareUrlValidator < ActiveModel::EachValidator
   def validate_each(record, attribute, value)
-    if !value.match(/^https?:\/\/x.thunkable.com\/projects\/.+$/)
```
viviancan, line 3: The "project" version is actually correct here! Sry about the confusion Alex!

viviancan (changes requested):
Hi Alex, thanks for tackling this 👍 After verifying w/ deb this morning, I realized I misunderstood the requirement for the the thinkable url this morning. The original version with "project" is actually the correct version of the link we are accepting now - not the version with "copy". The change will have to be reverted back to the original URL that includes "project" - I think  in file `app/validators/thunkable_share_url_validator.rb` and the spec file.

viviancan (changes requested):
Hey Alex, LGTM! Tested it locally and it is working as expected.

The only thing is on `def standardize_url( url )` there is extra whitespace around `url`.

devkonrad:
> Hey Alex, LGTM! Tested it locally and it is working as expected.
> 
> The only thing is on `def standardize_url( url )` there is extra whitespace around `url`.

Great, @viviancan! Thanks for the review. I will fix it and merge to QA. :)

dboyer:
@devkonrad Feel free to merge this to QA if it's just reverting the previous change.

devkonrad:
> @devkonrad Feel free to merge this to QA if it's just reverting the previous change.

Great. Doing that. :)

Issue #3413, opened 2022-05-09
Update home page text


dboyer:
The home page text does not accurately reflect registration. 
<img width="749" alt="Screen Shot 2022-05-09 at 12 05 36 PM" src="https://user-images.githubusercontent.com/516764/167451043-f66533a0-b3bd-4f43-bc38-32e71a56f830.png">

Let's change it to the following text. This text should appear when judge registration is on and student/mentor registration is off.

Technovation Girls
Registration is currently open for judges.

Thank you for your interest in Technovation Girls. Registration is currently closed for students and mentors, although students with existing accounts are welcome to log-in [link to sign in page] and review their existing projects.

You can learn more about Technovation on our website [link to https://www.technovation.org/] or follow us at: (keep list of social media icons)

dboyer:
Note that this will need to be hard coded as the judge toggle for registration doesn't exist yet. Check with @viviancan before starting as she looked into it.

viviancan:
@dboyer Should there be text referring to judges signing up? It seems that the same issue where judges won't know how to get to sign up could happen again. When reviewing this locally I did not realize that the text "Registration is currently open for judges" was actually a link to the registration page, it looks like a heading.

![Screen Shot 2022-05-12 at 4.30.55 PM.png](https://images.zenhubusercontent.com/606b0f4fd13d53dff3d14fa5/4d6d815a-5067-4de6-bf79-7a6af34aa77e)

devkonrad:
> Oh good catch. Yep, the color on that does make it look just like a heading instead of a link. Let's:
> 
> * Remove the link from "Registration is currently open for judges"
> * Change the first sentence to read: Thank you for your interest in Technovation Girls. Register here to judge projects."
> * Link "register here" to https://my.technovationchallenge.org/signup#/

Great, @dboyer and @viviancan ! I'll do that!

dboyer:
Reopening this so we remember it still needs to be deployed

dboyer:
No worries! Confirmed that the text looks good on staging.

PR #3416, opened 2022-05-11
Updates home page splash for 'open for judges' version

devkonrad:
Updates home page splash for 'open for judges' version

Changes:
modified:  app/views/application/_open_judge_splash.en.html.erb
modified:  app/views/public/dashboards/show.en.html.erb
modified:  spec/features/admin/off_season_splash_spec.rb

Refs: https://github.com/Iridescent-CM/technovation-app/issues/3413


devkonrad:
As siscussed with @dboyer (just to register) we need to do this changes:
- [ ] Remove the sign-up link (the words after login)
- [ ] Link "Registration is currently open for judges" to https://my.technovationchallenge.org/signup#/
- [ ] Link the word log-in to https://my.technovationchallenge.org/login

app/views/public/dashboards/show.en.html.erb
```diff
@@ -9,7 +9,7 @@
 <% else %>
   <div class="grid grid--justify-space-around">
     <div class="panel grid__col-12 grid__col-md-6">
-      <%= render 'off_season_splash' %>
+      <%= render 'open_judge_splash' %>
```
shaunxp20, line 12: The name of this new partial is popping out to me, I wonder if it can be named better, some suggestions: `judging_season_splash` or `judging_open_splash`. 🤔 😄

viviancan: +1 here - I like the `judging_open_splash`

devkonrad: Thank you @viviancan and @shaunxp20 for the review i'll proceed with the rename suggestion! :)

devkonrad:
> I left a clarifying question/comment on the ticket for Deb. When reviewing this locally, it is unclear that the text "Registration is currently open for judges" is actually a link to the registration page. This may be confusing, especially given recent feedback about this page.
> 
> I will wait to approve until a response from Deb. Otherwise, this looks great to me. Thanks for handling this one, Alex +1

Thanks, @viviancan. Meanwhile i'll do the other suggestion changes.

app/views/application/_judging_open_splash.en.html.erb
```diff
@@ -0,0 +1,44 @@
+<h1 class="appy-title">Technovation Girls</h1>
+
+<h2>Registration is currently open for judges</h2>
+
+<p>
+  Thank you for your interest in Technovation Girls. Register <%= link_to "here", "https://my.technovationchallenge.org/signup#/" %> to judge projects.
```
viviancan, line 6: I would recommend using the `signup_path, data: { turbolinks: false }` so the signup link is not hard coded here.

app/views/application/_judging_open_splash.en.html.erb
```diff
@@ -0,0 +1,44 @@
...
+</p>
+
+<p>
+  <strong>
+    Registration is currently closed for students and mentors, 
+    although students with existing accounts are welcome to 
+    <%= link_to "log-in", "https://my.technovationchallenge.org/login" %> and review their existing projects.
```
viviancan, line 13: Same here, I would recommend using the login path ` login_path, data: { turbolinks: false }`

viviancan (changes requested):
Hey Alex, thanks for tackling this one! I would update the links so they are not hardcoded! Besides that, the changes look good 👍

app/views/application/_judging_open_splash.en.html.erb
devkonrad: Great tip, Vivian! Thank you!

viviancan (approved):
Looks good, thx Alex 💯

Issue #2960, opened 2021-05-24
Provide an option to expand images when viewing as a judge


rsgonzal:
As a judge, it is sometimes difficult to see images that teams have uploaded while scoring.

Goal: Give an option to expand the size of images, either on the screen or to open them bigger in another tab.

More context: Vertical images seem to show up bigger than horizontal images
![Screen Shot 2021-05-24 at 2.41.27 PM.png](https://images.zenhubusercontent.com/60527169cde2772242d0de95/fd2020ca-f00a-4ca3-be2c-fabda982dc41)

viviancan:
### Developer Notes - Adding Submission Pictures Locally
1. Login under admin account - admin@admin.com (username & pw)
2. Click content & settings tab (right hand nav bar)
3. Turn off judging --> judging > off
4. Allow submissions to be edited --> teams& submissions > team submissions are editable
5. Click review bottom corner and then save these settings button
6. Log out as admin
7. Log in as student account --> student@student.com (username & pw)
8. Edit submission. Add additional photos
9. Log out of student and log back in as admin
10. Turn judging back on. save settings
11. View judge account that has judged the default student submission --> Ozark Mountaineers - The Best App
12. You should see the uploaded images from earlier

Check out this [walkthrough video](https://drive.google.com/file/d/16srP1gPd2ec53lAODYBQZQT5ddw0qDQ6/view?usp=sharing) for more details

tamtran-agilityio:
@viviancan Thank you so much!

tamtran-agilityio:
@rsgonzal @viviancan I will set the hard height size for images on carousel and open view new tab with an image default when click on image. Is is correct? You can see code on pull request https://github.com/Iridescent-CM/technovation-app/pull/2964. If have any suggestions please let me know, thanks!

shaunxp20:
Moving this one to the Ice Box since we don't have time to look at it right now.

dboyer:
Pulling this into the sprint backlog as a potential task to revisit. The related PR is likely not usable at this point, but there may be some other helpful context here.

Note: it may be most useful to just open the image in a new tab.

dboyer:
Looking at QA:

- The "Done" button on the image pop-up is now fairly low on the screen. Screenshot is from my 13 inch Macbook Air. Is there a way to bring it up a bit? I don't think we need to optimize for screens smaller than that but 13 inch is common enough that it would be nice for the button to be visible.

[...]

PR #3423, opened 2022-05-12
Updated judge sweet alert modal to expand thumbnail size based on judge feedback

viviancan:
This change was made in order to increase the size of the image/modal when a judge clicks on a screenshot in the learning journey. I added an additional event listener so these changes would only affect these screenshots. The original event listener/sweet alert code affects the student image submissions. 

I explored a few options but ultimately settled on this approach. We may want to discuss this further as part of the 6-9 month clean up tasks. Some things to consider:
1. Sweetalert2 has its own styles/constraints
2. I was unable to add tailwind styles so I had to resort to the original `judge.scss` file
3. Would it make sense in the future to add a custom modal component, especially since this lives in the vue world and we also have access to tailwind components

Refs: #2960

### DIFF
--- a/app/assets/stylesheets/judge.scss
+++ b/app/assets/stylesheets/judge.scss
@@ -53,3 +53,25 @@
     opacity: 1;
   }
 }
+
+.judge-sub-screenshot {
+  width: 100% !important;
+
+  .screenshot-nav {
+    display: flex;
+    justify-content: space-between;
+    height: 100%;
+    align-items: center;
+  }
+
+  .screenshot__img {
+    object-fit: contain;
+    height: 40rem;
+  }
+}
+
+@media only screen and (min-width: 992px){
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/judge.scss
```diff
@@ -53,3 +53,25 @@
     opacity: 1;
   }
 }
+
+.judge-sub-screenshot {
+  width: 100% !important;
```
viviancan, line 58: I don't really like the use of `!important`, however, in this case, this was the only way I could override the sweetalert default styles while also considering mobile responsiveness

devkonrad (approved):
All the code looks good to me :)

shaunxp20 (approved):
Didn't try it locally, but looks good to me! Thanks for working on this one! And for the details in the write-up! 👍 🎉 🥳

PR #3428, opened 2022-05-13
Wire up registration settings

shaunxp20:
With this PR, student and mentor registrations will be controlled via the admin controls. Judge registrations will be turned on via hard-coding (if that makes sense).

After this PR will come some clean up w/ the hard-coded dates and making some verbiage dynamic.

### DIFF
--- a/app/javascript/new_registration/components/StepOne.vue
+++ b/app/javascript/new_registration/components/StepOne.vue
@@ -1,5 +1,5 @@
 <template>
-  <div id="step-one">
+  <div v-if="isRegistrationOpen" id="step-one">
     <ContainerHeader header-text="Your Profile Type" />
 
     <div id="profile-type" class="form-wrapper">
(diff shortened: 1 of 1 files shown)


app/javascript/new_registration/components/StepOne.vue
```diff
@@ -49,13 +57,27 @@ export default {
...
       isMentorRegistrationOpen: false,
       hasValidationErrors: true
     };
   },
   methods: {
+    async getRegistrationSettings() {
+      try {
+        const response = await axios.get('/api/registration_settings')
+
+        this.isRegistrationOpen = true
```
shaunxp20, line 71: This will eventually be set with what is returned from the back-end, but right now it has to be hard-coded in order for judges to be able to register.

devkonrad: > This will eventually be set with what is returned from the back-end, but right now it has to be hard-coded in order for judges to be able to register.

I liked the solutiuon, its just waiting for the implementation of backend return and will be automatic. Great! :)

devkonrad (approved):
Looks great!

Issue #3427, opened 2022-05-13
Add automatic email sent to judges after registration


dboyer:
After registration, judges should automatically receive an email with information on training and judging.

Text is [available here](https://docs.google.com/document/d/1SHFC6IpjfBnM4HC1wHN6AGGW1tKR4TncKPlDy-7J9tc/edit#heading=h.qx69ywuh0brr). Please note the comment from Mackenzie about editing the mailing header.

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/3427?planning-poker) with ZenHub @shaunxp20 @viviancan

viviancan:
Verified this locally and on QA. After registering as a judge on QA, I received the welcome email. I also clicked the links in the email and they seem appropriate. Would be helpful to get a few more eyes to verify!

viviancan:
QA Steps
1. Register as a new judge
2. Confirm auto email is sent to your email
3. Verify the links are correct
4. Verify HTML body text is correct

viviancan:
Registered as a judge on QA, I received the email and verified the links.

dboyer:
A few edits to the email:

1. Change "So, let's get started! This is what you can do in under 40 minutes in order to start matching with teams" TO "Let's get started! This is what you can do in under 40 minutes to start getting submissions to review:"

2. Remove  4. in front of the last bullet and make it its own sentence. So it reads: "You are now all set to start Scoring! (link) Log on to your dashboard and click “start score” to begin." 

3. Conclude the email with "Thank you for your time and support! 

Regards, 
The Technovation Team"

viviancan:
Welcome judge email text updates are on code review!

dboyer:
Confirmed that I see the updates on staging

dboyer:
One last round of changes:

- Take out the ~ in front of 4000 in the second paragraph
- Remove "and your team" from the last sentence of the second paragraph. Most judges do not judge on a team.
- In #2 in the list, add "on your dashboard" and link to the consent waiver tab on the judge dashboard so they can click through to sign

viviancan:
Updates on QA!

dboyer:
Confirmed text updates are on staging.

PR #3430, opened 2022-05-16
Added email flow for judges to receive an automatic email post registration

viviancan:
The program team requested for judges to receive an automatic email after they register. This flow follows the auto mentor email that is sent after an individual registers as a mentor. 

I added a new `welcome_judge` method as well as the email text. I have tried this locally and verified with the email popup in a new tab. 

Refs
#3427


devkonrad (approved):
The changes looks great. Its follow the system patterns and are very clean and clear. Thanks Vivian! :)

PR #3436, opened 2022-05-19
Update embed format for Vimeo videos

shaunxp20:
This will make our platform understand Vimeo links in the format: `https://vimeo.com/703480970/a4333a9b0c`.

Let me know if anyone has any questions! It's not the most straight forward one!

### DIFF
--- a/app/models/questions.rb
+++ b/app/models/questions.rb
@@ -95,10 +95,10 @@ class Questions
           }
         },
 
-        demo_video_id: submission.video_id(:demo),
+        demo_video_id: submission.video_id(:demo).first(5),
         demo_video_url: judge_embed_code_path(submission, piece: :demo),
 
-        pitch_video_id: submission.video_id(:pitch),
+        pitch_video_id: submission.video_id(:pitch).first(5),
         pitch_video_url: judge_embed_code_path(submission, piece: :pitch),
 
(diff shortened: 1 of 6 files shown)


app/models/questions.rb
```diff
@@ -95,10 +95,10 @@ def as_json(*args, &block)
           }
         },
 
-        demo_video_id: submission.video_id(:demo),
+        demo_video_id: submission.video_id(:demo).first(5),
         demo_video_url: judge_embed_code_path(submission, piece: :demo),
 
-        pitch_video_id: submission.video_id(:pitch),
+        pitch_video_id: submission.video_id(:pitch).first(5),
```
shaunxp20, line 101: This is needed because `video_id` now returns `703480970?h=a4333a9b0c` for these new Vimeo URLS (`https://vimeo.com/703480970/a4333a9b0c`). These `id`s are used in HTML to open the corresponding modal - the special chars (?=) weren't working w/ the modals, so this change will just use the first 5 characters for the modals ids. 😅

viviancan (approved):
LGTM! Tried this locally by viewing different submissions with different types of videos (vimeo & YT)

devkonrad (approved):
Great solution. Looks great to me! :)

Issue #3355, opened 2022-04-20
Update main navbar to be mobile responsive


viviancan:

viviancan:
![IMG_1ECFD6F943D9-1.jpeg](https://images.zenhubusercontent.com/606b0f4fd13d53dff3d14fa5/a18c38fa-b59a-4df2-abc9-1f91c5b9e63a)

dboyer:
@devkonrad This could be a good task for you to dig a bit more into Tailwind and some front-end work. 

As shown in the screenshot, the main navbar looks off on https://my.technovationchallenge.org/signup#/ and other pages where there are many buttons at the top. We need to adjust the styling so that the Donate button isn't appearing over other text.

devkonrad:
As discussed with @dboyer, we need to:
 - [ ] Adjust the bugs that remain open the burger menu itens on resize
 - [ ] Confirm if we will leave visible the donate for small sizes screen

dboyer:
We do not need to have the Donate button visible. It is fine to put all items behind the hamburger menu on small screens.

devkonrad:
> We do not need to have the Donate button visible. It is fine to put all items behind the hamburger menu on small screens.

Thanks, @dboyer! So i will wait for the code review :)

dboyer:
Confirmed on QA that this switches to the hamburger menu at small screen sizes

PR #3438, opened 2022-05-23
Adjusts Navbar responsivity

devkonrad:
Makes the navbar responsible. When detects that the screen is too small to classic navbar version, changes to the mobile one. This mobile navbar have a "burger menu" thats toogle between "show/hide" the menu itens. 

Changes:
modified:  app/javascript/new_registration/components/Banner.vue
 modified:  app/javascript/new_registration/components/Navbar.vue

Refs: https://github.com/Iridescent-CM/technovation-app/issues/3355


app/javascript/new_registration/components/Navbar.vue
```diff
@@ -1,68 +1,93 @@
...
-  //   const btn = document.querySelector("button.mobile-menu-button");
-  //   const menu = document.querySelector(".mobile-menu");
-  //
-  //   // Add Event Listeners
-  //   btn.addEventListener("click", () => {
-  //     menu.classList.toggle("hidden");
-  //   });
-  // }
+  methods: {
+    toogleMenu() {
```
viviancan, line 86: There is a typo here! Should be `toggleMenu` :)

viviancan (changes requested):
Looks good to me! Thanks for tackling this one Alex! I just saw 1 typo. I also left a question for deb on the ticket, wondering if we should remove the signup link so this is already the registration page. We can regroup at tomorrow's standup!

app/javascript/new_registration/components/Navbar.vue
devkonrad: > There is a typo here! Should be `toggleMenu` :)

Good catch, Vivian! Thank you :)

Issue #3446, opened 2022-05-25
Update heading names (logic) for students and add text for new scores explained


dboyer:
Split off #3410. Needs a wireframe

TO DO:

- [x] Rename headings to match new sections - Learning Journey, Demo, etc - can likely use content from #3380  
- [x] Adjust text associated with judging rubric

viviancan:
![Screen Shot 2022-05-25 at 12.27.16 PM.png](https://images.zenhubusercontent.com/606b0f4fd13d53dff3d14fa5/31068ad1-6c69-4d92-a047-97f5913876e9)

viviancan:
![Screen Shot 2022-05-25 at 12.27.29 PM.png](https://images.zenhubusercontent.com/606b0f4fd13d53dff3d14fa5/ac9febd9-7917-449a-8767-a344eff9f430)

viviancan:
![Screen Shot 2022-05-25 at 12.29.19 PM.png](https://images.zenhubusercontent.com/606b0f4fd13d53dff3d14fa5/93cfdc38-d5bc-44c6-a80a-dc92c0dce0f4)

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/3446?planning-poker) with ZenHub @devkonrad @shaunxp20 @viviancan

viviancan:
Current todos are complete 
 - Rename headings to match new sections - Learning Journey, Demo, etc - can likely use content from #3380
 - Adjust text associated with judging rubric

Current blockers: 
- Wireframe for the "rebrand" part of the ticket

I would be ok with splitting the rebrand work into its own ticket, as these (rebrand scores and the heading updates) are distinct pieces of work

Issue #3380, opened 2022-04-28
Judging admin issues


dboyer:
- [x] Add beginner to the division drop-down list at /admin/scores#/ - VC
- [x] Score summary pages (e.g. /admin/scores/620#/) does not have headings that match the new headings for judging
- [x] Not all scores appearing on score summary page (e.g. /admin/scores/620#/). Junior division, for example, is missing project name and description, demo, and user adoption plan
- [x] Adjust low score threshold to -  Senior: 25/90, Junior: 20/80, Beginner: 15/39 - SA
- [x] Low score threshold not calculating accurately. There are scores in the junior division of 46/80 and 42/30 flagged as too low but scores of 37/80 not flagged as too low.
<img width="747" alt="Screen Shot 2022-04-28 at 2 28 11 PM" src="https://user-images.githubusercontent.com/516764/165821875-555d4d25-4fab-4bd0-b315-553296deef34.png">

shaunxp20:
> Low score threshold not calculating accurately. There are scores in the junior division of 46/80 and 42/30 flagged as too low but scores of 37/80 not flagged as too low.

Both of the "seems to low" scores were completed on the 26th. The low score threshold calculations were updated after that. So those two scores aren't anything to worry about, but let's keep an eye out if new ones pop up like this.

Side notes: I think if someone updates those scores (edit a score or comment) that should re-trigger the score too low logic/flag. Also, I think it would be nice/useful to display the last updated date/time and/or completed at date/time in on the score details page in the admin. 😄

dboyer:
Confirmed on QA that beginner division filter is on the scores page and filtering correctly and that the correct questions and score totals are showing in each division.

I am still seeing the old headings of ideation rather than learning journey and entrepreneurship rather than business plan or adoption plan. 
<img width="668" alt="Screen Shot 2022-04-29 at 5 43 38 PM" src="https://user-images.githubusercontent.com/516764/166072953-e4e9ce4f-8a9b-4190-b0de-f5b1b039e47b.png">

viviancan:
That is how they are still stored behind the scenes! Ideation = Learning Journey & entrepreneurship = business plan. I can make a fix on the front end to display the updated titles, however, the backend will stay the same for now.

dboyer:
Eh, I don't think changing the labels is a huge priority. Thanks for explaining!

dboyer:
[...]

PR #3454, opened 2022-06-02
Updated score headings for student view scores page

viviancan:
This change was made to account for new rubric sections as well as the mismatched names (Ideation = Learning Journey, Entrepreneurship = User Adoption Plan VS Business Plan).

The same concern from #3380 came up again, where we need to have one "source of truth" for the naming issues. We have had to use a lot of view layer logic to display the correct names for the section. 

We should also discuss how we can "total" up the section points so it can be more dynamic compared to how the section points are totaled now in `app/models/submission_score.rb`

Refs Issue #3446

### DIFF
--- a/app/views/student/scores/_scores.html.erb
+++ b/app/views/student/scores/_scores.html.erb
@@ -23,37 +23,9 @@
   <% end %>
 </div>
 
-<div class="grid judging-banner">
-  <div class="grid__col-6">
-    <h3 class="judging-banner__header">
-      Scores Explained
-    </h3>
-
-    <dl>
-      <dt>1: Incomplete</dt>
-      <dd>The work is missing.</dd>
-
-      <dt>2: Needs Improvement.</dt>
-      <dd>The work needs major improvement.</dd>
-    </dl>
-  </div>
-
-  <div class="grid__col-6">
-    <dl>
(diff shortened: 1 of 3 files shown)


app/views/student/scores/_scores.html.erb
```diff
@@ -56,64 +43,25 @@
...
-        </dd>
-
-        <% if current_team.senior? %>
-          <dt>Entrepreneurship
+        <% questions.sections_for(division: score.team.division_name).each do |section| %>
+          <dt>
+            <% if section === "ideation" %>
+              Learning Journey
+            <% elsif section === "entrepreneurship" %>
+              <%= current_team.senior? ? "Business Plan" : "User Adoption Plan" %>
```
shaunxp20, line 51: I wonder if `sections_for` could have the logic/translations, then it could be the source of truth for section names. 💡 😃

shaunxp20 (approved):
Overall, looks good to me! Didn't test it locally, but should have enough time to put on QA for some feedback! 👍 😁 *excited! 🙌

app/views/student/scores/_scores.html.erb
shaunxp20: Hmm, or maybe a new method that returns section names? Just thinking out loud. 😃

Issue #3444, opened 2022-05-25
Rebrand judge main container certificate view


viviancan:

viviancan:
Refs #3411

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/3444?planning-poker) with ZenHub @devkonrad @shaunxp20 @viviancan

dboyer:
Updated text for judge dashboard for both people who received a certificate and those who didn't is [available here](https://docs.google.com/document/d/1Kd2z3Ju4drkm8gZmfCpN3HS6WItSQDTjmQBXfpWtBgs/edit#heading=h.y0kdco877i2r).

dboyer:
A few small things:

- When I hover over the sign-up link, it turns from blue to the older purple. I tried to find an example somewhere else of whether we've switched that to another color in the new branding palette but couldn't find one on the platform. My preference would be to get rid of the purple but if that's a new task, we can split it into a different ticket.
- Could there be a blank line between "ways for you to be involved" and the bullet point list? I think it would read easier.
- The sign-up text can link to http://eepurl.com/gzPycj


<img width="870" alt="Screen Shot 2022-06-07 at 9 33 29 AM" src="https://user-images.githubusercontent.com/516764/172393323-10fd538f-bd83-4728-9ad7-c777ae351b1a.png">

viviancan:
Update - Just submitted a PR with the above changes

PR #3461, opened 2022-06-07
Additional judge text updates + added general_newsletter_url variable

viviancan:
These changes were made based on QA feedback in #3444 

1. When hovering over text links that have the class `.tw-link` the text color will change to the new magenta that aligns with current branding
2. Added global tailwind style `.tw-flag-magenta` to `app/javascript/stylesheets/layout.css` so we can integrate updated styling for the "flags" that are displayed under certain conditions
3. Added a new env variable `GENERAL_NEWSLETTER_URL` which will link to the general newsletter sign up. This is different than `NEWSLETTER_URL` which links to the newsletter for the season updates. I think a better name for that variable would be something more specific like `SEASON_NEWSLETTER_URL` because currently it is not very clear.

TO DO
@viviancan - I will add the `GENERAL_NEWSLETTER_URL` variable to heroku for QA and PROD (if this approach is approved)

@shaunxp20 and @devkonrad you may also need to add the `GENERAL_NEWSLETTER_URL` to your local `.env`


shaunxp20:
> I think a better name for that variable would be something more specific like SEASON_NEWSLETTER_URL because currently it is not very clear.

Or maybe even `NEXT_SEASON_NEWSLETTER_URL` or `UPCOMING_SEASON_NEWSLETTER_URL`. 🤔 🤷

shaunxp20 (approved):
Awesome, looks good to me, thanks Vivian! 👍 🥳 🙌

Issue #3463, opened 2022-06-08
When clicking the "view scores" button on judging portal, app errors out


viviancan:

viviancan:
When the app is switched to Semifinals, as a judge I can view score details for finished quarterfinals scores. Right now, when I click the "view score" button, the app will error out. 
![Screen Shot 2022-06-08 at 8.05.09 AM.png](https://images.zenhubusercontent.com/606b0f4fd13d53dff3d14fa5/093434bd-0269-4dac-9e93-1bd0e2049e2e
)

dboyer:
Confirmed that this is now working on QA

PR #3464, opened 2022-06-08
Updated team link for mentor scope on the score details view in the judge portal

viviancan:
When the app is switched to Semifinals, as a judge I can view score details for finished quarterfinals scores. Right now, when I click the "view score" button, the app will error out.

The change was made in `app/views/admin/scores/show.en.html.erb`. I think the bug was happening at `send("#{current_scope}_team_path", @score.team)` because when logged in as a judge, the link would be `judge_team_path`, which does not exist. 

Refs #3463

### DIFF
--- a/app/views/admin/scores/show.en.html.erb
+++ b/app/views/admin/scores/show.en.html.erb
@@ -8,7 +8,7 @@
       <span>Team</span>
       <%= link_to_if current_scope == "mentor",
         @score.team_name,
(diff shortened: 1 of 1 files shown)


devkonrad (approved):
The change make sense. The fixes looks good to me. :)

shaunxp20 (approved):
Pulled it down, and verified it! It works! Good catch, and thanks for taking care of this Vivian! 👍 💯 🥳

Issue #3468, opened 2022-06-08
Upgrade Tailwind to V3


dboyer:
Documentation available at https://tailwindcss.com/docs/installation

dboyer:
Prioritize this for earlier in iteration to ensure longer time on QA for testing before deployment.

dboyer:
Not sure if this is Tailwind or some home page changes, but the globe on the home page is left aligned on QA now rather than centered. 
<img width="653" alt="Screen Shot 2022-06-28 at 12 07 53 PM" src="https://user-images.githubusercontent.com/516764/176227903-432d81ed-3aa0-4c1a-8834-498d96e22f49.png">

shaunxp20:
> Not sure if this is Tailwind or some home page changes, but the globe on the home page is left aligned on QA now rather than centered. This is only at the home page if you don't have the # after the main home page URL <img alt="Screen Shot 2022-06-28 at 12 07 53 PM" width="653" src="https://user-images.githubusercontent.com/516764/176227903-432d81ed-3aa0-4c1a-8834-498d96e22f49.png">

Interesting, I keep trying to remove the `#` in the URL and it keeps getting added back, so I can't reproduce this issue. ☹️  It's not a Tailwind page though, so maybe we should create another issue for it? 🤔

dboyer:
- [ ] Nick found the action button to create a team was greyed out until hovered over.

https://drive.google.com/file/d/1t3MgyAmwAKUotWuT-dviSMaadH9ZlB-8/view

- [ ] The to do list has hover text appearing on top of other list options
 
https://drive.google.com/file/d/1it4z-R40A0L7uP-J61Y4I-7Q7Kw0SubR/view

viviancan:
I think the "to do" list hover may be related to tailwind config on QA/Prod because locally I am unable to recreate this issue. 

<img width="1742" alt="Screen Shot 2022-06-30 at 11 25 36 AM" src="https://user-images.githubusercontent.com/29210380/176729110-3ddedc0d-324e-4571-994d-a53a0cf9e2dd.png">

viviancan:
Update on this ticket - We thought there were some config issues related to Tailwind v3 so Shaun has rolled this back. We may wait until the season is over to put this on QA/Prod to avoid any potential styling issues. Added a blocked tag for now.

dboyer:
Fine by me!

shaunxp20:
This one is back on QA! 🙌 😃

[...]

PR #3492, opened 2022-06-16
Upgrade Tailwind CSS to version 3.1

shaunxp20:
I followed the upgrade guide to get us on Tailwind version 3.1:
https://tailwindcss.com/docs/upgrade-guide

Refs: #3468


app/javascript/tailwind.config.js
```diff
@@ -1,6 +1,9 @@
 module.exports = {
-  purge: [],
-  darkMode: false,
+  content: [
+    './app/helpers/**/*.rb',
+    './app/javascript/**/*.{js,vue}',
+    './app/views/**/*',
+  ],
```
shaunxp20, line 6: More of an FYI - `content` is a new option, and as I understand it, it's where Tailwind will look for Tailwind CSS classes. If we're ever seeing missing styles, we'll want to check that the paths/files are included here.

I copied these settings from their Ruby on Rails guide (and also added Vue files):
https://tailwindcss.com/docs/guides/ruby-on-rails

Issue #3485, opened 2022-06-15
Close judge registration


dboyer:
We need to close judge registration when semifinal judging ends, which will require a deployment.

Judging closes on Thursday, June 23. Coordinate a time with @mdancho to line up closing registration and judging as close together as possible.

dboyer:
Since all registration is closed, we will redirect /signup#/ to the home page.

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/3485?planning-poker) with ZenHub @devkonrad @shaunxp20 @viviancan

viviancan:
Verifying when I go to /signup on QA I am redirected to the main landing page w/ season close text

dboyer:
Also confirmed on staging.

PR #3507, opened 2022-06-22
Redirect signups route to homepage

shaunxp20:
After this gets merged/deployed, it will redirect "signups" to go to the home page (as a way to disable registrations for everyone).

Refs: #3485


viviancan (approved):
Did not pull down locally, but looks good to me!! Thanks for handling this one, Shaun 💯

Issue #3486, opened 2022-06-15
Update home page text to reflect closing all registration


dboyer:
After we close registration at the end of semi-finals, we need to update the home page text. Checking with PT on what text changes are needed.

dboyer:
Home page text available at https://docs.google.com/document/d/1mG9hnI972ohnyGz1rlcRNl1-swoVXCqSmOkZCSO0_aM/edit?usp=sharing

viviancan:
Verifying the updated text on QA on the main landing page

viviancan:
Note** Judging must be set to "finished" in the Content & Settings

PR #3508, opened 2022-06-22
Updated home page text for when registration is closed and judging is finished

viviancan:
This change is a sort of "band-aid" until the rest of the registration toggles are hooked up. I think there will be some updates to this flow once that is complete. This change will update the text on the landing page once registration toggles are closed and judging is set to off. We already had the `off_season_splash` which I added back in and I also updated the text on that page.

Refs #3486


shaunxp20 (approved):
Only thing that popped out to me were some links and if they should open in a new window and some hard-coded dates - but neither of those are blockers since this has to be deployed tomorrow. So, I think it's good to go! Thanks Vivian! 🎉 🙌 🚀 :shipit:

PR #3509, opened 2022-06-23
Updated home page text for when registration is closed and judging is finished

viviancan:
2nd update to home page text after QA 

Refs #3486


dboyer:
I reviewed the dif and the changes look right to me.

Issue #3495, opened 2022-06-17
Add  "view details" functionality to student scores view


viviancan:

viviancan:
Refs #3455

dboyer:
Confirmed on QA that the view details is working on expected.

The text does blur together a bit. The section titles (demo, user adoption plan, etc) look to be the same size as the section total and judge feedback headings. Maybe the section titles could be bigger as they are in the admin view of the scores?

Let's see what other feedback we get from testing.

Issue #3476, opened 2022-06-09
Update student certificate specs


viviancan:

viviancan:
Refs #3449

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/3476?planning-poker) with ZenHub @devkonrad @shaunxp20 @viviancan

PR #3510, opened 2022-06-23
Added student view score details functionality + created score details page with rebrand

viviancan:
This change completes the student view scores/certs rebranded functionality. This change specifically includes the creation of the student score details page using the new rebrand. This approach was selected because originally we thought we could reuse the admin score details view. Ultimately we decided to go with a new rebranded version of the score details page since we are in this in between stage where not all user types are rebranded. The specs still need to be updated to account for these changes which has been captured in #3476 .

Refs: #3495

### DIFF
--- a/app/controllers/student/scores_controller.rb
+++ b/app/controllers/student/scores_controller.rb
@@ -22,5 +22,17 @@ module Student
 
       render template: 'student/scores/index'
     end
+
+    def show
+      submission_id = TeamSubmission.where(team_id: current_team.id)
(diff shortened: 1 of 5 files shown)


devkonrad (approved):
Looks good to me.

shaunxp20 (approved):
Awesome, looks good to me, didn't pull it down, but the video you posted, looks great! Thanks for taking care of this one, Vivian! 🙌 🎉 👏

Issue #3504, opened 2022-06-22
Create card component for search results


dboyer:
This component will be used to display search results and can be shared across pages such as team search and mentor search. An example is below.

We'll need to consider the width of the card component, particularly to ensure card layout works on multiple screen sizes.

<img width="520" alt="Screen Shot 2022-06-22 at 1 50 44 PM" src="https://user-images.githubusercontent.com/516764/175103780-75e45648-e5c2-4345-b3ca-b6e8b4f65221.png">

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/3504?planning-poker) with ZenHub @devkonrad @shaunxp20 @viviancan

devkonrad:
:) I published this PR for this issue: https://github.com/Iridescent-CM/technovation-app/pull/3519

PR #3503, opened 2022-06-22
Rebrands team search for students

devkonrad:
Rebrands team search for students
Captura de tela de 2022-07-12 16-45-34

Captura de tela de 2022-07-12 16-45-56

_NOTE: As the "Anywhere"  and "Near me" filter are excludent, i implemented a radio_button instead of a checkbox._

Changes:
modified: app/views/searches/rebrand/_location_options.en.html.erb
modified: app/views/team_searches/_filter_options.en.html.erb
modified: app/views/team_searches/_new.html.erb
modified: app/views/team_searches/_result.html.erb
modified: app/views/team_searches/_search.html.erb
modified: app/views/team_searches/rebrand/_filter_options.en.html.erb
modified: app/views/team_searches/rebrand/_result.html.erb
modified: app/views/team_searches/rebrand/_search.html.erb
modified: spec/features/student/find_a_team_spec.rb

Refs: https://github.com/Iridescent-CM/technovation-app/issues/3203

PR #3519, opened 2022-06-28
Rebrands search result card component

devkonrad:
This card will be used on the student rebranded search for Teams and Mentors. This issue scope is the card creation and tests, but the visual feedback will be available when we merge the PR #3503

Changes:
modified: app/javascript/components/ResultCard.vue
modified: spec/javascript/components/ResultCard.spec.js

Captura de tela de 2022-06-28 13-37-26

Refs: https://github.com/Iridescent-CM/technovation-app/issues/3504

### DIFF
--- a/app/javascript/components/ResultCard.vue
+++ b/app/javascript/components/ResultCard.vue
@@ -1,5 +1,5 @@
 <template>
-  <div class="card-result grid__col--bleed">
+  <div class="card-result">
     <img :src="cardImage" class="w-full" @error="imgBroken" />
 
     <div class="card-photo-placeholder hidden" :id="imgPlaceholderId">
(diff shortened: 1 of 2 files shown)


app/javascript/components/ResultCard.vue
```diff
@@ -0,0 +1,97 @@
...
+    },
+    computed: {
+      imgPlaceholderId() {
+        return `img-ph-${this.cardId}`;
+      }
+    }
+  }
+</script>
+
+<style lang="scss">
```
viviancan, line 87: I would recommend using tailwind styles here by incorporating the `@apply` directive! This will also keep the styles consistent with rebrand efforts we have completed so far! 

like `@apply hidden w-full....`

devkonrad: > I would recommend using tailwind styles here by incorporating the `@apply` directive! This will also keep the styles consistent with rebrand efforts we have completed so far!
> 
> like `@apply hidden w-full....`

Thank you, @viviancan! I will make the change :)

devkonrad:
After a functionality review by @dboyer, the following changes are needed:
- [ ]  Display default photo or a blank tile the size of the image space if the image uploaded by the team is broken
- [ ]  Remove the highlighting on the Division label on the team card and left align that text below the location

app/javascript/components/ResultCard.vue
```diff
@@ -0,0 +1,95 @@
+<template>
+  <div class="card-result grid__col--bleed">
```
viviancan, line 2: Is this (`grid__col--bleed`) a tailwind class?

viviancan (approved):
Besides my last comment, everything LGTM! Looking forward to seeing all these pieces come together. Thanks Alex 💯 !

app/javascript/components/ResultCard.vue
devkonrad: Hi Vivian! Oh, i forgot to update this one to the TW version, thanks for catch.

PR #3549, opened 2022-07-21
Update total score functionality and presentation

shaunxp20:
Here is a summary of changes:

- Fixing the total score calculation
- Adding a basic test for score calculations
- Updating the score summary page
  - Adding a new total score row for each section
  - Also, I did this pretty roughly since this page will get rebranded soon, hopefully!

### DIFF
--- a/app/views/admin/scores/show.en.html.erb
+++ b/app/views/admin/scores/show.en.html.erb
@@ -75,6 +75,14 @@
 
   <div class="admin-score-sections">
     <% questions.sections_for(division: @score.team.division_name).each do |section| %>
+      <% if section === "ideation" %>
+        <% section_name = "Learning Journey" %>
+      <% elsif section === "entrepreneurship" %>
+        <% section_name = @score.team.senior? ? "Business Plan" : "User Adoption Plan" %>
+      <% else %>
+        <% section_name = section %>
+      <% end %>
+
       <h3>
         <% if section === "ideation" %>
           Learning Journey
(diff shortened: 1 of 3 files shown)


devkonrad (approved):
The changes looks good. Thanks, @shaunxp20 !

app/views/admin/scores/show.en.html.erb
```diff
@@ -75,24 +75,41 @@
 
   <div class="admin-score-sections">
     <% questions.sections_for(division: @score.team.division_name).each do |section| %>
+      <% if section === "ideation" %>
```
viviancan, line 78: I'm not sure where this is happening, but verified on both admin and mentor view, but there is an extra "business plan"/ "user adoption plan" text header showing up right before the business plan/user adoption plan section

CleanShot 2022-07-21 at 13 50 34@2x

CleanShot 2022-07-21 at 13 47 21@2x

viviancan (approved):
Looks great, Shaun! 💯 Nice coverage too. I was able to test this on the Review Apps (which wow such an awesome addition!!) and only saw one issue with the extra heading showing up! I added the screenshot to the actual code review! Besides that, the changes are nice and very clear. Excited to see this on QA!

app/views/admin/scores/show.en.html.erb
shaunxp20: > I'm not sure where this is happening, but verified on both admin and mentor view, but there is an extra "business plan"/ "user adoption plan" text header showing up right before the business plan/user adoption plan section

Thanks for catching this! 💯 I noticed/fixed this for the other sections, but missed this one, good catch! 👍 🙌

PR #3557, opened 2022-07-29
Basic Setup for Prettier and ESLint

shaunxp20:
This will install ESLint and Prettier and setup some basic configuration.

Additionally, I added: 
- [eslint-config-prettier](https://github.com/prettier/eslint-config-prettier) which turns off ESLint rules, that according to the README: "are unnecessary or might conflict with Prettier", which as far as I understand it makes ESLint and Prettier more compatible

- [eslint-plugin-prettier](https://github.com/prettier/eslint-plugin-prettier) which from the README: "Runs Prettier as an ESLint rule and reports differences as individual ESLint issues." (This one is also included in a new Vue.js project)

And I created #3565 for some followup work/ideas and #3566 to integrate with CircleCI.


devkonrad (approved):
Thanks, @shaunxp20! I will try it locally, but the changes looks great to me. ;)

viviancan (approved):
Did not try locally, but looks good to me!! 💯 Woohoo really excited for this, Shaun!

Issue #3014, opened 2021-08-05
Need to make sure pitch video and demo videos are not identical urls


rsgonzal:

dboyer:
Checked on QA and confirmed that I could not enter the same URL for both videos. I tested on both a new submission and existing submission and was stopped from saving a duplicate URL in both cases.

dboyer:
One edge case identified during QA. My feeling is that it is not worth fixing this.

1. Enter a URL for the pitch video
2. Go to the demo video and add the same URL but delete the last character
3. The page will allow you to save those videos and both play as expected

In the example for the team "mon equipe" on QA, the last characters are =3s which perhaps refers to 3 seconds? I don't this is terribly likely to occur and I'm okay with pushing this issue to prod with this edge case still in place. What are your thoughts @viviancan

PR #3582, opened 2022-08-11
Compare pitch and demo video links to ensure they are different links

viviancan:
Refs #3014 

This change was needed to ensure pitch and demo video links were not identical. This has been an issue in the past with team submissions. Now, a team submission cannot be saved if the video URLs are the same.

This ticket ended up being a little more involved than just adding a validation to the model. 
- `app/controllers/concerns/team_submission_video_link_review_controller.rb` I had to check if the update was successful and if so create the team activity tracker record and redirect with a success message. If the the update was not successful then check for any error messages and update the alert message accordingly

-  In `app/models/team_submission.rb` I added a custom validation method `demo_and_pitch_video_links_are_different` that compares the demo video link and the pitch video link and adds an error message. This validation only runs if the demo_video_link and pitch_video_link are present. I came to this conclusion because there are essentially 4 scenarios for the video links 
           - Both links are blank - We don't need validation here because only 1 link can be saved at a time
           - Demo link is present & pitch link blank - Validation needed
           - Demo link is blank & pitch link is present - Validation needed
           - Demo link is present and pitch link is present - Validation needed

- In `app/controllers/admin/team_submissions_controller.rb` I added a flash message if the save was unsuccessful. It was kind of confusing if both links were the same you would be redirected to the edit page without any feedback

- Finally, I added testing for the demo and pitch video links to both the mentor and student flow. I also updated a few of the previous tests in order to account for these updates. 

Hopefully, this is in a good place for code review now 🤞 You can see the updated functionality below. This represents a scenario where both the demo link and pitch link are already present, but I am trying to update my pitch link now with the same link that is used for the demo video.

https://user-images.githubusercontent.com/29210380/184239219-7db9ddb4-f802-4c1b-b946-cd14c8a187d0.mp4

### DIFF
--- a/app/models/team_submission.rb
+++ b/app/models/team_submission.rb
@@ -259,8 +259,8 @@ class TeamSubmission < ActiveRecord::Base
     }, 
     allow_blank: true
 
-  validate :demo_and_pitch_video_links_are_different, on: :update, 
-           if: ->(s) { !s.demo_video_link.nil? || !s.pitch_video_link.nil? }
+  validate :demo_and_pitch_video_links_are_different, on: :update,
+           if: ->(s) { s.compare_demo_and_video_links? }
 
   def demo_and_pitch_video_links_are_different
     errors.add(:base, "Demo and pitch video links cannot be the same! Please add a valid video link.") if demo_video_link == pitch_video_link
@@ -301,6 +301,10 @@ class TeamSubmission < ActiveRecord::Base
 
   def random_id
     SecureRandom.hex(4)
+  end
+
+  def compare_demo_and_video_links?
+    (demo_video_link.present? && pitch_video_link.nil?) || (pitch_video_link.present? && demo_video_link.nil?)
   end
 
   def qualifies_for_participation?
--- a/app/models/team_submission.rb
+++ b/app/models/team_submission.rb
@@ -259,7 +259,8 @@ class TeamSubmission < ActiveRecord::Base
     }, 
     allow_blank: true
 
-  validate :demo_and_pitch_video_links_are_different, on: :update
+  validate :demo_and_pitch_video_links_are_different, on: :update, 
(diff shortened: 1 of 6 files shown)


devkonrad (approved):
Looks perfect to me, @viviancan! Thanks for take care of it! :)

app/models/team_submission.rb
```diff
@@ -258,6 +258,13 @@ def developed_on?(platform_name)
       with: /[a-zA-Z0-9][a-zA-Z0-9-]{1,61}[a-zA-Z0-9]\.[a-zA-Z]{2,}(.[a-zA-Z]{2,63})?/ 
     }, 
     allow_blank: true
+
+  validate :demo_and_pitch_video_links_are_different, on: :update,
```
shaunxp20, line 262: 👍 I like the name you picked! 🙌 🎉 👏

spec/features/mentor/submission_piece_spec.rb
```diff
@@ -181,6 +181,47 @@
...
+          piece: :demo_video_link
+        )
+      )
+    end
+
+    within(".pitch_video_link.incomplete") do
+      click_link "Add the pitch video link"
+    end
+
+    video_id = "XMzMyNzg3OTY1Mg"
```
shaunxp20, line 215: Just a minor thing, but I had to do a double-take to make sure the `video_ids` were the same, I wonder if having only one `video_id` would be better or making this one `duplicate_video_id` or something? 🤔 😅 #JustARandomThought!

shaunxp20 (approved):
Looks good to me! Nice write-up explaining it all! Great improvements/catches too (regarding the admin and controller stuff)! And sweet tests! 💯 👏 🎉 🙌 :shipit:

spec/features/mentor/submission_piece_spec.rb
viviancan: Thanks for the suggestion 👍  I'll make that change!

Issue #3578, opened 2022-08-09
Rebrand reset password page


dboyer:
The reset password page at https://my.technovationchallenge.org/password_resets/new#/ has the old branding.

<img width="1440" alt="Screen Shot 2022-08-08 at 8 06 40 PM" src="https://user-images.githubusercontent.com/516764/183535051-59095e68-c9a2-488b-8d7d-bae482a1a970.png">

There's no existing mockup so let's use the general layout of the sign-in page.

![login](https://user-images.githubusercontent.com/516764/183535625-ee9515cc-e740-4f1c-8bd7-a165588dda2f.png)

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/3578?planning-poker) with ZenHub @devkonrad @shaunxp20 @viviancan

dboyer:
Follow the same pattern and use the same energetic components/integration as the sign-in page.

devkonrad:
Created the PR https://github.com/Iridescent-CM/technovation-app/pull/3601 to solve this issue. :)

dboyer:
Confirmed this page is rebranded on QA

PR #3601, opened 2022-08-23
Rebrands the password reset screen

devkonrad:
This PR rebrands the Reset Password Screen.

Captura de tela de 2022-08-23 08-13-28

Refs.: #3578

### DIFF
--- a/app/views/password_resets/new.html.erb
+++ b/app/views/password_resets/new.html.erb
@@ -1,12 +1,13 @@
 <% provide :title, "Reset your password" %>
-<% provide :main_header, render('application/templates/tg_green_header', heading: "Welcome Back") %>
-<% provide :global_nav_links, render('application_rebrand/global_nav_links') %>
+<% provide :main_header, render("application/templates/tg_green_header", heading: "Welcome Back") %>
+<% provide :global_nav_links, render("application_rebrand/global_nav_links") %>
 
 <div class="w-11/12 lg:w-1/2 mx-auto">
-  <%= render layout: 'application/templates/dashboards/energetic_container', locals: { heading: t("views.password_resets.new.title")} do %>
+  <%= render layout: "application/templates/dashboards/energetic_container", locals: {heading: t("views.password_resets.new.title")} do %>
+
     <%= simple_form_for @password_reset do |f| %>
-      
-      <%= f.input :email, label_html: {class: 'tw-label'} %>
+
+      <%= f.input :email, label_html: {class: "tw-label"} %>
 
       <%= f.button :submit,
         t("views.password_resets.new.submit"),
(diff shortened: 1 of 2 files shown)


dboyer:
@devkonrad Looks good! Could you check if the font size on the button text is the same font size as the button text on the sign in and register pages? I'd like them to be consistent.

devkonrad:
> @devkonrad Looks good! Could you check if the font size on the button text is the same font size as the button text on the sign in and register pages? I'd like them to be consistent.

Hi, @dboyer! I had the impression that the button font size was bigger than the sign up, too. But i used the exact same class, I guess that its a optcal illusion because here we have more text than the sign up screen.

dboyer:
Interesting! That makes sense that the amount of text could fool the eye. Thanks for checking!

viviancan (changes requested):
Hey Alex! Thx for tackling this one! I am leaving one comment on the ticket about the font for the "Email" label! Besides that, it looks good!

viviancan:
The "email" label font should match the updated font like on the sign-in page. I think this is our old branding font :) 
CleanShot 2022-08-24 at 09 54 28@2x

devkonrad:
> @devkonrad The "email" label font should match the updated font like on the sign-in page. I think this is our old branding font :) CleanShot 2022-08-24 at 09 54 28@2x
> 
> Example below CleanShot 2022-08-24 at 10 01 25@2x

Good caught, @viviancan ! I will fix it! ;)

app/views/password_resets/new.html.erb
```diff
@@ -1,15 +1,16 @@
 <% provide :title, "Reset your password" %>
+<% provide :main_header, render('application/templates/tg_green_header', heading: "Welcome Back") %>
```
shaunxp20, line 2: I tried this one locally and I was able to reset a password! 👍 The one thing popping to me is this header "Welcome Back", maybe that should say something else for the password reset page? 🤔  @dboyer

dboyer: Hmmmm.......good question. It's not the best fit but I'm also not coming up with anything better so let's keep it as "welcome back" for now. Probably if they are resetting a password they've been on the site at least once so it's not _wrong_ to welcome them back, just not expected.

shaunxp20 (approved):

devkonrad:
> > @devkonrad The "email" label font should match the updated font like on the sign-in page. I think this is our old branding font :) CleanShot 2022-08-24 at 09 54 28@2x
> > Example below CleanShot 2022-08-24 at 10 01 25@2x
> 
> Good caught, @viviancan ! I will fix it! ;)

Fixed it yet, @viviancan! :)

PR #3610, opened 2022-08-29
Update default team image

shaunxp20:
For teams that are using the default team image:
- if the team is new (>= 2023) they will get the new (star hands) default team image
- if the team is old they will get the old (collage style) team image


devkonrad (approved):
Looks perfect, @shaunxp20! :)

viviancan (approved):
Did not try locally, but looks good to me! Thanks for tackling this one Shaun 💯

PR #3611, opened 2022-08-29
Update invite summary page for mentors

shaunxp20:
This will add a simple conditional to either render a new rebranded view (for students) or the old one (for mentors).


viviancan (approved):
Did not try locally, but seems straightforward and follows previous pattern! Thx, Shaun 🚀

Issue #3605, opened 2022-08-24
Complete the filestack image upload process on student profile - Saving image ref to DB 


viviancan:

viviancan:
Res #3594

viviancan:
This process is working as expected now.
As a student, I can select a photo from my local machine through the filestack picker and update my profile image. Filestack returns a filelink (which is the filestack reference URL) that points to the image in the s3 bucket. This link is saved in the DB as the profile_image. I still need to write up a TLDR, but the functionality of this ticket is complete. I also need to clean up my branch before it's ready for code review!

viviancan:
@dboyer @shaunxp20 
These are the two versions for filestack profile image upload. Do you all have a preference? I have been going back and forth between the two options. Let me know what you all think! 

V1 - Uploading an image on the basic information page

https://images.zenhubusercontent.com/29868115/2cc33e5f-94a9-4d5e-bb40-537a9933cf36/filestack_upload_v1_on_profile.mp4

v2 - Uploading profile image through basic details page (where other profile info is edited)

https://images.zenhubusercontent.com/29868115/7bb3f0e5-e844-4981-a5ba-56ab44c92323/filestack_upload_v2_on_basic_details.mp4

dboyer:
Hmmmmmmmm......this is tough. From a UX flow perspective, I pick the first one as I've generally found that most images on websites are changed via a button right by the image, not as part of another form. I don't feel super strongly about it though so if there are technical reasons to prefer v2, I'm okay with that.

After a user uploads a new photo in v1, do they have to click the save button? Uploading it in the modal doesn't automatically save it?

viviancan:
The original flow got around the "save" button because it would create a job that would be processed at a later time.  Attached is the old flow. 

I can touch base with shaun to discuss the flow of keeping it on the main my profile page without the save btn! 

https://images.zenhubusercontent.com/29868115/373834d3-07c9-4557-9352-b6bcb9a9f968/original_profile_photo_upload_2.mp4

viviancan:
@dboyer Here is an updated recording that demonstrates the flow where the image is automatically saved without the user having to click save! It does seem to be much smoother, and I do agree with your previous comment regarding the typical UX flow!

 https://images.zenhubusercontent.com/29868115/8c798d44-773c-4e57-bf45-65a93b209626/updated_filestack_profile_img_upload.mp4

dboyer:
[...]

Issue #3594, opened 2022-08-17
Integrate Filestack on a branch


dboyer:
Depends on outcome of questions from #3501.

- [x] Install gem

- [x]  Get picker working in student profile (or one area where photo uploads are) so that a photo can be uploaded by filestack (not saved to the DB)

- [x] Decide if Filestack is handling S3 link or if we will use current system of how we use S3

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/3594?planning-poker) with ZenHub @devkonrad @shaunxp20 @viviancan

viviancan:
Confirming I was able to install filestack gem and integrate the filepicker widget on the student profile. I am able to upload/search/transform an image. This has not been set up to save the image yet. I will need to investigate the S3/filestack/rails workflow more. 
![filestack-student-profile.gif](https://images.zenhubusercontent.com/606b0f4fd13d53dff3d14fa5/8ecb753c-9daa-4334-8482-0a2a5c398c16)

viviancan:
The filestack engineering account has been configured and connected to AWS S3. A new IAM user was created with specific properties. The policy also allows for filestack to connect to multiple buckets which is necessary for us since we have different environments. This has been verified through a basic proof of concept. The flow is very streamlined and is almost instant. I have also verified the images are in the correct s3 buckets. Filestack returns a reference URL that points to the s3 image URL. 

I am confident in closing this ticket and adding an additional ticket to handle the fallout tasks.

Here are some resources that were helpful while integrating this feature. 

- [Filestack s3 buckets](https://www.filestack.com/docs/uploads/storage/#amazon-s3)
- [StoreOptions](https://www.filestack.com/docs/uploads/pickers/web/#store-options)
- [PickerStoreOptions Docs](https://filestack.github.io/filestack-js/interfaces/pickerstoreoptions.html)
- [Filestack-Rails gem](https://github.com/filestack/filestack-rails)

viviancan:
[...]

PR #3619, opened 2022-09-09
Completed filestack image upload process for profile images 

viviancan:
This change refs #3605 and was the 2nd step after #3594 

This change completes the filestack image upload process for profile images. There were 3 main changes
1. `app/views/profiles/_filestack_profile_image_tag.html.erb` - This snippet is what allows the filestack image to be displayed in the view. The `filestack_image` tag is part of the rails/js filestack gem. 
2. `app/views/profiles/_filestack_profile_image_upload.en.html.erb` This file contains the logic for filestack upload. Once the file is finished uploading, the `onFileUploadFinished` callback function is run that makes the ajax call to update the DB
3. removes the ImageProcessor on the profile_image and also allows for profile_image to be added/updated


app/views/profiles/_filestack_profile_image_upload.en.html.erb
```diff
@@ -0,0 +1,43 @@
+<%= javascript_tag do %>
```
viviancan, line 1: I tried adding this logic to its own JS file, but I think something weird was happening with the loading/webpacker/etc, where the function was always undefined, so I left it here instead.

app/models/account.rb
```diff
@@ -697,8 +693,26 @@ def reminded_about_survey!
     )
   end
 
+  def default_image_url
```
viviancan, line 696: I'm not sure if this chunk of logic should live somewhere else 🤔 Feedback would be great here!

shaunxp20: I think this is a good for now, still wrapping my head around everything, a few things I noticed:
- the methods not used outside of `account` should be `private`
- passing `profile_image` to `s3_profile_image_url` and `determine_profile_image_url` shouldn't be necessary

And the thought to just put everything in one method also crossed my mind. 😁 🤔 😅

app/views/profiles/_filestack_profile_image_upload.en.html.erb
```diff
@@ -0,0 +1,43 @@
...
+  function onFileUploadFinished(data) {
+    $("#account_profile_image").val(data.url);
+
+    $.ajax({
```
shaunxp20, line 5: Def not a blocker, but we could use the Fetch API here.

shaunxp20 (approved):
LGTM! No blockers (except the existing images not working, but that can be resolved after). Really excited to see this coming together! 👏 💯 🥳 🎈 🙌

app/views/profiles/_filestack_profile_image_upload.en.html.erb
viviancan: Sounds good, I will look at this process with my update branch :)

app/models/account.rb
viviancan: I updated this!

PR #3580, opened 2022-08-09
Change de accepting team invites opt-in for new mentors

devkonrad:
Currently the default value for "accepting_team_invites" for mentors is "True". This PR changes this value to "False", so each mentor have a choice do decide if want to accept or not invitations.

Refs.: #3249

Issue #3249, opened 2022-02-02
Do not automatically opt-in mentors to the "connect to mentors" platform


dboyer:
Currently, mentors are automatically opted in to "allow other mentors to find you in search results and connect via email" via a check-box on their profile (/mentor/profile#!basic).

We should remove the default check-box so that mentors have to choose to opt-in to having their info shared with other mentors. The other two boxes can be checked by default as they are used for team matching and do not share email addresses with other users.

<img width="670" alt="Screen Shot 2022-02-02 at 10 28 23 AM" src="https://user-images.githubusercontent.com/516764/152184361-30659d78-3662-4608-b6a3-246a727544b5.png">

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/3249?planning-poker) with ZenHub @devkonrad @shaunxp20 @viviancan

devkonrad:
Created PR https://github.com/Iridescent-CM/technovation-app/pull/3580 to tackle with this issue. ;)

viviancan:
QA Steps
1. Register as a new mentor
2. Complete mentor onboarding (location, training, & consent waiver)
3. View mentor profile by clicking "Profile" link in subnav
4. Verify the "Allow other mentors to find you in search results..." checkbox is not checked

viviancan:
@dboyer @shaunxp20 

GM! Something very strange is happening with the "accepting team invites" field. When a new mentor is created, the correct process is happening where "connect with mentors" is automatically set to false. However, I noticed (and verified in the DB on QA) that the "accepting team invites" field is now automatically being set to false as well - which is what we don't want! The strange part is that locally this behavior is not replicated and the PR I made does not alter the "accepting team invites." Unfortunately, I am not sure if this change caused that to happen or if it was an existing change. 

I checked the DB and there were 3 mentor accounts created between this change and previous PRs. I also cross-referenced those accounts with the "activity tracker" table and what's interesting is there is 1 mentor account where it looks like the "accepting team invites" field was automatically set to false because there is not an associated logged activity. When you update the searchable checkboxes an update activity gets logged. So it kind of seems like this could be related to something else? The other 2 accounts have the associated account updated logged. A lot of ifs here 🤔

[...]

PR #3631, opened 2022-09-14
3249 Sets the default value for mentors connecting with other mentors to false

viviancan:
Refs #3249 

This change was needed so mentors would not be automatically opted in to the "connect with mentors" feature. Now a mentor will have to check the appropriate checkbox to connect with other mentors.

Once this is merged, we can delete/close PR #3580 since this change will update the DB

### DIFF
--- a/spec/features/mentor/connect_with_mentors_spec.rb
+++ b/spec/features/mentor/connect_with_mentors_spec.rb
@@ -1,0 +1,12 @@
+require "rails_helper"
+
+RSpec.feature "Mentors connect with other mentors" do
+  scenario "A mentor has finished onboarding and views their profile for the first time" do
+    mentor = FactoryBot.create(:mentor, :onboarded)
+
+    sign_in(mentor)
+    visit mentor_profile_path
+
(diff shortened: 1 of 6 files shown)


spec/features/mentor/connect_with_mentors_spec.rb
```diff
@@ -0,0 +1,12 @@
+require "rails_helper"
+
+RSpec.feature "Mentors connect with other mentors" do
+  scenario "A mentor has finished onboarding and views their profile for the first time" do
```
viviancan, line 4: Not sure if this scenario description should be updated...
Yes, the first time a mentor logs in this is the expected behavior, but if a mentor has not opted in, this would also be the expected behavior #RandomThoughts! 🤔

shaunxp20: Imo, the scenario should include what's being `expect`ed. Also, nice spec! 👏 🎉

shaunxp20 (approved):
LGTM! Thanks for taking care of this one, Vivian! 👍 💯 🥇 🚢

PR #3634, opened 2022-09-15
Mentor search updates

shaunxp20:
General overview of the changes:
- Inline all of the partials
- General styling updates/clean-up
- Fix labels (location and search by name)
- Update headers
- Change `<p>` tags around the form elements to `<div>` tags (since they're
  not paragraphs)
- Update expertise to use flex box so they display nicer

-----


viviancan (approved):
Did not try locally, but LGTM! I like the approach of bringing everything into the rebrand partial! I think it would be good to get another set of 👀  on it/feedback from PT! Thx Shaun!

Issue #3603, opened 2022-08-23
Update platform email text for 2023 season


dboyer:
Once approved by VET and PT, we'll need to make any necessary email text updates. [Doc with text is available here](https://docs.google.com/document/d/1fB4XXgQ_oc14sdwg-sgqnoxQJM3Nm8Sf1XY_YpYMo7A/edit#).

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/3603?planning-poker) with Zenhub @shaunxp20 @viviancan

dboyer:
The text for the emails is ready to update:

- [ ] Welcome Email - [Students (Jr/Sr only)](https://docs.google.com/document/d/1fB4XXgQ_oc14sdwg-sgqnoxQJM3Nm8Sf1XY_YpYMo7A/edit#heading=h.n9ff7zr68rrs)
- [ ] Welcome Email - [Beginner Division](https://docs.google.com/document/d/1fB4XXgQ_oc14sdwg-sgqnoxQJM3Nm8Sf1XY_YpYMo7A/edit#heading=h.l191k1iv37zx)
- [ ] [Parents Thank You](https://docs.google.com/document/d/1fB4XXgQ_oc14sdwg-sgqnoxQJM3Nm8Sf1XY_YpYMo7A/edit#heading=h.mizgvsy4x0fl) (for Jr/Sr participants)
- [ ] Parent Consent [Signature Request](https://docs.google.com/document/d/1fB4XXgQ_oc14sdwg-sgqnoxQJM3Nm8Sf1XY_YpYMo7A/edit#heading=h.8j0op1ikf5bv)

viviancan:
@dboyer Ty for the direct links to each of the sections, it was very helpful!

dboyer:
Tested on QA:

- [x] Welcome Email - [Students (Jr/Sr only)](https://docs.google.com/document/d/1fB4XXgQ_oc14sdwg-sgqnoxQJM3Nm8Sf1XY_YpYMo7A/edit#heading=h.n9ff7zr68rrs)
- [X] Welcome Email - [Beginner Division](https://docs.google.com/document/d/1fB4XXgQ_oc14sdwg-sgqnoxQJM3Nm8Sf1XY_YpYMo7A/edit#heading=h.l191k1iv37zx)
[...]

PR #3637, opened 2022-09-19
3603 - Email text updates for season opening

viviancan:
Refs #3603 

The biggest change here is in `app/views/team_mailer/join_request_accepted.en.html.erb`

A request was made to add additional text including next steps once a mentor is approved/accepted to join a team. This was a bit trickier than the other email updates because this email template is triggered after different actions. I added a conditional to check the `role_name` (which is also dynamic and set in `config/locales/mailers.en.yml`. In this case, the role_name is `their_mentor`.

I am requesting review on this since it was a little more involved than just a text change :)

### DIFF
--- a/app/models/season.rb
+++ b/app/models/season.rb
@@ -9,7 +9,7 @@ class Season
   end
 
   def self.deadline
-    day = ImportantDates.submission_deadline.day.ordinalize
+    day = ImportantDates.submission_deadline.day
     ImportantDates.submission_deadline.strftime("%B #{day}")
   end
 
--- a/app/views/team_mailer/join_request_accepted.en.html.erb
+++ b/app/views/team_mailer/join_request_accepted.en.html.erb
@@ -22,6 +22,34 @@
           <%= link_to "Visit Your New Team", @url, class: "btn-primary", itemprop: "url" %>
         </td>
       </tr>
+      <% if @role_name.include? 'mentor' %>
+        <tr>
+          <td class="content-block">
+            <p>
+              Next steps:
+              <br>
+              Now that you have joined the team, it is time to introduce yourself via email. Here is an example:
+            </p>
+            <p>Hi <%= @team_name %>,</p>
+            <p>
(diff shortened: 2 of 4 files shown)


app/models/season.rb
```diff
@@ -9,7 +9,7 @@ def initialize(year)
   end
 
   def self.deadline
-    day = ImportantDates.submission_deadline.day.ordinalize
+    day = ImportantDates.submission_deadline.day
```
viviancan, line 12: `.ordinalize` was removed so the day suffix for the submission deadline matched the other days listed in key dates

app/views/team_mailer/join_request_accepted.en.html.erb
```diff
@@ -22,6 +22,34 @@
           <%= link_to "Visit Your New Team", @url, class: "btn-primary", itemprop: "url" %>
         </td>
       </tr>
+      <% if @role_name.include? 'mentor' %>
```
viviancan, line 25: The `role_name` for mentors is `thier_mentor` and is set in `config/locales/mailers.en.yml`

viviancan: I wasn't sure if I should just use the variables that were already available or if I should access the scope here instead.

shaunxp20: I think it's okay. I think it's not clear as it could be (at least to me) b/c this mailer is being used for both student and mentors, at some point making them separate mailers would be a nice refactoring, imho. 😃

shaunxp20 (approved):
LGTM! Thanks Vivian! 👍 🙌 🥳 🚢

app/views/team_mailer/join_request_accepted.en.html.erb
viviancan: Yes, I agree! This was weirdly tricky to figure out! I logged the refactoring as a new issue - #3638

PR #3639, opened 2022-09-20
Remove "profile detail" section for mentors

shaunxp20:
This will remove the "profile detail" link from the mentor dashboard. In order to make this change, I also had to update the "Complete your Profile" link (which I left a comment about), and remove the Next button on the "Choose your profile type" page (which I also left a comment about). 😄

### DIFF
--- a/app/javascript/mentor/DashboardMenu.vue
+++ b/app/javascript/mentor/DashboardMenu.vue
@@ -2,7 +2,7 @@
   <ul class="tabs__menu tabs-menu__parent-menu padding--none">
     <tab-link
       :class="registrationTabLinkClasses"
-      :to="{ name: 'basic-profile', meta: { active: registrationPagesActive } }"
+      :to="{ name: 'data-use', meta: { active: registrationPagesActive } }"
       :condition-to-enable="true"
       :condition-to-complete="true"
     >
--- a/app/javascript/registration/components/ChooseProfile.vue
+++ b/app/javascript/registration/components/ChooseProfile.vue
@@ -59,13 +59,7 @@
       >
         Back
       </a>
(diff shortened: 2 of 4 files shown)


app/javascript/mentor/DashboardMenu.vue
```diff
@@ -2,7 +2,7 @@
   <ul class="tabs__menu tabs-menu__parent-menu padding--none">
     <tab-link
       :class="registrationTabLinkClasses"
-      :to="{ name: 'basic-profile', meta: { active: registrationPagesActive } }"
+      :to="{ name: 'data-use', meta: { active: registrationPagesActive } }"
```
shaunxp20, line 5: This is the "Complete your Profile" link (the header in the sidebar nav), it was linking to the "profile detail" page, but that was removed, so I just updated it to link to the first subnav item, the "data agreement".

app/javascript/registration/components/ChooseProfile.vue
```diff
@@ -59,13 +59,7 @@
       >
         Back
       </a>
-      <button
-        class="button"
-        :disabled="!nextStepEnabled"
-        @click.prevent="handleSubmit"
-      >
-        Next
-      </button>
```
shaunxp20, line 68: This page is the "Choose your profile type" page from the old registration flow that is reused on the mentor dashboard! The Next button was going to the "profile detail" page which was removed, so I removed the Next button since there is no next step. 😅

viviancan (approved):
Did not try locally, but it looks good to me! I remember from the student profile, there were a lot of similar pieces to remove! Thx for tackling this one, Shaun :)

Issue #3644, opened 2022-09-21
On mentor dashboard profile image does not display - path is broken. The profile image does display on the my profile. 


viviancan:

viviancan:
In the mentor profile, the profile image (avatar) does not display on the dashboard header. The profile image does display on the profile. See images below. I suspect the path is incorrect or some wired are crossed since the dashboard header is in the vue world (a vue component) and the profile is in the rails world.

Starting place: `app/javascript/components/DashboardHeader.vue`

![CleanShot 2022-09-21 at 15.07.22.png](https://images.zenhubusercontent.com/606b0f4fd13d53dff3d14fa5/255f8b21-e0ad-4261-ae7c-8c468023187d)

![CleanShot 2022-09-21 at 15.08.27.png](https://images.zenhubusercontent.com/606b0f4fd13d53dff3d14fa5/881aba79-10f8-4b27-a749-93fbca913f63
)

Issue #3613, opened 2022-08-31
Finalize/complete filestack flow (cleanup tasks - more info to come)


viviancan:
- [x] Admin portal ability to update profile images
- [x] Decide what "sources" are allowed for profile image upload (social media, webcam, web search, etc)
- [x] What do we do about icons 🤔
- [x] Verify s3 bucket policies for QA
- [x] Verify s3 bucket policies for prod
- [x] Verify team photo 
- [x] Decide if we will have default profile image (similar to team photo)
- [x] Add some styling to the "change image" button as it is currently just text and doesn't look like a button
~~- [x] The page to upload via URL looks like the button and text field aren't aligned
<img width="465" alt="Screen Shot 2022-09-21 at 10 01 54 AM" src="https://user-images.githubusercontent.com/516764/191525145-b51bafe8-81a6-458f-82eb-4398dca311b0.png">~~
We are not able to update this, as it is determined by filestack UI (VC)

viviancan:
Refs #3605

viviancan:
Added functionality to admin portal to update the participant profile image 

 https://images.zenhubusercontent.com/29868115/2557d8b0-1692-4b7b-9fb1-ba4ef8c24701/admin_filestack_profile_img_upload.mp4

viviancan:
@dboyer 
2 questions - 
1.  What "sources" are allowed for profile image upload (social media, webcam, web search, etc)
2. What do we do about icons? Filestack allows for us to have a "custom source" to choose from, but I think we need the next higher paid plan.

dboyer:
Answers!

1. Please remove web search, Facebook, and Instagram from the available sources
2. There is no need to keep the custom icons. They're super cute but no reason they have to be there.

viviancan:
- [ ] ChA Profile: app/views/chapter_ambassador/profiles/show.en.html.erb
- [ ] Judge Profile: app/views/judge/profiles/rebrand/_basic.en.html.erb
- [ ] Mentor Profile: app/views/mentor/profiles/show.html.erb
- [ ] FetchAPI: app/views/profiles/_filestack_profile_image_upload.en.html.erb

viviancan:
Verifying the clean-up tasks are on QA. You should now be able to update profile image for all profile types. For accounts that do not have a profile image, you should see the hedgehog default image. If there is a previous profile image that should also display correctly.

viviancan:
I added a few notes to our documentation repo!

[...]

PR #3645, opened 2022-09-21
Filestack profile image cleanup - includes all profile types

viviancan:
Refs #3613 

This is the first part of the general filestack cleanup.

Changes include
1. Style updates - Something I did not originally consider is that some of the profile types that have not been update will be using filestack so custom styles had to be included since tailwind was not accessible in those profile types
2. Updated ChA and mentor profile to use rails image tag
3. Updated `app/views/profiles/_filestack_profile_image_tag.html.erb` to only use the rails tag for now. I would like to use the filestack img tag eventually, however there was an issue with the placeholder image. This has been logged in #3643 
4. Added the placeholder image for accounts without a profile img

Note ** I also needed to comment out the mentor dashboard header profile image since it was causing tests to fail. This issue has been logged in #3644. The profile image avatar is in `app/javascript/components/DashboardHeader.vue`

### DIFF
--- a/app/javascript/components/DashboardHeader.vue
+++ b/app/javascript/components/DashboardHeader.vue
@@ -33,12 +33,12 @@
     <div class="grid__col-sm-6 grid__col--bleed-y grid--align-end">
       <div class="grid__cell">
         <h1 class="page-heading">
-          <img
-            :src="currentAccountAvatarUrl"
-            class="profile-image"
-            width="40"
-            height="40"
-          />
+<!--          <img-->
+<!--            :src="currentAccountAvatarUrl"-->
+<!--            class="profile-image"-->
+<!--            width="40"-->
+<!--            height="40"-->
+<!--          />-->
 
           {{ currentAccountName }}
 
--- a/app/javascript/stylesheets/buttons.css
+++ b/app/javascript/stylesheets/buttons.css
@@ -25,3 +25,7 @@ a.link-button-danger, .submit-button-danger {
 a.link-button-neutral, .submit-button-neutral {
     @apply text-gray-500 hover:text-gray-500 bg-gray-200 hover:bg-gray-300 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-gray-400
 }
+
+.filestack-picker-btn {
+    @apply w-auto mr-auto mt-4 text-sm rounded bg-zinc-300 p-2
+}
--- a/app/javascript/stylesheets/buttons.css
+++ b/app/javascript/stylesheets/buttons.css
@@ -25,7 +25,3 @@ a.link-button-danger, .submit-button-danger {
 a.link-button-neutral, .submit-button-neutral {
     @apply text-gray-500 hover:text-gray-500 bg-gray-200 hover:bg-gray-300 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-gray-400
 }
-
-.filestack-picker-btn {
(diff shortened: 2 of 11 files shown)


app/javascript/stylesheets/buttons.css
```diff
@@ -25,3 +25,7 @@ a.link-button-danger, .submit-button-danger {
 a.link-button-neutral, .submit-button-neutral {
     @apply text-gray-500 hover:text-gray-500 bg-gray-200 hover:bg-gray-300 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-gray-400
 }
+
+.filestack-picker-btn {
```
viviancan, line 29: I needed to add this with the apply directive bc the inline tailwind styles were not applied as expected

app/javascript/stylesheets/layout.css
```diff
@@ -48,4 +48,8 @@ html{
 
 .tw-flag-magenta{
     @apply text-white block p-2 text-right text-base font-bold shadow-sm bg-tg-magenta
+}
+
+.filestack-profile-img {
```
viviancan, line 53: Same issue as above -  added this with the apply directive bc the inline tailwind styles were not applied as expected

app/views/profiles/_filestack_profile_image_tag.html.erb
```diff
@@ -1,5 +1,3 @@
-<%= filestack_image current_profile.profile_image_url,
-                    transform: filestack_transform.resize(width: 200, height: 200) ,
-                    id: 'filestack-img-tag' %>
-
-<div id="thumbnail-container"></div>
\ No newline at end of file
+  <%= image_tag current_profile.profile_image_url,
```
viviancan, line 1: I would like to return to the filestack image tag in order to utilize other features like the resize. Logged in #3643

app/javascript/components/DashboardHeader.vue
```diff
@@ -33,12 +33,12 @@
     <div class="grid__col-sm-6 grid__col--bleed-y grid--align-end">
       <div class="grid__cell">
         <h1 class="page-heading">
-          <img
-            :src="currentAccountAvatarUrl"
-            class="profile-image"
-            width="40"
-            height="40"
-          />
+<!--          <img-->
```
viviancan, line 36: Temporarily commenting this out in order to pass tests. This will need to be solved!

app/models/account.rb
```diff
@@ -695,8 +695,8 @@ def reminded_about_survey!
 
   def profile_image_url
     if profile_image.blank?
-      "https://s3.amazonaws.com/#{ENV.fetch("AWS_BUCKET_NAME")}/placeholders/1.svg"
-    elsif profile_image.include? "filestackcontent"
+      "placeholders/avatars/1.svg"
```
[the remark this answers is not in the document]
shaunxp20: 🦔 👍

shaunxp20 (approved):
LGTM! The only thing that crossed my mind is if the filestack styles should all be together in a separate stylesheet file. Thanks for the writeup and comments! 👍 🚢 🎉 🖼️ 🙌

viviancan:
> LGTM! The only thing that crossed my mind is if the filestack styles should all be together in a separate stylesheet file. Thanks for the writeup and comments! 👍 🚢 🎉 🖼️ 🙌

Thx for the feedback! I moved the styles to a filestack specific file! :)

PR #3662, opened 2022-09-28
Add output comments to season reset rake task

shaunxp20:
I didn't make any functional changes here, I just adding some output comments to make it clearer what is happening if the rake task is successful or providing some insights if it didn't run.

Example outputs:

```
Resetting consent waivers
Resetting data use terms
Resetting onboarded flags
Resetting mentor trainings
Resetting judge trainings
Finished resetting
```

```
The reset needs to happen on the season start date (2022-10-01) or you can run this rake task with the 'force' option (e.g. `bundle exec rake reset_for_season[force])`
```


viviancan (approved):
👍 Nice improvements here! Very helpful for whoever has to run the rake task! 💯

Issue #3659, opened 2022-09-27
Potential unwanted behavior -  When logging into you old judge account you receive the welcome/start judging email. See attached screenshot


viviancan:

viviancan:
![IMG_1315.jpg](https://images.zenhubusercontent.com/606b0f4fd13d53dff3d14fa5/a7d9962d-0afc-4b2b-ac1b-be798fe278ad)

viviancan:
@dboyer tagging you since it may want to be tackled in the next iteration before the season opens!

dboyer:
The welcome judge email was added for the first time in the 2022 season. Need to add conditional to not send welcome email if judge registration is off AND judging is set to finished or off.

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/3659?planning-poker) with Zenhub @shaunxp20 @viviancan

PR #3666, opened 2022-09-30
Added a conditional to only send judge welcome email when judge registration is open

viviancan:
Refs #3659 

This change was added to prevent the judge welcome email from being sent unless judge registration is open. 

We will need to revisit this in the future because any previous judges the login to their account before judge registration is open will be registered to the current season and we believe since that job will be completed, they will not receive the welcome email at all.


shaunxp20 (approved):
I didn't try it locally, but it looks good/straight forward to me! 👍 :shipit: 🎉

Issue #3658, opened 2022-09-27
Integrate filestack with team photo upload  


viviancan:

viviancan:
We have decided to temporarily update the text for the "change image" button while filestack is integrated for team photo upload. [PR 3667](https://github.com/Iridescent-CM/technovation-app/pull/3667) changes the text.

viviancan:
@dboyer confirming updated button text is on QA

dboyer:
Confirming I also see the coming soon button text on QA

viviancan:
With this change:
1. Students can upload a team photo using filestack
2. Admin can also now update team photos from the admin portal! This is a new feature. Admin can only update team photos for current season teams in order to preserve historical data for future seasonality updates.

**Student Team Photo Upload**
https://images.zenhubusercontent.com/29868115/30c2f862-a5ce-4a82-9d2e-8eb68e0392d9/team_photo_upload.mp4

**Admin Team Photo Upload for current season teams**
![CleanShot 2022-10-13 at 16.11.51@2x.png](https://images.zenhubusercontent.com/606b0f4fd13d53dff3d14fa5/fc1712e9-b247-4521-ba65-91832e82e027)



**Admin Team Photo Upload for previous season teams**
![CleanShot 2022-10-13 at 16.28.19@2x.png](https://images.zenhubusercontent.com/606b0f4fd13d53dff3d14fa5/4f7c5a99-2a65-429f-9480-492e351d13e7)

viviancan:
Confirming that this is on QA and working as expected for students and admin.

@dboyer Should ChA have the ability to update team photos? If so, that will need to be in a splitoff ticket. Right now as a ChA the change image button is visible to a ChA (from the shared partial with an admin) but it is not working. I can either add a conditional where only admin can view/use that button or investigate the issue for ChA

dboyer:
Texted on QA and was able to upload a team image which then appeared on the team page and in the team search process. Could also edit the photo as an admin.

Will ask about ChAs editing photos at dev-priorities meeting.

viviancan:
Verifying team photo is on production

dboyer:
Also confirmed team photo upload is working on production

PR #3674, opened 2022-10-13
Integrate filestack for team photo upload

viviancan:
Refs #3658 

This change includes filestack integration for team photo upload. It would be great if you can test this one locally. I tried to login as different scenarios but there are so many places the team photo is integrated it would be helpful to have a second set of eyes 👀 ! 

With this change:
1. Students can upload a team photo using filestack
2. Admins can also now update team photos from the admin portal! This is a new feature. Admin can only update team photos for current season teams in order to preserve historical data for future seasonality updates. 

I added a beginning version of tests for this in `spec/features/student/submission_piece_spec.rb`. I spent some time researching, but it seems there is a known issue with rspec and filestack. I had difficulty getting the modal to actually open. I will continue to research this issue. 

Finally, the tests kept failing due to an undefined method `team_photo.url`. I think this was related to the old photo upload process/carrierwave/minimagick. I updated each of those references to `team_photo_url` which is an updated method in the team model that determines the correct team photo URL. This is similar to the approach with the account profile image.

Note: Team photos are stored on AWS S3 - `environment_bucket_here/uploads/team/team_photo/#{id}/#{team_photo}`

### DIFF
--- a/app/models/team.rb
+++ b/app/models/team.rb
@@ -412,7 +412,7 @@ class Team < ActiveRecord::Base
   private
 
   def default_team_photo_url
-    if Season.current.year >= 2023
+    if Season.current.year >= 2023 && current?
       ActionController::Base.helpers.asset_path("placeholders/default-team-avatar.png")
     else
       ActionController::Base.helpers.asset_path("placeholders/team-missing.jpg")
--- a/app/models/team.rb
+++ b/app/models/team.rb
@@ -155,8 +155,6 @@ class Team < ActiveRecord::Base
     team_photo
   end
 
-  mount_uploader :team_photo, TeamPhotoProcessor
-
   Division.names.keys.each do |division_name|
     scope division_name, -> {
       joins(:division).where(
@@ -261,7 +259,6 @@ class Team < ActiveRecord::Base
 
   validates :name, presence: true, team_name_uniqueness: true
   validates :division, presence: true
-  validates :team_photo, verify_cached_file: true
 
   delegate :name, to: :division, prefix: true
 
--- a/app/models/team.rb
+++ b/app/models/team.rb
@@ -274,6 +274,16 @@ class Team < ActiveRecord::Base
     has_students? && all_students_onboarded?
   end
 
+  def team_photo_url
+    if team_photo.blank?
+      default_team_photo_url
+    elsif team_photo.include?("filestackcontent")
+      team_photo
+    else
+      "https://s3.amazonaws.com/#{ENV.fetch("AWS_BUCKET_NAME")}/uploads/team/team_photo/#{id}/#{team_photo}"
+    end
+  end
+
   def photo_url
     team_photo_url
   end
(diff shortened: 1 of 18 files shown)


app/models/team.rb
```diff
@@ -401,4 +408,14 @@ def assigned?
   def assigned_judge_names
     assigned_judges.flat_map(&:full_name)
   end
+
+  private
+
+  def default_team_photo_url
+    if Season.current.year >= 2023 && current?
```
viviancan, line 415: This is from the work you (@shaunxp20 ) added in `app/uploaders/team_photo_processor.rb`. I added the 2nd condition because I realized that if a team from a previous season did not have a team photo and you viewed the team the image would be the new team placeholder. I think we wanted there to be a difference in the team photo to preserve the seasonality for the future!

app/views/admin/participants/edit.html.erb
```diff
@@ -44,7 +44,7 @@
     </p>
 
     <p>
-      <%= f.label :profle_image %>
+      <%= f.label :profile_image %>
```
viviancan, line 47: fixed typo 😅

app/views/filestack_uploads/_team_photo_upload.en.html.erb
```diff
@@ -0,0 +1,35 @@
+<%= javascript_tag do %>
```
viviancan, line 1: I decided to add a new directory for any filestack functionality partials. Especially since there will be differences for 
1. profile image
2. team photo
3. screenshots
4. potentially document upload

After working with the current upload system where there was a single dynamic modal used for all of the different scenarios, I felt this was a better approach for long-term support and maintenance.

spec/features/student/submission_piece_spec.rb
```diff
@@ -39,6 +39,13 @@
     end
   end
 
+  scenario "Set the team photo" do
```
viviancan, line 42: I started to build out the testing, however there seems to be a known issue when interacting with filestack through rspec. I will continue to research this issue.

app/views/filestack_uploads/_team_photo_upload.en.html.erb
shaunxp20: Awesome, great idea! 🙌

app/views/admin/participants/edit.html.erb
shaunxp20: 👍 😅

spec/features/student/submission_piece_spec.rb
shaunxp20: 👍

app/models/team.rb
shaunxp20: Cool, sounds good! Seeing this again, maybe if we're going to reuse this logic, it could be nice to have it wrapped in a method like `.using_new_file_upload_system?` (not sure where it live, and just a random thought/idea too). 😄

shaunxp20 (approved):
Changes look good to me! I tried it locally and it didn't work, but I think we determined it's something w/ my local setup, which I'll investigate later! Thanks for your help troubleshooting! Excited to see this coming together! 👍 🙌 🎉

Issue #3107, opened 2021-11-17
Onboarding flow improvements


dboyer:
Draft list of onboarding improvements to discuss:

- [x] After guardians complete parental consent and media consent forms, they receive a "Thank you for signing the media consent form" banner at the top of the page. Perhaps either change the text to include about what happens next (i.e. "You are all set. Your daughter can now join a Technovation team. You can now close this page.") or redirect them to a different landing or login page.

- [x] When the user clicks between the second (enter name, etc) and third (agree to terms) screens of registration, the page stays scrolled to the bottom. It would be helpful to have the third page load with the user viewing the top of the page so they don't have to scroll up to read the terms of use.

- [x] Replace circles at top of parental consent/media consent with Step 1 and Step 2. The circles make it look like the user should be able to click on them, which is not possible.

<img width="474" alt="Screen Shot 2021-11-23 at 11 23 49 AM" src="https://user-images.githubusercontent.com/516764/143063288-658e6e01-a9fe-4806-9993-4944eeb7939f.png">

shaunxp20:
> - [X] When the user clicks between the second (enter name, etc) and third (agree to terms) screens of registration, the page stays scrolled to the bottom. It would be helpful to have the third page load with the user viewing the top of the page so they don't have to scroll up to read the terms of use.

This one is regarding the registration flow and was fixed at some point.

> - [X] Replace circles at top of parental consent/media consent with Step 1 and Step 2. The circles make it look like the user should be able to click on them, which is not possible.

The circles have been replaced w/ a Tailwind component, hopefully it's better! 🤞

dboyer:
Bringing this up again as we work on registration/onboarding for the new season and QA revealed a few potential improvements for parental consent onboarding:

- [ ] Could the cursor be a hand and not an arrow on the buttons to submit the parental consent and media forms? (i.e. when hovering on the Submit button in this screenshot) That is the behavior elsewhere on the platform

<img width="357" alt="Screen Shot 2022-09-23 at 10 46 20 AM" src="https://user-images.githubusercontent.com/516764/191988119-f2572506-7359-42ba-8b29-d916b5497118.png">

- [ ] If a user forgets to fill in a field, the form reloads at the top of the screen so the user can't see the error easily. Could it load at the bottom of the screen so the needed information is visible? The error text should also be a red color or have some other visual indication to draw attention.

[...]

PR #3675, opened 2022-10-13
Consent Form Improvements

shaunxp20:
From #3107, here are the things being addressed in this PR:

- Make the cursor be a hand and not an arrow on the buttons to submit the parental consent and media forms

-  If a user forgets to fill in a field, the form reloads at the top of the screen so the user can't see the error easily. Could it load at the bottom of the screen so the needed information is visible? The error text should also be a red color or have some other visual indication to draw attention.

- After guardians complete parental consent and media consent forms, they receive a "Thank you for signing the media consent form" banner at the top of the page. Let's change the text to say:

> "Thank you for signing the forms. Your child can now join a Technovation Girls team. Please find more information on supporting her participation in Technovation on our Parent Resources page."

### DIFF
--- a/app/views/media_consents/edit.en.html.erb
+++ b/app/views/media_consents/edit.en.html.erb
@@ -46,9 +46,6 @@
 
   <div>
     <%= f.input :token, as: :hidden, input_html: { name: :token, value: params[:token] } %>
-    <%= f.button :submit,
-      t("views.media_consents.submit"),
-      class: "tw-green-btn"
-    %>
+    <%= f.submit t("views.media_consents.submit"), class: "tw-green-btn cursor-pointer" %>
   </div>
 <% end %>
--- a/app/views/media_consents/edit.en.html.erb
+++ b/app/views/media_consents/edit.en.html.erb
@@ -7,7 +7,10 @@
   }
 %>
 
-<%= simple_form_for @media_consent, html: {class: "consent-form"} do |f| %>
+<%= simple_form_for @media_consent,
+  html: {class: "consent-form"},
+  url:  media_consent_path(@media_consent, anchor: "media-consent-form") do |f| %>
+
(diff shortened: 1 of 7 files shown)


app/views/media_consents/edit.en.html.erb
```diff
@@ -7,7 +7,10 @@
   }
 %>
 
-<%= simple_form_for @media_consent, html: {class: "consent-form"} do |f| %>
+<%= simple_form_for @media_consent,
+  html: {class: "consent-form"},
+  url:  media_consent_path(@media_consent, anchor: "media-consent-form") do |f| %>
```
shaunxp20, line 12: I added `anchor` which corresponds to the new `id` below, this is basically so that if there are validation errors (e.g. someone submitted the consent form and didn't provider their name), this will re-render the page and take them to the input fields vs. putting them back at the top of the page and to them scroll all the way down to the fields. I made this change for the parental consent form too.

More about and where I discovered this technique:
﻿https://stackoverflow.com/a/14430409

app/views/parental_consents/edit.en.html.erb
```diff
@@ -24,7 +27,7 @@
 
   <div class="w-full border-gray-400 my-8"></div>
 
-  <div class="my-4">
+  <div id="parental-conssent-form" class="my-4">
```
viviancan, line 30: `Conssent` has an extra `s` here!

viviancan (approved):
💯 Did not try locally, but looks good to me Shaun! I just made one comment about a small typo. Thanks for sharing the link with the anchor technique. 🚀

app/views/parental_consents/edit.en.html.erb
shaunxp20: Great catch, thank you! 👍 😊 😅

Issue #3670, opened 2022-10-11
Only show link to judge training on judge dashboard if judge registration or judging are toggled on


dboyer:
If judges from a previous season login, they are directed to their judge dashboard where they see a link to complete judge training. The training materials are not yet updated for the 2023 season.

Ideally, we should have the training link appear as an option in the left sidebar only if judge registration is turned on or if judging is available. If judge registration are off and judging is off, no one needs to be completing training.

If that is not an option, for now we should just comment out that link in the sidebar so judges can't complete registration early.

DECISION: will use conditional logic to add a tooltip to judge training saying training is not available (if judge registration is not on)

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/3670?planning-poker) with Zenhub @shaunxp20 @viviancan

viviancan:
Confirming this is on QA

dboyer:
Confirmed on QA that if logged in as a judge (not a mentor and judge) I go to my judge dashboard where I can sign the consent waiver but cannot complete training.

If I log in as a mentor (who is also a judge), I go to my mentor dashboard where I can complete mentor training.

viviancan:
Verifying this is on Production

dboyer:
Confirmed the training link is not accessible on production

PR #3676, opened 2022-10-17
3670 only show judge training under certain conditions

viviancan:
Refs #3670 

This change includes new css additions for use of a custom tooltip.
There are 2 main new classes `has-tooltip` and `tooltip-text`. `has-tooltip` is used as a wrapper and `tooltip-text` contains the actual text of the tooltip.

CleanShot 2022-10-17 at 17 37 46


app/javascript/packs/application_rebrand.js
```diff
@@ -1,15 +1,16 @@
-import "../stylesheets/tailwind.css"
+import "../stylesheets/tailwind.css";
```
viviancan, line 1: My linter automatically added the semicolons! I tried removing them, but every time I save they are automatically added before I am able to commit. Not sure if this is part of our standard. I can look at disabling this setting if this is an issue!

app/views/judge/dashboards/rebrand/_side_nav_content.html.erb
```diff
@@ -26,11 +26,18 @@
   <% else %>
     <%= web_icon("circle-o") %>
   <% end  %>
-  <%= link_to "Judge Training",
-              "https://technovationchallenge.org/judge-training/",
-              target: "_blank",
-              id: "judge-training-tab"
-  %>
+
+  <% if SeasonToggles.judging_enabled? || SeasonToggles.judge_registration_open? %>
```
viviancan, line 30: I researched a few libraries, but they all seemed heavier than what we needed. So I decided to go with a basic CSS custom route. 

I also tried a variety of approaches with the `link_to`method including the `link_to_if` using a block to dynamically display the tooltip span. I had a working version, but it was so dense with all of the conditional logic and extra spans, that it just didn't seem very clear in the end.

app/javascript/packs/application_rebrand.js
shaunxp20: Hmm, not sure off the top of my head, but I'm guessing it's a default from ESLint or Prettier. But anywho, yeah, we should probably get specific on our rules and what we want, and make them explicit! 😃

shaunxp20 (approved):
Didn't try it locally, but LGTM, thanks Vivian! 👍 🙌 :shipit:

Issue #3685, opened 2022-10-21
Restrict file types that can be selected for image uploads


dboyer:
On both profile and team photos, users can select file types (zip, pdf, heic, etc) that cannot be displayed on the platform.

We should restrict file type uploads to png, jpg, and other image files.

viviancan:
This issue has been solved. The `accept` option was added to the `picker options`. See details from [documentation here](https://filestack.github.io/filestack-js/interfaces/pickeroptions.html#accept)

dboyer:
Confirmed on QA that I can only upload image files. The window greyed out any files in my downloads folder that weren't images so selecting them wasn't an option.

dboyer:
Confirmed on production that only certain image types can be selected for upload

PR #3686, opened 2022-10-21
Added accepted filetypes for filestack profile image upload - file types include jpeg, jpg, png

viviancan:
Refs #3685 

This change adds accepted file types for the filestack profile image upload process


viviancan:
@shaunxp20 Merging this to QA without review so it is able to go out in the next deployment next week!

PR #3697, opened 2022-10-31
Update demo video to technical video (for occurances in Rails)

shaunxp20:
This will update all instances of "demo video" to "technical video" on the Rails side, including the specs.

### DIFF
--- a/config/locales/models/en.yml
+++ b/config/locales/models/en.yml
@@ -1,0 +1,3 @@
+en:
+  submissions:
+    demo_video: "technical video"
--- a/app/models/team_submission.rb
+++ b/app/models/team_submission.rb
@@ -247,25 +247,31 @@ class TeamSubmission < ActiveRecord::Base
     presence: true,
     if: ->(team_submission) { team_submission.game? }
 
-  validates :pitch_video_link, 
-    format: { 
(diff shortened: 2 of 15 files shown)


config/locales/models/en.yml
```diff
@@ -0,0 +1,3 @@
+en:
+  submissions:
+    demo_video: "technical video"
```
shaunxp20, line 3: Regarding the organization of our translations, right now everything is in one directory (there is very little organization/structure). This directory structure is from the Rails Guides on i18n: https://guides.rubyonrails.org/i18n.html#organization-of-locale-files

The config I added above ☝️ is also from the guides (in the same section).

I originally had this in a `views` directory, but I ended up putting it in a `models` directory because I thought the translation should happen closer to the "source" of what is being translated. I would be happy to move it back to `views`, I'm not sure which is more ideal. 🤔 😅

viviancan: Shaun, I like the reasoning you shared! I am +1 for keeping it in models!

viviancan (approved):
Did not try locally, but looks pretty straight forward to me! Great job tackling this one, Shaun 💯

PR #3698, opened 2022-11-01
Export i18n translations to JSON

shaunxp20:
This adds and sets up the `i18n-js` gem that will export i18n translations to json.

### DIFF
--- a/app/javascript/config/locales/en.json
+++ b/app/javascript/config/locales/en.json
@@ -1,0 +1,7 @@
+{
+  "en": {
+    "submissions": {
+      "demo_video": "technical video"
+    }
+  }
+}
--- a/config/i18n.yml
+++ b/config/i18n.yml
@@ -1,0 +1,4 @@
+translations:
+  - file: app/javascript/config/locales/:locale.json
+    patterns:
+      - "*.submissions.*"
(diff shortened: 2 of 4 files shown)


config/i18n.yml
```diff
@@ -0,0 +1,4 @@
+translations:
+  - file: app/javascript/config/locales/:locale.json
```
shaunxp20, line 2: This is where the JSON file will get exported to, it roughly mirrors the Rails directory structure `app/config/locales/en.yml`. Right now only one file is getting exported, but in the future we have the potential to export to multiple files and/or directories. There is a command that needs to be run to create this exported file, I still need to document that part.

config/i18n.yml
```diff
@@ -0,0 +1,4 @@
...
+    patterns:
+      - "*.submissions.*"
```
shaunxp20, line 4: This is *what* we're exporting, basically just "submission" translations for now. 👍

app/javascript/config/locales/en.json
```diff
@@ -0,0 +1,7 @@
+{
```
shaunxp20, line 1: This is the JSON file the is created/generated (based on the config settings below).

viviancan: 👍

config/i18n.yml
viviancan: 👍

viviancan:
Hi Shaun, are there steps I need to take locally to generate the file?

viviancan (approved):
LGTM! Did not try locally but can revisit once steps are documented! Not a blocker or anything 👍

Issue #3692, opened 2022-10-28
Integrate filestack into photo submission upload for students


viviancan:

viviancan:
Verifying this is on QA. I have a follow-up PR ([PR #3702](https://github.com/Iridescent-CM/technovation-app/pull/3702)) that will fix the photo grid styling in some of the other places where screenshots are viewed.

dboyer:
Testing on QA and it's looking good. I love the process of selecting multiple images and then having an option to edit each - so smooth. A couple things:

- I see that only uploading from the computer is possible and not the web cam or link to a URL options that are available for profile pics. I think that makes sense but wanted to confirm that was an intentional choice.

- When uploading three or more images, the image extends beyond the container. I don't think this is a huge deal but a heads up in case it is specifically not supposed to do that.
<img width="720" alt="Screen Shot 2022-11-09 at 11 11 46 AM" src="https://user-images.githubusercontent.com/516764/200881880-79cab3e2-3d13-4928-8aa8-feceb5ebcd57.png">

dboyer:
Three people on PT and I ran through photo submissions again as both a student and mentor and did not find any issues with the photo upload process. We're good to deploy this!

viviancan:
Re: images extending the container. I could not reproduce this locally! I can try another solution using flexbox. I do not think this is a blocker though! Another ticket will be created so I can investigate further.

PR #3700, opened 2022-11-02
Part1 - Updated student/mentor screenshot upload process to use filestack

viviancan:
Refs #3692

CleanShot 2022-11-02 at 14 32 41

The "meat" of this PR is in `app/javascript/components/ScreenshotUploader.vue`. This component was updated to use filestack to upload and process screenshot images. 

Basic flow for filestack screenshot image upload
1. User adds image using filestack picker (widget/modal that opens to add photo)
2. Filestack begins uploading the photo through filestack and stores to the appropriate s3 bucket (`AWS_ENV_BUCKET/uploads/screenshot/filestack/team_submission_id/image`) 
3. Once process is complete, filestack returns a filestack reference URL that points to the image in the s3 bucket. This URL reference is some version of `https://cdn.filestackcontent.com/RANDOM_STRING_HERE`
4. A filestack specific callback function is called (`onFileUploadFinished`) and a form is created with appropriate form data
5.  A post request is made to save the screenshot records which uses the filestack reference URL for the image column in the DB

Note - I had to update the s3 bucket path where screenshots are being stored. They were originally stored in the following path: `AWS_BUCKET_NAME/uploads/screenshot/image/screenshot_id/image`. Now they are stored `AWS_ENV_BUCKET/uploads/screenshot/filestack/team_submission_id/image`.

This update was made due to the change in the image upload process. Previously, we handled the connection to the s3 bucket. When the process to store the image to the s3 bucket occurred, we already had the screenshot id, so it was possible to use that id as part of the path. Now, filestack handles the connection and upload process to the s3 bucket. This process occurs before the image is stored in our DB, so we do not yet have the screenshot id. I updated the path to use the team submission id instead. I also added a note in the code.

### DIFF
--- a/app/assets/stylesheets/_screenshots.scss
+++ b/app/assets/stylesheets/_screenshots.scss
@@ -1,17 +1,25 @@
 .submission-pieces__screenshots {
-  display: flex;
-  flex-flow: row nowrap;
+  margin-top: 30px;
+  margin-bottom: 50px;
   max-width: 100%;
-  margin: 1rem 0 1rem;
+  display: grid;
+  grid-template-columns: repeat(3, 1fr);
+  grid-column-gap: 20px;
+  grid-row-gap: 50px;
 }
 
-.submission-pieces__screenshot {
-  margin-right: 1rem;
+.screenshot-wrapper {
+  width: 200px;
+  height: 200px;
+  margin: auto;
 
   img {
-    max-width: 100%;
+      height: 100%;
+      width: 100%;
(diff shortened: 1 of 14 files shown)


shaunxp20 (approved):
Awesome, LGTM! Thanks for tackling this one, exciting to see the new image uploaded come together!! 🙌  Only thing I noticed were some old BEM? style class names, but we can probably address those when we get to rebranding submissions! 👍 :shipit:

PR #3702, opened 2022-11-03
Extracted screenshot grid into a single partial for reuse

viviancan:
Refs #3692 

General style cleanup for filestack + screenshots


shaunxp20 (approved):
Awesome, nice work! I like it! 🥳 🙌 🎉

PR #3707, opened 2022-11-07
Add rake task to reset judge trainings

shaunxp20:
This will add a rake task that will reset judge trainings for judges in the current season (2023) who completed the judge training for the 2022 season.

I tested this locally w/ production data and it reset 23 records (I removed production data from my machine already). I also wrote a dataclip and there were 23 records being returned, so it's lining up.

### DIFF
--- a/lib/tasks/reset_judge_training.rake
+++ b/lib/tasks/reset_judge_training.rake
@@ -1,0 +1,14 @@
+desc "Reset judge trainings for judges in the current season"
+task reset_judge_training: :environment do
+  accounts = Account.current.joins(:judge_profile).where.not(judge_profile: {completed_training_at: nil})
+  judges_updated_count = 0
+
+  accounts.each do |account|
+    puts "Resetting judge training for account: #{account.id} - #{account.first_name} #{account.last_name}"
+
+    account.judge_profile.update(completed_training_at: nil)
+    judges_updated_count += 1
+  end
+
+  puts "Reset #{judges_updated_count} judges"
+end


lib/tasks/reset_judge_training.rake
```diff
@@ -0,0 +1,14 @@
+desc "Reset judge training for 2023 judges who completed the 2022 judge training"
+task reset_judge_training: :environment do
+  accounts = Account.current.joins(:judge_profile).where.not(judge_profile: {completed_training_at: nil})
```
shaunxp20, line 3: Just FYI (because it wasn't super clear to me), `.current` ☝️ means the current season (2023).

lib/tasks/reset_judge_training.rake
```diff
@@ -0,0 +1,14 @@
+desc "Reset judge training for 2023 judges who completed the 2022 judge training"
```
viviancan, line 1: Should the description be more general in terms of the year since the actual rake task is based on the `current` year which will change depending on the season? Just a thought 🤔

lib/tasks/reset_judge_training.rake
```diff
@@ -0,0 +1,14 @@
...
+  accounts = Account.current.joins(:judge_profile).where.not(judge_profile: {completed_training_at: nil})
+  judges_updated_count = 0
+
+  accounts.each do |account|
+    puts "Resetting judge training for account: #{account.id} - #{account.first_name} #{account.last_name}"
+
+    account.judge_profile.update(completed_training_at: nil)
+    judges_updated_count += 1
+  end
+
```
viviancan, line 12: Looks good! I like the extra output information. Very helpful 💯

viviancan (approved):
Tried it locally and worked as expected! Thanks, Shaun 🚀

lib/tasks/reset_judge_training.rake
shaunxp20: Good catch, good eye! My original thought was to remove this rake task after it has been run (we used to do that in the past w/ one-off rake tasks). Anywho, I updated the description to make it more general (nice suggestion!), since I'm not sure what the plan is for this one after it's been run. 👍 😅

Issue #3721, opened 2022-11-16
Fix ChA profile image size issues on mentor dashboard


viviancan:
- Profile image in viewing "MyChA" on mentor profile is very large. Needs to be resized for the dropdown
- If a ChA does has not uploaded a profile image, their account defaults to the placeholder hedgehog profile image. On the mentor dashboard this image link is broken in the "meet your chA" heading dropdown. See photo below. This image should not be broken and should display the hedgehog

viviancan:
![CleanShot 2022-11-16 at 12.36.47@2x.png](https://images.zenhubusercontent.com/606b0f4fd13d53dff3d14fa5/8a0f66cb-8923-4497-98ae-25c223ba0caf)

dboyer:
Confirmed that this is now working on QA. I can see the hedgehog image if the ChA hasn't uploaded a profile photo and the image size is correct as a thumbnail and in the drop-down screen.

PR #3723, opened 2022-11-17
Fixed ChA profile image issues on mentor dashboard

viviancan:
Refs #3721 

These changes were needed due to 2 issues on the mentor dashboard regarding ChA profile images.
1. Added style `og-style-filestack-img` to the ChA profile image in the "Meet your ChA" dropdown on the mentor dashboard so all profile image sizes are the same
2. While fixing the first issue, I realized if the ChA did not upload a profile image, they default to the hedgehog placeholder image. This image link was broken on the mentor dashboard. I added a method to determine the correct src since the placeholder image is in our assets directory


shaunxp20 (approved):
LGTM! 👍 🚀 🚢 :shipit:

PR #3727, opened 2022-11-21
Add beginners division to RPEs

shaunxp20:
This will allow chapter ambassadors to create regional pitch events for the beginners division.


viviancan (approved):
Tried locally and works as expected! As a ChA, I was able to add a beginner event and add a team to that event. I could view the event details as an admin. Way to go, Shaun 💯 Huge win!

PR #3742, opened 2022-12-02
Update submission deadline text

shaunxp20:
Before this PR we had two methods for displaying the submission deadline - `deadline` and `submission_deadline` - that would display the submission deadline in slightly different formats.

With this PR:
- `deadline` was removed
- `submission_deadline` was updated and will now be used to display the submission deadline everywhere
- New timezone formats (WAT, CEST, IST) were added for the submission deadline
- A new ENV (`DATES_SUBMISSION_DEADLINE_HOUR_IN_24_HOUR_FORMAT`) was added for the time of the submission deadline


viviancan:
@shaunxp20 Is there anything to test specifically?

shaunxp20:
> @shaunxp20 Is there anything to test specifically?

I think this one would be fine to test/check on QA - basically just making sure the submission deadline is displaying everywhere correctly.

PR #3744, opened 2022-12-05
WIP Rebrand new password page

shaunxp20:
This will rebrand the page that prompts someone for their new password (after clicking a link in the password reset email).


viviancan (approved):
LGTM! 💯 Tried it locally and I saw the updated reset your password page when I clicked the link!

PR #3746, opened 2022-12-06
Fix RPE polymorphic path on mentor dashboard

shaunxp20:
This path was updated as part of the Rails 6 upgrade, but it got updated incorrectly, the path used to take a `submission` and that was incorrectly removed.

See commit:
02a55fa3aadb0ae2baa74120b04c7e2db6cda9b5

`_pitch_presentation.en.html.erb`:
```
<%= form_with model: [current_scope, submission], multipart: true,
```


viviancan (approved):
LGTM! 💯

Issue #3733, opened 2022-11-23
Integrate filestack screenshot upload system for admin portal


viviancan:

dboyer:
Confirmed on QA that I could upload images as an admin!!!! I logged in as the student associated with the team and was able to see and edit the images.

dboyer:
Program team tested this and confirmed it is working as expected. Moving to ready to deploy.

Issue #3703, opened 2022-11-03
Add/check screenshot specs for filestack 


viviancan:

viviancan:
test

PR #3753, opened 2022-12-13
3733 screenshot filestack in admin portal

viviancan:
Refs: #3733 

This change integrates filestack in the admin portal. Currently, as an admin you can only upload 1 screenshot at a time. I also added the groundwork for the filestack spec. Similar to the other filestack specs, I am running into issues actually trying to open the filestack picker with Capybara. After some research, it could be a driver issue since filestack requires JS.  I will be tackling this issue in #3703.

Next steps include the following: 
- Adding delete functionality
- Ability to upload more than 1 screenshot at a time

CleanShot 2022-12-13 at 09 41 01

### DIFF
--- a/app/views/admin/screenshots/new.html.erb
+++ b/app/views/admin/screenshots/new.html.erb
@@ -28,7 +28,7 @@
                                     intelligent: true,
                                   },
                                   fromSources: ["local_file_system"],
-                                  onFileUploadFinished: 'onFileUploadFinished',
+                                  onFileUploadFinished: "onFileUploadFinished",
                                   storeTo: {
                                     location: "s3",
                                     container: ENV.fetch("AWS_BUCKET_NAME"),
--- a/app/views/admin/screenshots/new.html.erb
+++ b/app/views/admin/screenshots/new.html.erb
@@ -1,0 +1,45 @@
+<%= javascript_tag do %>
+  function onFileUploadFinished(data) {
+    const screenshot = document.getElementById("screenshot");
+    screenshot.src = filestack_client.transform(data.handle);
+  }
+<% end %>
+
+<div class="grid">
+  <div class="grid__col-10">
+    <div class="panel">
+
+      <h1>Add a screenshot</h1>
+      <% @team_submission.while_no_screenshots_remaining do %>
+        <p>Please delete a screenshot(s) in order to upload additional images</p>
+      <% end %>
+
+      <% @team_submission.while_screenshots_remaining do %>
+        <img src="" alt="" id="screenshot" class="og-style-filestack-img">
+
+        <%= form_with model:[@team_submission, @screenshot], url: admin_team_submission_screenshots_path(@team_submission), local: true do |f| %>
+          <%= f.label :image, "Screenshot" %>
+          <%= f.filestack_field :image,
+                                "Select image",
+                                id: 'admin-screenshot-picker',
+                                pickerOptions: {
+                                  accept: ["image/jpeg", "image/jpg", "image/png"],
+                                  uploadConfig: {
+                                    intelligent: true,
+                                  },
+                                  fromSources: ["local_file_system"],
+                                  onFileUploadFinished: 'onFileUploadFinished',
+                                  storeTo: {
+                                    location: "s3",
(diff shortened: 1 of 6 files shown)


app/views/admin/screenshots/new.html.erb
```diff
@@ -0,0 +1,45 @@
...
+  }
+<% end %>
+
+<div class="grid">
+  <div class="grid__col-10">
+    <div class="panel">
+
+      <h1>Add a screenshot</h1>
+      <% @team_submission.while_no_screenshots_remaining do %>
+        <p>Please delete a screenshot(s) in order to upload additional images</p>
```
shaunxp20, line 14: A very minor thing about the verbiage/code, sometimes I seem them as screenshots and others its images. 😅

app/views/admin/screenshots/new.html.erb
```diff
@@ -0,0 +1,45 @@
...
+      <% end %>
+
+      <% @team_submission.while_screenshots_remaining do %>
+        <img src="" alt="" id="screenshot" class="og-style-filestack-img">
```
shaunxp20, line 18: I see! Interesting! IIUC, after a screenshot is uploaded, it gets "populated" here. 🤔

app/views/admin/screenshots/new.html.erb
```diff
@@ -0,0 +1,45 @@
...
+                                    intelligent: true,
+                                  },
+                                  fromSources: ["local_file_system"],
+                                  onFileUploadFinished: 'onFileUploadFinished',
+                                  storeTo: {
+                                    location: "s3",
+                                    container: ENV.fetch("AWS_BUCKET_NAME"),
+                                    path: "uploads/screenshot/filestack/#{@team_submission.id}/",
+                                    region: "us-east-1"
+                                  }
```
shaunxp20, line 37: Minor thing, I think the indentation is slightly off (from `fromSources` and below), also there are mixed single/double quotes. 😅 

Also, random thought, it might be nice to extract some of this out into default settings (like in `storeTo` and `pickerOptions`), seems like we'd want the same image types and S3 settings pretty much everywhere, and they could always be overridden if necessary too. 🤔 😄

app/views/admin/team_submissions/edit.html.erb
```diff
@@ -78,6 +78,27 @@
...
+
+        <div>
+          <h5>Screenshots</h5>
+
+          <% if @team_submission.screenshots.any? %>
+            Total Uploaded Images: <%= @team_submission.screenshots.count %>
+
+            <div class="submission-pieces__screenshots">
+              <% @team_submission.screenshots.each do |screenshot| %>
+                <% if screenshot.image_url.present? %>
```
shaunxp20, line 90: Interesting! I wonder what causes `image_url` not to be present?! A random thought, I wonder if `screenshots` should only return screenshots w/ an `image_url` to protect against this (here and everywhere else)? 🤔 😄

config/routes.rb
```diff
@@ -270,6 +270,7 @@
     resources :teams, except: :destroy
     resources :team_submissions, except: :destroy do
       resource :judge_assignments, only: :create
+      resources :screenshots
```
shaunxp20, line 273: I think this can be limited to just `:new` and `:create` (by default it will define all of the CRUD actions). 😃

shaunxp20 (approved):
Looking good to me! Left a few comments, no real blockers, just some suggestions. Excited to see this coming together, thanks for working on this one @viviancan! 👏 🎉 🥳 🖼️ 🙌

app/views/admin/screenshots/new.html.erb
viviancan: Yes, this happens in `onFileUploadFinished`

app/views/admin/team_submissions/edit.html.erb
viviancan: Hmmm I'm not sure if there is a case of the `image_url` wouldn't be present... but I think after all of these random edge cases I was feeling slightly paranoid 😆 But I do like the idea of only returning screenshots w/ image_url

config/routes.rb
viviancan: 👍

Issue #3739, opened 2022-12-01
Update the layout for students when selecting an RPE 


viviancan:
When students view all of the available RPEs, the layout is a little strange since the "event details" partial is reused. See attached screenshot
![CleanShot 2022-12-01 at 16.05.11@2x.png](https://images.zenhubusercontent.com/606b0f4fd13d53dff3d14fa5/d7170c45-1050-4916-ae0c-858bbe1107cd)t

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/3739?planning-poker) with Zenhub @shaunxp20 @viviancan

dboyer:
Confirming that I see the updated layout for the events on QA

PR #3758, opened 2022-12-16
3739 update student layout for selecting rpe

viviancan:
Refs #3739 

This change updates the layout for student RPE event selection. I mirrored the mentor view for selecting events. I created a partial called `avaialble_event` that is used by the view. The information is very similar to the `event` partial that displays when students want to view the full event details of their selected RPE. There is definitely room for improvement in the flow, but I thought this was a good starting point.

I also found 1 broken link on the mentor dashboard for selecting an RPE, which has been included in this PR. 

CleanShot 2022-12-15 at 21 15 54@2x

### DIFF
--- a/app/views/regional_pitch_events/_event.en.html.erb
+++ b/app/views/regional_pitch_events/_event.en.html.erb
@@ -80,11 +80,9 @@
                   not event.at_team_capacity? %>
         <p>
           <%= link_to "Select this event",
-            [
-              current_scope,
-              :regional_pitch_event_selection,
-              { event_id: event.id, team_id: current_team.id },
-            ],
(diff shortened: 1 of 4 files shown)


viviancan:
@shaunxp20 i am also going to make a 2nd version of this that only uses the existing `event` partial similar to how it was before! The more I thought about this change the `available_event` partial is pretty much the same as the `event` partial, so it seems repetitive. I'll leave this up for now and then a final decision can be made!

viviancan:
Closing this. We went with the other approach in [PR3759](https://github.com/Iridescent-CM/technovation-app/pull/3759)

PR #3759, opened 2022-12-16
3739 v2 update layout for students selecting rpe

viviancan:
Refs #3739 

This change updates the layout for student RPE event selection. I mirrored the mentor view for selecting events. This is version 2 for this ticket. I prefer this approach because it only uses the existing `event` partial with a bit of conditional styling. The 1st version of this ticket duplicated the `event` partial with an updated layout, but the content was pretty much the same. 

I also found 1 broken link on the mentor dashboard for selecting an RPE, which has been included in this PR. 

**Horizontal layout for available events**
CleanShot 2022-12-16 at 12 36 34@2x

**Event details**
CleanShot 2022-12-16 at 12 39 38@2x

### DIFF
--- a/app/views/regional_pitch_events/index.en.html.erb
+++ b/app/views/regional_pitch_events/index.en.html.erb
@@ -19,5 +19,6 @@
   </div>
 <% else %>
   <%= render partial: 'regional_pitch_events/rebrand/event',
+             locals: { multiple_events: true },
              collection: @regional_events %>
 <% end %>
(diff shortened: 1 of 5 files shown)


app/views/regional_pitch_events/index.en.html.erb
```diff
@@ -19,5 +19,6 @@
   </div>
 <% else %>
   <%= render partial: 'regional_pitch_events/rebrand/event',
+             locals: { multiple_events: true },
```
viviancan, line 22: I added this variable for conditional styling in the event partial. If there are `mulitple_events` then the border style is added around the event.

shaunxp20 (approved):
Awesome, looks good! I like what you've done with the layout! 🙌 🎉 🥳

Issue #3586, opened 2022-08-15
Update new registration system so students can send invites even if registration is closed


dboyer:
In the previous registration system, admins and ChAs could send invites even if registration was closed. That feature is not available in the new registration system.

We should review the code for this feature from the old registration system and determine whether any aspects of it can be used in the new system. If not, we'll need to figure out another option.

Split off of #3497

viviancan:
We will need to review what is there and maybe some of the existing code can be refactored or referenced. It may be a heavier ticket, though, since this feature was not considered when we planned/built out the new registration functionality.

viviancan:
@dboyer What is the expected time frame for this functionality?

dboyer:
Agreed, and I have a note to talk about this at the dev team priorities meeting today This may end up being just a research task for the next iteration. I mostly wanted to make sure we checked out the old code for anything that could be reused before we fully removed it, which is another lingering task from #3053.

Invites during closed registration isn't needed until March 2023 when we have the small window where student registration is closed but submissions are open. Admins occasionally need to send an invite then when a user isn't registered in the system but has been working with a team.

dboyer:
Revising this to clarify:

Students on a team should be able to continue to send an invite to another person to join that team in the time period between registration closing and projects being submitted. If a person receives an invite to register in that time period, they should be able to do so even if the student or parent registration toggles are off.

shaun-technovation:
> Revising this to clarify:
> 
> Students on a team should be able to continue to send an invite to another person to join that team in the time period between registration closing and projects being submitted. If a person receives an invite to register in that time period, they should be able to do so even if the student or parent registration toggles are off.

Thinking about this again, I think this will be similar to the admin registration invites. We'll need to attach an invite code of some kind to the signup URL, then we'll need to check the conditions of the invite (a student already on a team, in the time period, etc), and then enable/disable registration based on that. I'll keep noodling on this and create some tickets for the work I think we need to do. 👍

dboyer:
Note: students should be able to invite both students and mentors to join their team.

shaun-technovation:
I broke down this issue into the following issues: #4283, #4245 and #4284.

dboyer:
Closing as the work to implement this feature will happen in the three issues mentioned above.

Issue #3497, opened 2022-06-17
Update invitation system for use with new registration work flow


dboyer:
There are a few tasks here that may need to be split into multiple issues:

- [ ] Remove old invitation system
- [ ] Update new registration system to allow admins to invite users to register even if registration is closed in general (this was a previous feature)
- [ ] Pre-fill email address when users are clicking a link to accept an invitation

shaun-technovation:
I added #4243 and #4244 for the major things to remove, and I created #4246 to capture the lingering things that we'll run into that will also need to be addressed.

PR #3775, opened 2023-01-06
3773 pt2 remove old judge signup controller

viviancan:
Refs #3773 and #3497 

Removes the old judge registration with invite token. We will discuss further because of #3586 Update new registration system so admins can send invites even if registration is closed. Not sure if this functionality that I removed would be useful for that or if we will have to start from scratch 🤔


shaunxp20 (approved):
👍 Feels good to finally be cleaning up some of this old stuff! 🧹 🙌 🥳

viviancan:
Closing this since #4249 covers this!

PR #3780, opened 2023-01-12
"Additional Questions" improvements

shaunxp20:
This PR has a few improvements for the "Additional Questions":
- Add missing i18n translations for the older questions
- Use the new `humanize_boolean` method instead of `team_submission.question_response`
- Remove `team_submission.question_response` since it's not being used any longer


viviancan (approved):
LGTM!

Issue #3695, opened 2022-10-28
Fix broken View Details link on judge RPE dashboard


dboyer:
"View Details" button on judge dashboard currently directs to a broken consent waiver page and instead should go to the new consent page (if judge needs to give consent) or just to the details of the submissions.

Note: the invite system for RPEs worked as expected. As a ChA, I sent an invite to a judge and as that judge, I received the email and was able to log-in to see this page.

dboyer:
The screen recording below if for my deb+mentor0310@technovation.org account.

https://drive.google.com/file/d/17lAkCBpBK5msNBj2Uep2ea8YvG3FC8nq/view?usp=drivesdk

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/3695?planning-poker) with Zenhub @shaunxp20 @viviancan

dboyer:
One update: We may remove this button if it just links to a page with a few more details about the event and instead put all the details on the first page on the judge dashboard. So fixing this button is not a huge priority. However, I would like to know where it is supposed to go in case that is some other page we weren't expecting and not just the event details.

viviancan:
![CleanShot 2023-01-13 at 08.03.45@2x.png](https://images.zenhubusercontent.com/606b0f4fd13d53dff3d14fa5/0ff988eb-315f-4652-9547-1727b691b276)
![CleanShot 2023-01-13 at 08.04.03@2x.png](https://images.zenhubusercontent.com/606b0f4fd13d53dff3d14fa5/8ad5e015-fd86-486c-9c2f-0f2662a9eff5)

After researching this issue, I discovered that this link did lead to the event details. This was changed in 2021 by the contractors as part of [this commit](https://github.com/Iridescent-CM/technovation-app/commit/70ba70bbee6b5dbd0e45e56d6de3a8890c1206d9). The path was changed from `event_path` to `consent_waiver_path` which is why the consent waiver details were displayed. The reason it was a random consent waiver is due to the event id parameter, which in this case was event id 143 😱 I have made the appropriate fix for now. I will work on removing it as part of #3766. I would still like to get this on QA for additional feedback in the case where a judge may be part of many RPEs (not sure if that is possible).

dboyer:
[...]

PR #3784, opened 2023-01-13
Fixex broken view details rpe link on judge dashboard

viviancan:
Refs #3695 - Additional details on the issue

This change was needed to fix the broken "view details" button for RPEs on the judge dashboard. The path was incorrect and was linked to a random consent waiver view. We may still remove the button as part of the new RPE flow.

### DIFF
--- a/app/views/judge/dashboards/onboarded/rebrand/_events.en.html.erb
+++ b/app/views/judge/dashboards/onboarded/rebrand/_events.en.html.erb
@@ -89,7 +89,7 @@
         </p>
 
         <div class="step-actions margin--b-xlarge">
-          <%= link_to 'View details',
+          <%= link_to "View details",
                       send("judge_regional_pitch_event_path", event),
                       class: "tw-green-btn" %>
         </div>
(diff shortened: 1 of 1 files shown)


app/views/judge/dashboards/onboarded/rebrand/_events.en.html.erb
```diff
@@ -89,9 +89,9 @@
         </p>
 
         <div class="step-actions margin--b-xlarge">
-          <%= link_to 'View details',
-            send("#{current_scope}_consent_waiver_path", event),
-            class: "button" %>
+          <%= link_to "View details",
+                      send("judge_regional_pitch_event_path", event),
```
shaunxp20, line 93: A very minor thing, I don't think we need to use `send` if the route is hard-coded for judges now, also nice update on that too! 😃

shaunxp20 (approved):
Found something very minor, but not a blocker, overall looks good and pretty straight forward to me! 👍 🎉 🙌 🎉

Issue #3766, opened 2023-01-04
Implement event partial for use in judge dashboard


dboyer:
Reuse the event partial from the student view on the judge dashboard. Exact location to be determined.

<img width="901" alt="Screen Shot 2023-01-04 at 1 42 46 PM" src="https://user-images.githubusercontent.com/516764/210626690-53c81948-b389-4ec3-b085-7f6a4904eff6.png">

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/3766?planning-poker) with Zenhub @shaunxp20 @viviancan

dboyer:
Confirming I see the event partial on QA.

PR #3786, opened 2023-01-13
Added RPE event partial to judge dashboard

viviancan:
Refs #3766 

Basic refactoring of judge dashboard view to include event partial. This change also removes the "view details" button since all the details are now listed on the dashboard 
CleanShot 2023-01-13 at 14 54 41@2x


shaunxp20 (approved):
I didn't try it locally, but the changes look good to me! 👍 🎉 🙌 🥳

PR #3787, opened 2023-01-13
Update admin additional questions

shaunxp20:
Quite a few updates here for the additional questions...

- Displaying the correct additional questions based on the season for admins and ChAs
- Allowing admins to edit all additional questions (from every season)
- Restricting students to only be able to edit additional questions for the current season
- Fixing a bug with using Ruby's predicate methods which only return `true` or `false`, but we need `true`, `false`, and `nil` to display Yes/No/-
- Fixing a bug with the climate change question - I accidentally updated this to use the game value


viviancan (approved):
LGTM! Nice clean up! Excited to see all of this come together 💯

PR #3794, opened 2023-01-18
Convert game responses to health responses for 2022 submissions

shaunxp20:
A few things here:
- I forgot to add the original migration (to convert game to health), so submissions on QA were never updated
- The migration I originally wrote wasn't working correctly
  - A few submissions weren't getting saved b/c they were failing a validation check (b/c the demo video was the same as the project video) 

I had to rewrite the migration to use `update_columns` to bypass validations; I also decided to make this migration non-reversible (without a `down` step).


viviancan (approved):
Did not check locally, but LGTM!

PR #3805, opened 2023-01-24
Default season to the current season for judge questions

shaunxp20:
A ChA can print out the scores for an RPE - this will loop over all the RPE submissions and their associated scores from each judge. But it is possible that a judge hasn't scored one of these submissions - in which case we create a new `SubmissionScore` for them but this won't have a season associated with it, and this missing season is what causes an error, when trying to get the judging questions for that season.

This PR will default the season to the current season if one wasn't provided when trying to fetch the judging questions.

### DIFF
--- a/app/services/judge_questions.rb
+++ b/app/services/judge_questions.rb
@@ -1,6 +1,8 @@
 class JudgeQuestions
+  attr_reader :season, :division
+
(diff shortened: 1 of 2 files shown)


viviancan (approved):
👍

Issue #3790, opened 2023-01-18
Update what is displayed on judge dashboard and judge submissions tabs


dboyer:
Currently /judge/dashboard and /judge/dashboard#judge-submissions-tab-content show the same information if judging is turned on. If judging is on, the two pages should be different.

At /judge/dashboard display:

<img width="847" alt="Screen Shot 2023-01-18 at 10 57 05 AM" src="https://user-images.githubusercontent.com/516764/213223372-22c7fdaf-2573-4733-8de1-7cc3d764f42c.png">

At /judge/dashboard#judge-submissions-tab-content:

<img width="848" alt="Screen Shot 2023-01-18 at 10 57 58 AM" src="https://user-images.githubusercontent.com/516764/213223800-ec81d9df-4fbb-4256-baac-d9e4166d7f2d.png">

dboyer:
Fall out issue from the different screens outlined in #3764

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/3790?planning-poker) with Zenhub @shaunxp20 @viviancan

PR #3806, opened 2023-01-25
Separated the judge dashboard tab from the judge submissions tab

viviancan:
Refs #3790 

This change separates the judge dashboard tab, the judge submissions, and the judge scores and certs tab. I also updated the side nav by indenting the consent waiver and judge training links. I am open to updating that, though, based on any additional feedback!  I also have a 2nd PR (coming later) related to this that cleans up some of the files. 

CleanShot 2023-01-25 at 10 28 59@2x


app/assets/javascripts/student-tabs.js
```diff
@@ -1,7 +1,7 @@
 $(document).on("turbolinks:load", function() {
     const anchor = $(location).attr('hash');
 
-    $("#tab-wrapper div a").click(function(e){
+    $("#tab-wrapper div a:not(#exclude)").click(function(e){
```
viviancan, line 4: I had to include this `not` selector because of the "consent waiver" and "judge training" links that are nestled in the tabs in the side nav

shaunxp20: FWIW, instead of adding the `not` selector, I think a more specific selector would work too, something to target links that are direct children:
````
$("#tab-wrapper div.tw-tab-content>a")
```

And also, FWIW, I think I understand this selector, but the name `exclude` isn't obvious to me what's being excluded at first glance. Looks like this is for the "navigation tabs", this could be changed to a `a.navigation-tag` selector too or something similar or the inverse (`non-navigation-tab`).

app/views/judge/dashboards/rebrand/_side_nav_content.html.erb
```diff
@@ -1,55 +1,63 @@
...
-          text: "Consent Signed",
+<div class="tw-tab-content">
+  <a id="dashboard-tab" href="#dashboard-tab-content" data-heading="Judge Dashboard">Dashboard</a>
+  <div>
+    <% if current_judge.consent_signed? %>
+      <%= web_icon(
+            "check-circle",
+            class: "icon--green ml-4",
+            text: "Consent Signed",
+            id: "exclude"
```
shaunxp20, line 9: Another FWIW, it doesn't look like `web_icon` needs the `exclude`, since the selector is looking for links. And I think the general rule is `id`s should be unique, and `class` names can be repeated. 😅 😃 😄

app/views/judge/dashboards/show.en.html.erb
```diff
@@ -1,19 +1,29 @@
...
-    <% end %>
+  <%= render layout: 'application/templates/dashboards/energetic_container', locals: { heading: 'Judge Dashboard'} do %>
+    <div class="tab">
+      <div id="dashboard-tab-content" class="tab-content tw-active" >
+        <% if !SeasonToggles.display_scores? %>
+          <%= render 'judge/dashboards/onboarded/rebrand/events' %>
+        <% end %>
+      </div>
+      <div id="judge-submissions-tab-content" class="tab-content">
+        <% if !current_judge.onboarding? %>
```
shaunxp20, line 16: Wondering if this can be changed to `current_judge.onboarded?` vs "not onboarding", just to make it more specific and consistent w/ the conditional below, either way, not a big deal! 😄

app/views/judge/dashboards/show.en.html.erb
```diff
@@ -1,19 +1,29 @@
...
-    <%end %>
-
-    <% if !SeasonToggles.display_scores? %>
-      <%= render 'judge/dashboards/onboarded/rebrand/events' %>
-    <% end %>
+  <%= render layout: 'application/templates/dashboards/energetic_container', locals: { heading: 'Judge Dashboard'} do %>
+    <div class="tab">
+      <div id="dashboard-tab-content" class="tab-content tw-active" >
+        <% if !SeasonToggles.display_scores? %>
+          <%= render 'judge/dashboards/onboarded/rebrand/events' %>
```
shaunxp20, line 12: Re: my comment [here](https://github.com/Iridescent-CM/technovation-app/pull/3808/files#r1089564864), if the `events` partial only contains event-related things, I think something like `current_judge.live_event?` could be incorporated into the conditional above, which I think would make it more explicit/specific to only display events to the relevant judges. 😄

shaunxp20 (approved):
I pulled this down locally and it's working as expected! I left some FWIW feedback, definitely not any blockers! Looking good, some nice improvements here! 🙌 💯 🎉 👏

app/views/judge/dashboards/rebrand/_side_nav_content.html.erb
viviancan: 🤦 totally made a mistake re: id/classes 😅

app/assets/javascripts/student-tabs.js
viviancan: Thanks for this! I think I was deep in the woods lol! I went with the more detailed selector and removed "exclude"

app/assets/javascripts/student-tabs.js
```diff
@@ -1,7 +1,7 @@
 $(document).on("turbolinks:load", function() {
```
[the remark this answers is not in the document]
viviancan: I updated the file name to `rebrand-tabs`

app/views/judge/dashboards/show.en.html.erb
viviancan: 👍

app/views/judge/dashboards/show.en.html.erb
viviancan: I also think there is additional cleanup with some of the logic and file names! My other branch was mainly pulling the different views out into their own partials. I will tackle all of the additional cleanup on an additional PR :)

PR #3808, opened 2023-01-25
Cleanup on the judge dashboard

viviancan:
Refs #3790 

Created 2 partials for the judge dashboard/tabs


app/views/judge/dashboards/onboarded/rebrand/_events.en.html.erb
```diff
@@ -1,63 +1,21 @@
 <div class="relative">
   <% if SeasonToggles.judging_finished? %>
```
shaunxp20, line 2: Very minor reordering suggestion, making the conditionals flow from "beginning to end" vs "end to beginning"... something like:
```
<% if SeasonToggles.quarterfinals_or_earlier? %>
...
<% elsif SeasonToggles.between_rounds? %>
...
<% elsif SeasonToggles.judging_finished? %>
```

app/views/judge/dashboards/onboarded/rebrand/_judging_finished.html.erb
```diff
@@ -0,0 +1,23 @@
+<div class="flags -top-2 -right-6">
```
shaunxp20, line 1: I'm not super familiar with the flags, but just curious if those styles^ can be added the `flags` class. Also, super minor thing re: indentation below this. 😄

shaunxp20 (approved):
Left some minor/non-blocker feedback, but overall LGTM! Some nice judging improvements! 👍 🎉 🙌

app/views/judge/dashboards/onboarded/rebrand/_events.en.html.erb
```diff
@@ -1,63 +1,21 @@
...
-        Chapter Ambassador. You can preview your event's submissions beginning
-        on <%= ImportantDates.quarterfinals_judging_begins.strftime("%B %-d, %Y") %>.
-      </p>
-
-      <%= render partial: 'regional_pitch_events/rebrand/event',
-                 locals: { multiple_events: true },
-                 collection: current_judge.events %>
-
+    </div>
+    <% if current_judge.live_event? %>
```
shaunxp20, line 16: A minor/random piece of feedback, as I was looking through the other open PR, I think I have my head a little better wrapped around the changes. FWIW, this file is named `_events`, but a lot of non-event related stuff is in here, I would think only event-related stuff (looks like this line and below) would be in this file, definitely not a blocker but I think it makes it a little less intuitive where to find/looks for things. 🤔 😅 😀

viviancan: I agree that `_events` is not very clear. Any renaming can be completed once we have the final text/states for the judging experience in #3824. I don't want to change it now and then have to change it again when we have the final judging flow then! I did not name these files, just did a little refactoring by creating the partials bc this file was difficult to navigate/follow! 😄

PR #3810, opened 2023-01-26
Update "parent/guardian email address can't match a student's email address" validations

shaunxp20:
There are two main fixes/changes here:
- The `StudentEmailValidator` will now add an error to `student_profile.parent_guardian_email` (it used to just add an error to `account.email`
- The division check in the `student_profile` validation will now match how it's done in `StudentEmailValidator`, meaning it will use `Division.for` to determine a student's division (that way both validations are consistent)

### DIFF
--- a/app/controllers/concerns/profile_controller.rb
+++ b/app/controllers/concerns/profile_controller.rb
@@ -27,8 +27,6 @@ module ProfileController
             success: t('controllers.accounts.update.success')
         }
       end
-    elsif profile.errors["parent_guardian_email"].any?
-      render "student/parental_consent_notices/new"
     elsif profile.errors["account.password"].any? or
       profile.errors["account.existing_password"].any?
       if profile.account.email_changed?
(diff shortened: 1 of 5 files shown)


app/validators/student_email_validator.rb
```diff
@@ -6,6 +6,7 @@ def validate(record)
         (record.email == record.student_profile.parent_guardian_email)
 
       record.errors.add(:email, :matches_parent_guardian_email)
+      record.student_profile.errors.add(:parent_guardian_email, :matches_student_email)
```
shaunxp20, line 9: 1. This is where (before) we were only adding an error to `account.email`, but now we're adding another error on `student_profile.parent_guardian_email` so it will surface an error on the parental consent form.

app/controllers/concerns/profile_controller.rb
```diff
@@ -38,6 +36,8 @@ def update
       end
     elsif profile.errors["account.email"].any?
       render 'email_addresses/edit'
+    elsif profile.errors["parent_guardian_email"].any?
+      render "student/parental_consent_notices/new"
```
shaunxp20, line 40: 2. And then I had to update this condition (b/c we added the additional error on `student_profile.parent_guardian_email`, this condition had to be moved below the `account.email` condition (above it), so that it retains the same behavior (that the tests are expecting).

viviancan (approved):
Tried it locally and works as expected!

PR #3815, opened 2023-01-31
Fix show/hide visibility of hunger/food waste description field

shaunxp20:
The show/hide functionality (for the additional questions) works by having things properly named, which I didn't quite get right the first time for the hunger/food waste question! 😊 

Here's a snippet for what should be toggled for the hunger/food waste question:
```
data: {
  toggles: {
    "true": "#solves_hunger_or_food_waste_description"
  },
},
```

In this PR, I updated the `id` (of the `div` that should be shown/hidden) to match that corresponding field. ☝️ 😅


viviancan (approved):
LGTM!

PR #3820, opened 2023-02-01
Update judge sections and questions for RPE printable page

shaunxp20:
This will update the judge sections and questions on the ChA RPE printable page.

One of the more noteworthy changes here is that I updated `sections_for` to be a class method, but also, instead of it returning an array, it will return a hash (with the contents of the array like it did before, but also it will return the display names for each section).

I think there will still be some work to do after this, but this will at least update the sections and questions to use what we currently have for the 2023 season.

### DIFF
--- a/app/models/questions.rb
+++ b/app/models/questions.rb
@@ -31,6 +31,7 @@ class Questions
       .call
       .uniq(&:section)
       .map(&:section)
+      .each_with_object({}) { |section, new_hash| new_hash[section] = section_display_name_for(section, division) }
   end
 
   def sections
@@ -153,5 +154,20 @@ class Questions
       text: score["#{section_name}_comment"],
       word_count: score["#{section_name}_comment_word_count"]
     }
+  end
+
+  def section_display_name_for(section, division)
+    case section
+    when "project_details"
+      "Project Description"
+    when "ideation"
+      "Learning Journey"
+    when "demo"
+      "Technical"
+    when "entrepreneurship"
+      division == "senior" ? "Business Plan" : "User Adoption Plan"
+    else
+      section.titlecase
+    end
   end
 end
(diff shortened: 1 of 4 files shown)


app/models/questions.rb
```diff
@@ -26,11 +35,7 @@ def in_section(section_name)
   end
 
   def sections_for(division:, season: Season.current.year)
-    JudgeQuestions
-      .new(division: division, season: season)
-      .call
-      .uniq(&:section)
-      .map(&:section)
+    self.class.sections_for(division: division, season: season)
```
shaunxp20, line 38: I converted this instance method to be a class method, and this will just call the class method now.

app/models/questions.rb
```diff
@@ -15,6 +15,15 @@ def self.for(score)
     new(score.judge_profile, score.team_submission, score: score)
   end
 
+  def self.sections_for(division:, season: Season.current.year)
+    JudgeQuestions
+      .new(division: division, season: season)
+      .call
+      .uniq(&:section)
+      .map(&:section)
+      .each_with_object({}) { |section, new_hash| new_hash[section] = section_display_name_for(section, division) }
```
shaunxp20, line 24: This is the only addition I made (from the original instance method) - this will basically return the section name like before and also the "display name", returning something like: `{ideation: 'Learning Journey' ... }`.

app/views/chapter_ambassador/printable_scores/show.en.html.erb
```diff
@@ -41,44 +41,23 @@
...
+
+      <% sections = Questions.sections_for(division: team.division_name) %>
+
       <div class="page <%= team.division_name %>">
         <header>
           <h3><%= judge.name %></h3>
           <h4><%= team.name %></h4>
 
           <table>
-            <tr>
```
shaunxp20, line 50: The changes below are basically removing the hard-coded sections and making them dynamic.

viviancan (approved):
💯 LGTM!

Issue #3817, opened 2023-01-31
Portugal participants are not displaying appropriately when filtered by state and city on admin portal and ChA portal


viviancan:

viviancan:
<img width="1748" alt="CleanShot 2023-02-01 at 11 41 45@2x" src="https://user-images.githubusercontent.com/29210380/216159998-2c748423-8ba0-46b5-b5db-f1ba34695471.png">

This is an interesting bug - and actually may not be a bug but rather an issue because of all of the different technologies that are used in this filtering system. 
The admin/chA location filter system utilizes the city-state gem and the values are prepopulated in the the dropdown by `country state select` gem as seen in `app/data_grids/accounts_grid.rb` line 530. 

The first part of this issue has to do with the prepopulated values in the location filters. 
When a user selects a location, a certain value is returned which is used to query our DB and return the appropriate results. In the screenshot, for example, if I select Portugal the value of "PT" is returned for the country filter. From there, all of the state_provinces are prepopulated. Now if I select a state from the next filter, for example, Lisbon, I would expect the value of "lisbon" to be returned and used for the query. However, that does not happen, instead, a value of "11" is returned. All of the state provinces in Portugal have an associated numerical value (Lisbon is 11, Braga is 3, and so on).

The 2nd part of this issue is related to how locations are stored in the DB for different users. I did not investigate this in depth, however did notice for some participants in Portugal, their state_province was saved as the numerical version. For example a person in Lisbon had the value of `11` instead of `Lisbon`. Other participants had the string "Lisbon". I do not know why some had 11 vs. Lisbon, but it seemed that there were only a few participants that had a numerical version of the location.

This is the reason not all participants were displayed when filtering by location. If I filtered by Lisbon, only participants with the state_province value of 11 would be displayed.

To fix this issue 2 changes were made in the PR. First, a translation/formatting map was created for the numerical version of locations to be converted to the string version. This approach was already in place since this seems to be a preexisting issue for a few other countries (MX, India, and Kenya). 2nd I created a rake task to convert the numerical state_provice records to the correct string version of that location. The rake task will convert a total of 54 records as seen in the chart below (for the state_province values 03, 05, 11, 13, 14, & 15)

| state\_province | count |
| :--- | :--- |
| 03 | 3 |
| 05 | 1 |
| 11 | 44 |
| 13 | 4 |
| 14 | 1 |
| 15 | 1 |
| Aveiro District | 4 |
| Coimbra District | 2 |
[...]

PR #3821, opened 2023-02-01
3817 standardize portugal location data

viviancan:
Refs #3817 - I am including a detailed summary on the original issue for historical context!

This PR includes 2 changes
1. Added the translation/format for Portugal in `app/models/state.rb` following the other approaches for other countries that have this issue
2. Added `lib/tasks/update_portugal_location_data.rake` to update state_province data and city data. I followed the approach used when updating Spain location data previously!


app/models/state.rb
```diff
@@ -44,6 +46,25 @@ def format_map
   end
 end
 
+class PortugalState < State
+  def format_map
+    {
```
viviancan, line 51: These numerical values come from the city-state gem

viviancan:
**QA Steps**

There are sort of 2 updates happening here. 1st the "translation" fix for numerical value to string value for state_province that occurs in `app/models/state.rb`. The 2nd is the rake task. Here are the qa steps. Please let me know if you have any clarifying questions!

QA Steps

0. Import paticipant data locally
1. On local QA branch login as admin or ChA in portugal
*** This is without the numerical translation ***
2. Filter by country = Portugal
- There should be 336 total accounts
3. Filter by the following states and verify number of participants 
- Aveiro 0
- Braga	3
- Castelo Branco 1
- Coimbra 0
- Evora	0
- Faro	0
- Leiria 0
- Lisbon 44
- Madiera 0
- Porto	4
- Santarem 1
- Setubal 1



Checkout `3817-standardize-portugal-location-data` branch
1. Login as admin or ChA in Portugal
2. Go to participants
*** This branch contains the numerical translation  ***
3. Filter by country = Portugal
- There should be 336 total accounts
4. Filter by the following states and verify number of participants 
- Aveiro 4
- Braga	0
- Castelo Branco 0
- Coimbra 2
- Evora	1
- Faro	12
- Leiria 7
- Lisbon 148
- Madiera 1
- Porto	83
- Santarem 7
- Setubal 17

5. Filter by state = lisbon and city = lisbon
- There should be 126 results


6. Run the following rake tasks
- `update_portugal_state_province_data`
- `update_portugal_city_data`
*** This is with the numerical translation + the updated state and city locations from the rake tasks ***
7. Filter by the following states and verify number of participants 
- Aveiro 4
- Braga	3
- Castelo Branco 1
- Coimbra 2
- Evora	1
- Faro	12
- Leiria 7
- Lisbon 192
- Madiera 1
- Porto	87
- Santarem 8
- Setubal 18

8. filter by state = lisbon and city = lisbon
- There should be 165 results

lib/tasks/update_portugal_location_data.rake
```diff
@@ -0,0 +1,64 @@
+desc "Update accounts in Portugal to have the correct state_province location data"
+task update_portugal_state_province_data: :environment do
```
viviancan, line 2: This first rake task will convert the accounts that have the numerical version of the state_province to the string value of that location. For ex. some accounts of state_province value of `3` which is Braga. So those accounts have been updated.

lib/tasks/update_portugal_location_data.rake
```diff
@@ -0,0 +1,64 @@
...
+    163739
+  ]
+
+  Account.where(id: setubal_account_ids).update_all(state_province: "Setubal", country: "PT")
+
+end
+
+desc "Update accounts in Portugal to have the correct city location data"
+task update_portugal_city_data: :environment do
+  lisboa_account_ids = %w[
```
viviancan, line 59: This converts city data from "Lisboa" to "Lisbon"

shaunxp20:
Thanks for the detailed QA steps!

I pulled down production data, and tried it out and things are working as expected from what I can tell.

My numbers were slightly off though - I didn't check all of the state/regions - but for example, for Lisboa, I had 45 accounts (not 44) on the QA branch, 160 (not 148) on this branch, and 204 (not 192) after running the rake tasks. I think this just means more people registered. Just curious, will we have to run a rake task fix as the season goes on and/or at the end of the season to clean up the data (for future cases we run into)? Either way, I think this fix is doing what it's supposed to! Going to approve it!

shaunxp20 (approved):
LGTM! Thanks for tackling this one! 👍 🙌 💯 🥳

app/models/state.rb
shaunxp20: I see! Thanks for the explanation! I think I understand, this part here is where it will fix it for the admin/ChA area, and the rake task will update the data for when we do exports (so it shows the state/region name instead of a numerical value).

PR #3828, opened 2023-02-03
Add city and state to ChA activity page

shaunxp20:
This will add a city and state field/selector for the ChA activity page.


viviancan (approved):
Did not try locally, but looks pretty straight forward! Thx, Shuan 💯

PR #3829, opened 2023-02-06
Update student email link for pitch events

shaun-technovation:
This will update the link that is displayed in an email to the students about the pitch event - it will update the link so it goes to the events page/tab.


viviancan (approved):
LGTM!

Issue #3765, opened 2023-01-04
Update text for judging questions


dboyer:
Update text for questions as shown in these tracked changes docs

[Beginner Division](https://docs.google.com/document/d/1mYgrerIrkf80CWweE7DaQOv1NjpZPdOjVDAxb8uvIsE/edit)

 [Junior Division](https://docs.google.com/document/d/1Xoi_M9Ijtcm6bpRAPuLUp9o1_EYbKL-q7uJqstfjSno/edit)

 [Senior Division](https://docs.google.com/document/d/1QwVsf_4GGXzmBvwgiCZ2y5-uDp6BFJgi5sUASsF4Ngs/edit)

_Originally posted by @dboyer in https://github.com/Iridescent-CM/technovation-app/issues/3745#issuecomment-1370105386_

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/3765?planning-poker) with Zenhub @shaunxp20 @viviancan

dboyer:
Specific text changes are below. The changes are very similar between divisions but I have listed each division separately in case that makes it easier to edit. The full doc that program team reviewed [is here](https://docs.google.com/document/d/17tlQVdyI3Mm0DAbqGfuWicIbYAWUxT31NNa2teSshiQ/edit#).

**Beginner Division Changes:**

- [ ] Change the question under project description to "Is our 100 word problem/project description clear and compelling?"

- [ ] In the technical section if the project is a mobile app, the three questions should be:
Do we show what app we built and what parts work successfully so far?
Do we explain the coding we did for 1 or 2 important parts of our app, other than the login screen?
Do we show what doesn’t work yet and/or share future app features?

- [ ] In the technical section if the project is an AI project, the three questions should be:
#1 (no change)
Update #2: Do we show the invention we built or prototyped? Do we explain how we built it? Do we show the parts that work?
#3. Do we show what doesn’t work yet and/or share future prototype features?

- [ ] In the learning journey section, update the first question to read as follows. It is okay that there are two questions in this point: Do we share what we learned through a combination of words and pictures (eg screenshots, prototypes)? Do we share any technical sources used/remixed and/or our favorite technical resource?

**Junior Division Changes:**

- [ ] Change the hover text for the 4 in the scoring bubbles to read "Impressive"

[...]

PR #3840, opened 2023-02-13
Update 2023 judging questions

shaunxp20:
This should update the judging questions as specified here: https://github.com/Iridescent-CM/technovation-app/issues/3765#issuecomment-1428047492


viviancan (approved):
Did not try locally, but looks pretty straightforward! Thanks for tackling this, Shaun 💯

PR #3841, opened 2023-02-13
Adjust RPE judge assignment threshold

shaunxp20:
For RPEs, in order to manually assign a judge to a submission, a certain number of teams needs to be participating in the event.

This PR will update that number to 8, so that if 8 or more teams are participating, judges can be manually assigned to submissions.


viviancan (approved):
👍

Issue #3789, opened 2023-01-18
Add judging format column to judge grid


dboyer:
With the return of RPEs, we would like to note if a submission is judged as part of an RPE or online only. We should add a judging format column to any of the grids that lists submissions for a judge (i.e. submissions to score, in progress, finished scores). The column is visible in this mock-up. 

Let's use the phrases "Pitch Event" and "Online" for the judging format.

![Screen Shot 2023-01-30 at 7.53.11 PM.png](https://images.zenhubusercontent.com/5ca619b412380971eb9c9477/28f13b27-7ee8-4c66-8a68-c9f1dbf9f951)

dboyer:
Connected to issues outlined in #3764

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/3789?planning-poker) with Zenhub @shaunxp20 @viviancan

viviancan:
Verifying this is on QA

dboyer:
Confirming I see this column on QA. Super exciting!!!!

We may want "pitch event" and "online" rather than live and virtual for the event names. I'll confirm with PT and enter that as a separate text issue along with other text changes.

dboyer:
Let's change the wording to Pitch Event and Online

dboyer:
Confirming I see the words Pitch Event and Online now on QA

PR #3842, opened 2023-02-14
Updated judging format column text (using Pitch Event and Online)

viviancan:
Refs #3789 

Updated the judge score grids to display the correct name in the judging format column

### DIFF
--- a/app/models/submission_score.rb
+++ b/app/models/submission_score.rb
@@ -293,6 +293,10 @@ class SubmissionScore < ActiveRecord::Base
     team.selected_regional_pitch_event_name
   end
 
+  def event_type_display_name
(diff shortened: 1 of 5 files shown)


app/models/submission_score.rb
```diff
@@ -293,6 +293,10 @@ def event_name
     team.selected_regional_pitch_event_name
   end
 
+  def event_type_display_name
```
viviancan, line 296: Created this method to display the correct event type name. The naming choice was finalized in dev priorities

shaunxp20 (approved):
Looks good/straight forward to me! 👍 🙌 🎉 :shipit:

Issue #3837, opened 2023-02-13
Chapter Ambassador RPE admin page changes


dboyer:
Small to mid-sized changes requested to ChA RPE admin pages (/chapter_ambassador/events#/) as part of testing:

- [x] Add hover text to the edit icon (text: Edit event) and delete icon (text: Delete event). The other icons have hover text

<img width="416" alt="Screen Shot 2023-02-13 at 11 31 59 AM" src="https://user-images.githubusercontent.com/516764/218515835-08097c54-065f-4ee8-a0e0-80aca080ae53.png">

- [x] When creating or editing an event, can we add text near the time zone saying "Please select the time using a 24 hour clock. The time will display to students as the local time zone in your area."

- [x] If an event is full, ChAs cannot add more teams from the events page as the button to select a team is greyed out. However, there is no message about why they cannot select a team. Add a message saying: This event is currently at capacity. No additional teams can be added. 

<img width="978" alt="Screen Shot 2023-02-13 at 3 34 47 PM" src="https://user-images.githubusercontent.com/516764/218568786-98af6048-9e6c-44a7-86c8-b8fd1015cd16.png">

shaunxp20:
These changes should be on QA now! 😄

dboyer:
Confirmed that these changes are correct on QA.

PR #3843, opened 2023-02-14
ChA RPE admin updates

shaunxp20:
The changes listed below (from #3837) are in this PR:
- Add hover text to the edit icon (text: Edit event) and delete icon (text: Delete event).
- When creating or editing an event, can we add text near the time zone saying "Please select the time using a 24 hour clock. The time will display to students as the local time zone in your area."
- If an event is full, ChAs cannot add more teams from the events page as the button to select a team is greyed out. However, there is no message about why they cannot select a team. Add a message saying: This event is currently at capacity. No additional teams can be added.


viviancan (approved):
Did not try locally, but LGTM! 💯

PR #3845, opened 2023-02-14
Reorder judge questions/sections

shaunxp20:
The order of the judge _sections_ (for the admin and students) is implicitly based on the ordering of the judging _questions_:
https://github.com/Iridescent-CM/technovation-app/blob/4be85d37ab8c747aac4be9e014e9fe97902ab1b8/app/models/questions.rb#L18-L25

The changes in this PR will reorder the Learning Journey (aka ideation) questions so the Learning Journey section/questions will display last.


viviancan (approved):
LGTM 💯

Issue #3827, opened 2023-02-03
Make pitch events page accessible even if judging is turned on


dboyer:
Currently, if judging is turned on, students do not see any information on the pitch events page (/student/dashboard#/events).

It is correct that students shouldn't be able to select an RPE once judging is turned on. However, they should still be able to see information about the event they are attending. Could we make the event info visible but remove the sign up button?

<img width="701" alt="Screen Shot 2023-02-03 at 11 11 36 AM" src="https://user-images.githubusercontent.com/516764/216651789-41e453c8-5c22-4f0d-a4d1-aaa7390211cb.png">

viviancan:
I attached screenshots of all of the different "states" for the pitch your project tab

[https://app.zenhub.com/files/29868115/367e828f-6bb1-4ba3-b8a7-9b47e828f3d1/download](https://app.zenhub.com/files/29868115/367e828f-6bb1-4ba3-b8a7-9b47e828f3d1/download)

viviancan:
Confirming this is on QA!

dboyer:
Confirming that if I'm attending an event, I can see the event details even if judging is on. If I'm not attending an event already, I cannot see event details as the time period to select events closes once judging opens.

PR #3846, opened 2023-02-14
Make pitch event details accessible for RPE teams on the student dashboard when judging is on or off

viviancan:
Refs #3827 

This change allows for students that are registered for an RPE to view the event details when judging is on or off.

I also added some specs.

### DIFF
--- a/app/views/slots/student/rebrand/_events.en.html.erb
+++ b/app/views/slots/student/rebrand/_events.en.html.erb
@@ -4,9 +4,10 @@
   </div>
 
   <div class="tw-content=-wrapper p-8">
-    <% if current_student.onboarded? && SeasonToggles.select_regional_pitch_event? %>
+    <% if SeasonToggles.select_regional_pitch_event? || current_team.live_event? %>
       <%= render "student/dashboards/completion_steps/rebrand/regional_pitch_events" %>
-    <% else %>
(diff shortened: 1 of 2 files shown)


app/views/slots/student/rebrand/_events.en.html.erb
```diff
@@ -4,9 +4,10 @@
   </div>
 
   <div class="tw-content=-wrapper p-8">
-    <% if current_student.onboarded? && SeasonToggles.select_regional_pitch_event? %>
```
viviancan, line 7: I removed the `.onboarded` because students cannot even see this if they are not onboarded

app/views/slots/student/rebrand/_events.en.html.erb
```diff
@@ -4,9 +4,10 @@
...
+    <% if SeasonToggles.select_regional_pitch_event? || current_team.live_event? %>
```
viviancan, line 7: Students should be able to see event partial (which includes event details) when the select regional pitch events is enabled or if they are registered for an RPE (which takes care of the case when judging is on/off)

shaunxp20 (approved):
I pulled this down and tried it locally and confirmed I could see my RPE regardless of the judging/RPE admin controls! Nice tests! LGTM! 👍 🙌 🎉 🥳 :shipit:

PR #3862, opened 2023-02-17
Add JS URL validator for RPE links

shaunxp20:
This will add a very basic JS validation to ensure that the RPE links start with http:// or https://


viviancan (approved):
LGTM! 💯 🚀

PR #3864, opened 2023-02-23
Update v-html instances

shaunxp20:
This will remove `v-html` for the project description and learning journey when judging a submission.

For the project description, we were doing formatting on the Rails side, using `simple_format` which added `<p>` tags before it got to Vue.js. I removed this backend formatting and moved it to the Vue.js layer using `white-space: pre-line`. For some unknown reason, when I tried using the Tailwind class it wouldn't work, but adding it via `style` did. 🤷 

https://developer.mozilla.org/en-US/docs/Web/CSS/white-space#pre-line
https://tailwindcss.com/docs/whitespace#pre-line


viviancan (approved):
Tried it locally and worked as expected! 💯

Issue #3836, opened 2023-02-09
Investigate location issue re: Bishkek 


viviancan:

viviancan:
refs: 3545

viviancan:
I believe the participants are not able to be filtered because their location is not associated with a state/province. For some reason the geolocation is not assigning a state/province when they set their location.

 There are few of the `Bishkek` participants that are associated with `Chüy Region` and those individuals are filtered appropriately. 

When setting a participant location to `Bishkek` the state/province field comes back empty. I can only get the form to load Chüy Region using a very specific set of steps (shown in the attached video)

https://drive.google.com/file/d/1thjKSymzuIygVW1Fy3ejPNBNgGqZYjL2/view?usp=sharing


this may require data cleaning on our end.

viviancan:
Interesting data comparing 2022 season to 2023 season

Bishkek Stats

**2022**
`state_province`
`EMPTY` (' ') = 6
`NULL` = 0
Total = 6 accounts with missing `state_province` information

**2023**
`state_province`
`EMPTY` (' ') = 315
`NULL` = 14

Total = 329 accounts with missing `state_province` information

I think we may need to create a rake task to find accounts in `2023` with the city `Bishkek` and an `empty` or `null` state_province field, and update that value to `Chüy Region`. This would also allow us to run the rake task while registration is still open and also at the end of the season to clean up this location data.

viviancan:
@shaunxp20 @dboyer Thoughts on this approach?

viviancan:
Ok! 1 additional update here, we actually have a rake task called `fix_empty_states.rake` which does exactly what it says 🎉 With a little bit of tweaking I was able to run it against some local dummy data. It seems to work as expected and even picked up on some other missing state accounts that I had locally. The only downside is the reverse geocoding is registering the city `Bishkek` in the state_province of `Bishkek City` instead of `Chüy Region`. Chüy Region has been the confirmed location by the local ChA. I'm not sure why the reverse geocoder is returning bishkek city. Even if we left it as Bishkek City, our current location filters do not filter by that value anyway. So I do think the path that I mentioned above is the way to go. I was able to create a rake task mirroring the approach of the `fix_empty_states rake` task to update my local dummy data successfully for the accounts with missing state_province in Bishkek. Next steps are to confirm the fix with production data.

dboyer:
[...]

PR #3867, opened 2023-02-28
Added rake task to update bishkek location data

viviancan:
Refs #3836 

This rake task will find all accounts located in the city Bishkek/ country Kyrgyzstan with incorrect state_province data.
The state_province is considered incorrect if it is EMPTY, NULL, or not "gorod bishkek". 

I went with a more general approach instead of by ID in case we need to rerun this in the future.

QA Steps
-----------
1. Pull prod data
2. Filter by Country: Kyrgyzstan, State: Gorod Bishkek  (there should be 0 results) 
3. Query the number of participants with incorrect state_province data in the city of Bishkek (Note as of 2/27/23 that value is 393)
4. Run the rake task dry to get the count or actually run it to update
 `bundle exec rake 'update_bishkek_state_location_data[dry]'` 
 or 
 `bundle exec rake 'update_bishkek_state_location_data[run]'` 
5. Check the number of participants from step 3 (should be 0 now)
6. Login as an admin
7. Filter by Country: Kyrgyzstan, State: Gorod Bishkek, City: Bishkek (there should be 393 results)

### DIFF
--- a/app/models/state.rb
+++ b/app/models/state.rb
@@ -12,6 +12,8 @@ class State
       KenyaState.new(value)
     elsif countries.include?("PT")
       PortugalState.new(value)
+    elsif countries.include?("KG")
+      KyrgyzstanState.new(value)
     else
       new(value)
     end
@@ -83,6 +85,14 @@ class IndiaState < State
   def format_map
     {
       "TG" => "telangana",
+    }
+  end
+end
+
+class KyrgyzstanState < State
+  def format_map
+    {
+      "GB" => "gorod bishkek"
     }
   end
 end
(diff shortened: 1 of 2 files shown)


shaunxp20 (approved):
> 6. Login as an admin
> 7. Filter by Country: Kyrgyzstan, State: Gorod Bishkek, City: Bishkek (there should be 393 results)

For step 7, no results were showing in the admin, I discovered that I had to switch to the part 2 branch to see results in the admin.

Anywho, I think that means both parts are working as they should. Thanks for tackling yet another location issue! 🍪 🙇 🙌 🎉 🚢

PR #3868, opened 2023-02-28
Added format map for Kyrgyzstan to the state model 

viviancan:
Refs #3836 

This is similar to the translations needed for other countries (like MX and PT).

When filtering by country: Kyrgyzstan and state: Gorod Bishkek the value that is returned by the location filter is `GB`. This is causing issues when searching by `gorod bishkek`. This translation will map `GB` to `gorod bishkek` so the filtering works as expected.


shaunxp20 (approved):
👍

PR #3871, opened 2023-03-03
Update Checkr API

shaunxp20:
There are two main changes we had to make to incorporate the Checkr changes:
- Add support for the `result` parameter in the Reports payload
- Add support for the `canceled` value for status

From their documentation:
> If you have any automation tied to a report `status` of `clear` or `consider`, it should be adjusted to tie to the new report `result` value of `clear` or ` consider` instead.

> If any automation is built based on the report’s status, the new `canceled` status should be taken into account.

If you want to test locally, you'll have point your local environment to our Checkr staging environment:
```
CHECKR_API_BASE=https://api.checkr-staging.com 
CHECKR_API_KEY=[[in the password safe]]
```

And then you need to use one of the test candidates here:
https://docs.google.com/spreadsheets/d/11ZeECuXArp7x3DoBXW-pz-Lgbe5uHvpCGGK1ZrdOMkU/edit#gid=0

Any existing mentor (in the US) will work, their details don't need to match the test candidates in that^ spreadsheet, but on the Background Check form, that's where the test candidate details needs to match what's on the spreadsheet.

### DIFF
--- a/app/technovation/background_checking.rb
+++ b/app/technovation/background_checking.rb
@@ -9,26 +9,34 @@ class BackgroundChecking
 
   def execute
     if @report.present?
-      new_status = "engaged" == @report.adjudication ? "clear" : @report.status
+      case @report.result
+      when "clear"
+        @bg_check.clear!
+        do_clear
+      when "consider"
+        @bg_check.consider!
+      else
+        new_status = "engaged" == @report.adjudication ? "clear" : @report.status
 
-      if new_status != @bg_check.status
-        if @bg_check.respond_to?("#{new_status}!")
-          @bg_check.send("#{new_status}!")
-          @logger.info(
-            "Report UPDATED TO #{@bg_check.status.upcase} for #{@bg_check.account.email}"
-          )
+        if new_status != @bg_check.status
+          if @bg_check.respond_to?("#{new_status}!")
+            @bg_check.send("#{new_status}!")
+            @logger.info(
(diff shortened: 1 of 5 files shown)


app/technovation/background_checking.rb
```diff
@@ -9,26 +9,34 @@ def initialize(bg_check, options={})
 
   def execute
     if @report.present?
-      new_status = "engaged" == @report.adjudication ? "clear" : @report.status
+      case @report.result
```
shaunxp20, line 12: Here is where we're checking for the new `result` parameter, if the results is not "clear" or "consider" the code will fallback to what it is currently doing.

app/technovation/background_checking.rb
```diff
@@ -9,26 +9,34 @@ def initialize(bg_check, options={})
...
+      when "clear"
+        @bg_check.clear!
+        do_clear
+      when "consider"
+        @bg_check.consider!
+      else
+        new_status = "engaged" == @report.adjudication ? "clear" : @report.status
 
-      if new_status != @bg_check.status
```
shaunxp20, line 14: Just FYI, the changes below are just whitespace/indenting changes, if you select the "Hide whitespace" option, these changes won't be displayed.

viviancan (approved):
LGTM! I tried to test the mock accounts locally but had trouble with a bad request. I may have an old API key or am missing a step!! Either way, happy to keep the ball rolling on this to get some additional eyes 👀 💯 Thanks for tackling this one, Shaun!!

PR #3876, opened 2023-03-07
Add REDIS_URL to Sidekiq configuration

shaunxp20:
This is totally a "throw it against the wall and see if it will stick" PR.

Reviewing our SSL errors, it looks like we already incorporated some fixes, but I noticed that we're not including the `REDIS_URL` which is mentioned in the Heroku documentation and the following blog post:
- https://devcenter.heroku.com/articles/connecting-heroku-redis#connecting-from-sidekiq
- https://www.sandipmane.dev/openssl-connection-error-with-redis6-on-heroku

### DIFF
--- a/config/initializers/sidekiq.rb
+++ b/config/initializers/sidekiq.rb
@@ -4,7 +4,13 @@ Sidekiq.default_worker_options = {
 }
 
 Sidekiq.configure_server do |config|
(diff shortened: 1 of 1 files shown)


viviancan (approved):
Did not try locally, but looks like it could "stick" to the wall! 👍 💯  Thx Shaun!

Issue #3824, opened 2023-02-02
Update judge dashboard/different "states" text 


viviancan:

viviancan:
Refer to the [mural board ](https://app.mural.co/t/technovationgirls4378/m/technovationgirls4378/1674590684432/0f93038bbd4755c67632062ee6ab16cf1a1c6c90?sender=u9adab98cc36af2c4665e9477)

We need to update the text for the judge dashboard depending on the state of season

dboyer:
Add the following text:

- [ ] On the judge dashboard (/judge/dashboard#/) when judging is off and the judge is fully onboarded, the [text should read](https://docs.google.com/document/d/1-EpzEqGvSrYwXKbDw5vFvLn_E6svPlft5oT6CKXCK0w/edit#heading=h.rib14y9twmtd):

Welcome to the online judging portal! Judging is currently closed but will open soon. You will receive an email when judging opens. Please check back on the dates listed below!  DATE CHART

- [ ] On the judge dashboard when judging is on for quarterfinals. [Text shown here](https://docs.google.com/document/d/1-EpzEqGvSrYwXKbDw5vFvLn_E6svPlft5oT6CKXCK0w/edit#heading=h.ivur99ueghfl).

Welcome to the online judging portal for the first round of Technovation Judging! Click on the button below to open a submission and start scoring. 
Button: Start a new score (this can be the same button on the judge submissions tab)
<img width="118" alt="Screen Shot 2023-03-07 at 10 07 39 AM" src="https://user-images.githubusercontent.com/516764/223462497-912fbde2-06d0-4ad0-a0c9-f3ac75c44f01.png">

DATE CHART

- [ ] On the judge dashboard [in-between rounds](https://docs.google.com/document/d/1-EpzEqGvSrYwXKbDw5vFvLn_E6svPlft5oT6CKXCK0w/edit#heading=h.y3cialsnj520) the only changed text is after "What happens next?" That should read:

If you finished scoring 5 submissions or more, you will be invited to judge in our semifinal round starting on DATE.
Semifinals will take place from DATE, the number of scores you finish will determine what judge certificate you receive.
Your scores from the semifinal round will determine the top 15 teams that will be finalists in our World Summit event in October of 2023!

[...]

PR #3880, opened 2023-03-09
Cleanup of judge dashboard views - 3824pt2

viviancan:
Refs #3824 - part 2

Additional cleanup of the judge dashboard views. Deleted old event partial that was no longer used and moved/renamed other partials


shaunxp20 (approved):
LGTM, nice clean up! I'm glad we got rid of `_events.en.html.erb` 👍 💯 🎊 🚢

Issue #3885, opened 2023-03-22
Optimize team submission screenshots on user upload with filestack


viviancan:

viviancan:
Refs #2

viviancan:
Refs [ZH Issue #2](https://app.zenhub.com/workspaces/technovation-605271851956b1000ec2f24d/issues/zh/2)

viviancan:
Steps
1. Limit all photos to 1mb (max size)
2. Use filestack image tag for image display
3. Use the `auto_image` or `compress` transformation methods. We only have 25,000 transformations a month, so I think we should increase that amount. Transformations are ideal because they will store a cached version of the image that is accessible for 30 days. Each time the photo is accessed it will not count as an additional transformation.

viviancan:
**Example stats with the transformations**
- no transformation (original image): 898kb
- auto_image only: 260kb
- resize only: 59.1kb
- compress only: 681kb
- resize + compress: 52kb
- auto_image + resize + compress: 21kb

viviancan:
Confirming the 2MB size limit is on QA for 
- Admin upload a submission image
- Student upload a submission image

dboyer:
Confirming that I could only upload images below 2mb and received a warning message when I attempted to upload larger images. This worked for uploading as both a student and an admin.

viviancan:
Confirmed this is on production and working as expected for student image upload and admin image upload. Tested on team - Nov 29 Test Team

dboyer:
Also tested on production and confirmed I could only upload an image that was less than 2mb

PR #3888, opened 2023-03-27
Extract filestack picker options (in rails) + add 2mb image size maximum for submissions 

viviancan:
Refs #3885 

2 main changes here:
- Extracted filestack picker options to a helper method
- Added a 2mb file size limit for uploading submission images

I also updated the language from "screenshots" to "images" to align with the student experience. There will be a 2nd PR to add the file size limit on the student submission image upload.  

I could use some feedback on the "aws_path" method!

### DIFF
--- a/app/helpers/filestack_picker_helper.rb
+++ b/app/helpers/filestack_picker_helper.rb
@@ -1,0 +1,27 @@
+module FilestackPickerHelper
+  def filestack_picker_options(object)
+    {
+      accept: ["image/jpeg", "image/jpg", "image/png"],
+      uploadConfig: {
+        intelligent: true
+      },
+      maxSize: 2 * 1024 * 1024, # 2 MB LIMIT
+      fromSources: ["local_file_system"],
+      onFileSelected: "onFileSelected",
+      onFileUploadFinished: "onFileUploadFinished",
+      storeTo: {
+        location: "s3",
+        container: ENV.fetch("AWS_BUCKET_NAME"),
+        path: aws_path(object),
+        region: "us-east-1"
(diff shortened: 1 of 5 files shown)


app/helpers/filestack_picker_helper.rb
```diff
@@ -0,0 +1,26 @@
...
+        location: "s3",
+        container: ENV.fetch("AWS_BUCKET_NAME"),
+        path: aws_path(klass, record),
+        region: "us-east-1"
+      }
+    }
+  end
+
+  def aws_path(klass, record)
+    if klass.safe_constantize == TeamSubmission
```
viviancan, line 22: If I go the route of using a string like "TeamSubmission" + safe_constantize to represent the class I will need to use a regular if/else statement.

Another option is 
  ```
case klass
when TeamSubmission
    "uploads/screenshot/filestack/#{record.id}/"
 end
```

and just passing the instance variable like `@team_submission`

**Review by shaunxp20 — APPROVED** [review]
I didn't test it locally, but seems pretty straight forward, nice extraction w/ the options! 👍 🙌 :shipit:

File: app/helpers/filestack_picker_helper.rb [hunk 2]
```diff
@@ -0,0 +1,28 @@
...
+        path: aws_path(record),
+        region: "us-east-1"
+      }
+    }
+  end
+
+  private
+
+  def aws_path(record)
+    if record.is_a?(TeamSubmission)
```
viviancan, line 24: @shaunxp20 Can you take a look at this! Made an update!

PR #3889, opened 2023-03-27
Added 2MB max size for image uploads in student submission

viviancan:
Refs #3885 

Pt2 
This change adds 2MB max size for image uploads in the student submission flow

### DIFF
--- a/app/javascript/components/ScreenshotUploader.vue
+++ b/app/javascript/components/ScreenshotUploader.vue
@@ -143,6 +143,7 @@ export default {
 
       const fsPickerOptions = {
         accept: ["image/jpeg", "image/jpg", "image/png"],
+        maxSize: 2 * 1024 * 1024,
         fromSources: ["local_file_system"],
         maxFiles: this.maxFiles,
         storeTo: {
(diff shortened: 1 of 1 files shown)


app/javascript/components/ScreenshotUploader.vue
```diff
@@ -143,6 +143,7 @@ export default {
 
       const fsPickerOptions = {
         accept: ["image/jpeg", "image/jpg", "image/png"],
+        maxSize: 2 * 1024 * 1024,
```
shaunxp20, line 146: Random thought to abstract `2 * 1024 * 1024` to something like `const TWO_MB_FILE_SIZE = 2 * 1024 * 1024`, and then referencing that: `maxSize: TWO_MB_FILE_SIZE` (or whatever it would be called), FWIW. 😄

shaunxp20: `MAX_FILE_SIZE`? 🤷

shaunxp20 (approved):
Looks good/straight forward to me! Seems like a nice improvement, limited the file size! 👍 💯 🎉

PR #3893, opened 2023-03-28
Confirmation dialog and info text for saving production system settings 

shaunxp20:
A couple of changes to help us distinguish between QA and Prod (and our local env):
- Dynamic info text (above the save button) depending on the environment:
  - For our local environment: "The changes you are about to make are for your local environment."
  - For QA: "The changes you are about to make are for QA."
  - For Prod: "The changes you are about to make are for PRODUCTION. Please double check everything before saving."
- A confirmation dialog that will only display for Prod

### DIFF
--- a/app/javascript/admin/content-settings/components/AdminContentSettings.vue
+++ b/app/javascript/admin/content-settings/components/AdminContentSettings.vue
@@ -135,7 +135,7 @@
         ref="submitButton"
         type="submit"
         class="button primary"
-        @click.prevent="saveSettings"
+        @click.prevent="isProduction() ? confirmSaveSettings() : saveSettings()"
       >Save these settings</button>
       or
       <a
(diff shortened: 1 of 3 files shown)


viviancan (approved):
Tried locally and I saw the blue box with local env message!

viviancan:
Thanks for tackling this one! The only feedback I have is a different color for the production popup. Not sure if we can make the background a light red or something that is very obvious without reading the text.

Also I totally missed the blue box for the local environment message the first time around, I think the style sort of blended in with the rest of the page. Now that I know it's there though I will see it!

dboyer:
I agree with the suggestion of a different color for production!

shaunxp20:
I added a light red background, I picked one from Tailwind, but happy to change the color to whatever!

Issue #3890, opened 2023-03-28
Update how submission images are displayed in judging portal (using transformations)


viviancan:

viviancan:
Waiting for one final response from filestack

dboyer:
Confirming the images look correct in the judging portal and are high quality both as thumbnails and when I click to view the full size image.

PR #3904, opened 2023-04-03
Added filestack image url transformation method for submission images on judging portal

viviancan:
Refs #3890 

This change was needed to compress submission images to reduce bandwidth on the judging portal. Initially there were 3 main filestack transformations to reduce file size: resize, auto_image, and compress. Since we have limited transformations per month, I could only go with one. I chose the "resize" transformation because it reduced the file size the most compared to auto_image and compress. For example, on one sample submission the image sizes were reduced from 7.2mb to 131kb and 2.7mb to 93.9kb. Also noting there shouldn't be too many additional images in the future over 2mb with the updated file size limit.

### DIFF
--- a/app/javascript/judge/scores/pieces/Screenshots.vue
+++ b/app/javascript/judge/scores/pieces/Screenshots.vue
@@ -11,7 +11,7 @@
     >
       <img
         class="judge-screenshot-modal object-cover h-full w-full rounded"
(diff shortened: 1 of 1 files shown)


shaunxp20 (approved):
Looks good/straight forward to me! Nice improvement w/ the resizing! 🙌 🎉 👏 💯

Issue #3907, opened 2023-04-05
Update judge text in the judge submissions tab for virtual judges / verify text for rpe and virtual judges is correct on all tabs


viviancan:
I noticed the text on the judge submissions tab for virtual judges is incorrect
![CleanShot 2023-04-05 at 15.18.33@2x.png](https://images.zenhubusercontent.com/606b0f4fd13d53dff3d14fa5/9a669729-046d-4933-9323-8c2525ad88ee)

There should not be reference to "assigned submissions" or "chapter ambassadors"

viviancan:
Sort of related to #3824 - just a different tab!

dboyer:
@viviancan Can we just take out the "Submissions to Score" header and text below if there aren't any assigned scores and just have the green "start a new score" button? Or do we need to have that header there? 

I'm wondering if that header being there is fall out from #3869

viviancan:
Yes, we can remove and update the text. I think that text should really only be displayed if they are RPE judges. The submissions to score table should only display once they have started a new score.

dboyer:
Agreed!

dboyer:
Confirming the text is now correct for virtual judges. I checked for both judges with and without admin assigned submissions to judge.

PR #3912, opened 2023-04-10
Updated logic/text that displays for virtual judges on the judge submissions tab

viviancan:
Refs #3907 

Slight update to the logic that displays the "start score" button and the "Submissions to Score" table. This change was needed because information for live event judges was displayed even for virtual judges.


shaunxp20 (approved):
LGTM! 🙌 🎉 👍

PR #3913, opened 2023-04-10
Fix RPE tests

shaunxp20:
Just a few updates to fix the RPE tests.


viviancan (approved):
Did not try locally, but LGTM! 💯

PR #3924, opened 2023-04-13
Add additional columns to admin datagrid pages

shaun-technovation:
This will allow admins to select/view these additional columns:
- the account ID on the participants page
- the account ID on the judges page
- the team ID on the teams page


viviancan (approved):
Tried locally and works as expected! Nice addition 💯

PR #3926, opened 2023-04-13
Fix join request specs

shaun-technovation:
This will address our join request specs, a lot of the updates were updating the selectors (since the student flow was updated/rebranded), I also did some clean up, and removed some outdated specs.


spec/system/student/mentor_request_to_join_a_team_spec.rb
```diff
@@ -0,0 +1,45 @@
+require "rails_helper"
```
shaun-technovation, line 1: These look like new specs, but they're not, they came from `spec/features/student/mentor_join_request_spec.rb`, I just converted them to system specs (for some unknown reason, I couldn't get them to work/pass as feature specs). I also removed some of the specs where a student views a mentor's profile and approves or rejects a join request, since we don't have this functionality anymore.

viviancan (approved):
Did not try locally, but looks good to me! Thanks for tackling this one, Shaun! 💯

Issue #3923, opened 2023-04-12
Add filestack transformation for team photo image in judging portal


viviancan:

viviancan:
I verified images on judge portal in the learning journey section and the team photo

dboyer:
Thanks for checking this! Confirming I can also see the team photo when judging a submission.

PR #3935, opened 2023-04-17
Created filestack helper method to return dynamic transformation filestack resize url in vue.js

viviancan:
Refs #3923 

This PR has 2 main changes
1. Creates a helper method that returns the filestack transformation resize URL. The resize width is dynamic and can be adjusted.
2. Adds this method to the judging portal for the learning journey images (the resize already existed here, just updated to use new helper method) & added resize method for team photo in the submission overview and team info


shaun-technovation (approved):
Didn't try it locally, but looks good/straight forward to me! 👍 🎉 🚢

Issue #3932, opened 2023-04-14
Potential bug on judging portal with submissions to score "start" button when there are at least 2 scores


viviancan:
Weird behavior on judging portal - when there is a score that needs to be started (submission A) and you assign that judge another submission to score (submission B), the same submission is loaded regardless of which score you "start".

Then once you start the score (submission A) and it moves into the "in progress" table and you try to start the other score (submission B), submission A will continue to load. If you try to start scoring the submission just stays in the submissions to score grid.

https://user-images.githubusercontent.com/29210380/232040914-4e4d586a-5f7f-4d7d-b261-3ff1eb4f9e9b.mp4

viviancan:
@dboyer tagging you on this - issue not displaying in slack!

viviancan:
test test test

viviancan:
In `app/javascript/judge/scores/index.js` a url was crafted using the URL params. `score_id` was found by searching for `score_id` in the URL. The URL had `score` instead of `score_id`. Since `score_id` was `null`, once the request was made there was a piece of logic in `app/technovation/find_eligible_submission_id.rb` that would find the first available submission for that judge, which happened to be the one in progress. That is why the one in progress would always load. The PR updates `score` to `score_id`

dboyer:
Confirmed on QA that I can now start a new assigned submission even if I have a previous submission in progress. Great fix!!

PR #3940, opened 2023-04-19
Fixed bug on "submissions to score" table where incorrect submission was being loaded

viviancan:
Refs #3932 
In `app/javascript/judge/scores/index.js` a url was crafted using the URL params. `score_id` was found by searching for `score_id` in the URL. The URL had `score` instead of `score_id`. Since `score_id` was null, once the request was made there was a piece of logic in `app/technovation/find_eligible_submission_id.rb` that would find the first available submission for that judge, which happened to be the one in progress. That is why the one in progress would always load. 

This PR updates the new score url to `score_id` from `score`


shaun-technovation (approved):
LGTM, great catch on this!! 💯 🎉 🥳 🙌

PR #3944, opened 2023-04-20
Add export option to ChA activities datagrid

shaun-technovation:
This will enable the CSV export option for the ChA activities datagrid.

### DIFF
--- a/app/controllers/chapter_ambassador/activities_controller.rb
+++ b/app/controllers/chapter_ambassador/activities_controller.rb
@@ -33,9 +33,5 @@ module ChapterAmbassador
     def grid_params
       params[:activities_grid] ||= {}
     end
-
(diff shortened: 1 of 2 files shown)


viviancan (approved):
Hey Shaun! When I try to download the csv for the main activity page for a certain ChA there are 195 results. The DB Shows that export job as complete but there is not the little popup to actually download the file. If I try to download by a filtered item for example by accounts it shaves it down to 8 activity results the popup does appear 🤔 maybe its just some weird local stuff going so I think getting it onto QA would be good 💯

shaun-technovation:
> maybe its just some weird local stuff going so I think getting it onto QA would be good 💯

Cool, thanks Vivian! Sounds like a good idea, it's on the way to QA now! 👍

Issue #3914, opened 2023-04-11
Make the Scores & Certificates area available throughout the season


dboyer:
Make the Scores & Certificates tabs/area for students (show below) available throughout the season.

![scores-and-certs.png](https://images.zenhubusercontent.com/605406304f05d34fb2dcbc8c/209b7a3b-55e0-4c64-b391-d8f69bdc7289)

- The "Scores & Certificates Accessible" setting in the admin will still control access to the main content (for the Scores tab and the Certificates tab)
- The survey still has to be completed before accessing scores or certificates

dboyer:
Preference is for 2 where the certificates from past seasons are already available.

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/3914?planning-poker) with Zenhub @shaunxp20 @viviancan

viviancan:
@shaun-technovation @dboyer  I have two questions for this:

1. Should the "my scores" tab always be available in the green header
![CleanShot 2023-04-24 at 13.48.32@2x.png](https://images.zenhubusercontent.com/606b0f4fd13d53dff3d14fa5/70345114-d1e4-4d82-819f-254d6df14972)

2. Not sure if this will be part of a different ticket, but the text needs to be updated depending on the type of submission. For example this submission wont have scores or a cert.
![CleanShot 2023-04-24 at 13.44.14@2x.png](https://images.zenhubusercontent.com/606b0f4fd13d53dff3d14fa5/deacaeb9-6423-4839-81c0-7b3df9c24458)![CleanShot 2023-04-24 at 13.44.55@2x.png](https://images.zenhubusercontent.com/606b0f4fd13d53dff3d14fa5/1a3188fe-64be-4446-bf2d-54d4d791daf9)

dboyer:
My vote is to have the My Scores tab available all the time. 

Like with judging, maybe we need to do an inventory of the different possible states for that screen and then have PT provide suggested text.

shaun-technovation:
> My vote is to have the My Scores tab available all the time.

I made the change for the "My Scores" tab so it will always be visible. And the functionality for the work in this issue and #3916 is on QA now.

viviancan:
Verifying I could access the my scores tab and also see the scores/cert tabs with the correct text on production

PR #3945, opened 2023-04-21
Update student scores and certs functionality

shaun-technovation:
This will make the "score & certs" area for students available at all times, and it will retain our current behavior, which includes:
- the admin setting for "Scores & Certificates Accessible" will control access to them
- the survey will be required to access them too


viviancan (approved):
Tried this locally and it is working as expected for students! I left 2 clarifying questions[ on the issue ](https://github.com/Iridescent-CM/technovation-app/issues/3914). Besides those, looks good to me! 💯

PR #3953, opened 2023-04-25
Standardize IDs on datagrid column headers

shaun-technovation:
Just a very minor change here to make the datagrid columns that I recently added consistent.

Based on some quick research, it looks like ID and IDs is the right way to refer to them:
- https://ux.stackexchange.com/a/91759
- https://english.stackexchange.com/questions/212040/is-the-plural-form-of-id-spelled-ids-or-id


viviancan (approved):
LGTM!

PR #3956, opened 2023-04-27
Add groundwork to display previous certs for students

shaun-technovation:
This will add the basic functionality for students to be able to view their previous certificates. It will probably need some tweaks, but this should get us started at least. 😃


viviancan:
Looks good so far! My only feedback is adding a default message if there are no previous certificates. Besides that, I was able to log in as a student with a cert from 2022 and see it! 
CleanShot 2023-04-28 at 11 07 11@2x

Issue #3952, opened 2023-04-25
Support Ticket


dboyer:
Support requests starting April 26 with priority for judging and RPE related requests.

viviancan:
4/5/23

Teams not displaying for specific ChA even though the students are on teams. [See full context here](https://iridescentlearning.slack.com/archives/C06CHBQ4E/p1680709700859099)

viviancan:
4/27/23
Request for force fail a job by id

dboyer:
4/28/23

Request to unpublish a test submission that made it through to judging - /admin/team_submissions/no-name-yet-nov-29-test-team#/

viviancan:
Going to unpublish the test submission ^^

dboyer:
Three more to be unpublished:

/no-name-yet-unicodemy11
/no-name-yet-qalqilya-jeansafoot-school
/no-name-yet-mariateam

shaun-technovation:
> Three more to be unpublished:
> 
> /no-name-yet-unicodemy11 /no-name-yet-qalqilya-jeansafoot-school /no-name-yet-mariateam

I'll update these 👍

shaun-technovation:
> Three more to be unpublished:
> 
> - /no-name-yet-unicodemy11
> - /no-name-yet-qalqilya-jeansafoot-school
> - /no-name-yet-mariateam

These submissions have been updated. 👍 


```
irb(main):001:0> submission = TeamSubmission.find_by(slug: 'no-name-yet-unicodemy11')                                                                                               
=>                                                                                                                                                                                  
#<TeamSubmission:0x00007fabc54053d0                                                                                                                                                 
...
irb(main):002:0> submission.published_at = nil
=> nil                                                                           
irb(main):005:0> submission.save                                                           
=> true
irb(main):006:0> submission = TeamSubmission.find_by(slug: 'no-name-yet-qalqilya-jeansafoot-school')
=> 
#<TeamSubmission:0x00007fabc8949280                                                                        
...                                                                                                        
irb(main):007:0> submission.published_at = nil                                             
=> nil
irb(main):008:0> submission.save                                                           
=> true
irb(main):009:0> submission = TeamSubmission.find_by(slug: 'no-name-yet-mariateam')        
=> 
#<TeamSubmission:0x00007fabc52e7598                                                                       
...                                                                                                       
irb(main):010:0> submission.published_at = nil                                             
=> nil
irb(main):011:0> submission.save                                                           
=> true
[...]

PR #3957, opened 2023-04-27
Added rake task to force fail job by id

viviancan:
Refs #3952 

We had a request to force fail a specific export job. Our current rake to force fail queued jobs cannot do that. This rake task takes the job id and will force fail that job. There is also a dry mode. One thing to be aware of, the jobs table has an "id" field and a "job_id" field. This rake task uses the "id" field. 

QA Steps
1. Identify the job you want to force fail. Get the id
2. Run rake task locally - ` bundle exec rake 'jobs:force_fail_job_by_id[id_here, dry]' `
3. Verify the output
4. Run rake task in `run` mode
5. Verify the status of that job. It should be "failed"

### DIFF
--- a/lib/tasks/jobs.rake
+++ b/lib/tasks/jobs.rake
@@ -16,4 +16,19 @@ namespace :jobs do
       end
     end
   end
+
+  desc "Force fail job by id"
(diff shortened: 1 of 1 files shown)


shaun-technovation (approved):
👍

PR #3963, opened 2023-04-28
Rewrite admin registration toggle specs

shaun-technovation:
I used the original specs as a template, but I basically rewrote the specs for the admin registration toggles, rough outline of the tests:
"toggle on" -> should see the registration option
"toggle off" -> should not see the registration option
for students, mentors, and judges (new addition)


viviancan (approved):
LGTM! 💯

Issue #3958, opened 2023-04-28
Judge cannot access RPE submissions if started scoring online


dboyer:
Program Team reported an issue with a judge who started online judging but was then assigned to an RPE. When that happened, they did not have the ability to view their RPE submissions. [Details are here](https://iridescentlearning.slack.com/archives/C06CHBQ4E/p1682693216081869). I tried to recreate the issue on QA and ran into a similar but not exactly the same issue.

A judge clicked the button to start scoring online so there is an online submission in their "submissions to score" although they did not enter any data. 

<img width="1118" alt="Screen Shot 2023-04-28 at 12 11 00 PM" src="https://user-images.githubusercontent.com/516764/235199293-3f39a7fc-7fe8-4a02-9352-8d54c0cf2ba6.png">

They were then assigned to an RPE with specific submissions to judge. Those submissions appear in the submissions to score table too.

<img width="792" alt="Screen Shot 2023-04-28 at 12 22 16 PM" src="https://user-images.githubusercontent.com/516764/235201583-0f9807b5-6a33-4b73-9a3d-8e8bd461befb.png">

I clicked Start on "Coffee Defeats Orange Juice" and then clicked back to the submissions tab. Coffee Defeats Orange Juice is now listed twice - as online judging and a pitch event.

<img width="806" alt="Screen Shot 2023-04-28 at 12 24 17 PM" src="https://user-images.githubusercontent.com/516764/235201952-83b755ed-f39f-4e81-89ad-3faea67f948d.png">

If I actually fill out any of the score, the duplicate entry is removed and the score goes into the In Progress section. I was also able to remove myself from a submission.

<img width="829" alt="Screen Shot 2023-04-28 at 12 34 36 PM" src="https://user-images.githubusercontent.com/516764/235204082-7a02ce3f-23fa-41b4-9ff9-1f255b3915c2.png">

When I was looking at the issue Program Team reported, I did see the behavior where clicking on any RPE submission would only load the submission labeled for online judging, not the RPE submission. I did not have that same experience when testing on QA though.

viviancan:
Added #3965 that captures the duplication issue

viviancan:
Unable to recreate this bug locally. I also verified against prod data locally. With the fix for the duplication bug, the submissions are not duplicated in the submissions to score table. Locally, I deleted the one score they already started in the admin portal. I was able to go through the other RPE submissions without any issues. We can continue to monitor, but I am not able to verify if this was due to platform or user error.

[...]

PR #3964, opened 2023-04-28
potential fix for rpe score duplication bug on the frontend

viviancan:
Refs #3958 

Weird bug where an RPE submission score would display twice if the judge clicked "start", then the submission loads, and then the judge returns to the dashboard without actually starting the score. The score would display as a pitch event score and also an online score in the submissions to score table. The submissions to score table scores come from the not started  list which includes not started RPEs and virtual not started scores. 

`not_started`  is returned from `judge/scores_controller.rb` The overall list is created using `not_started_rpe_assigned_submissions + not_started_scores` . Once an RPE judge clicks on "start", a new record is created in the submission_scores table. If they don't actually start anything the updated_at and created_at timestamp stay the same. So then if you reload the page, when the `not_started` list is created and the submission_score is now in 2 categories (RPE and online). This change will remove any "live" (RPE) scores from the `not_started` list which removes the duplication on the front end. There was never an issue with a duplicated `submission_score` record

### DIFF
--- a/app/controllers/judge/scores_controller.rb
+++ b/app/controllers/judge/scores_controller.rb
@@ -146,7 +146,7 @@ module Judge
     end
 
     def not_started_scores
-      current_judge.scores.not_started
+      current_judge.scores.not_started.virtual
         .map do |score|
         {
           id: score.id,


shaun-technovation (approved):
LGTM! Simple fix for such a strange bug! Good find, good fix! 💯 👍 🙌 🎆

Issue #3690, opened 2022-10-27
Team invite to an RPE includes a link that causes a loop


dboyer:
A ChA can invite specific teams to join an RPE on this admin page.

<img width="1017" alt="Screen Shot 2022-10-27 at 4 07 50 PM" src="https://user-images.githubusercontent.com/516764/198387937-ac158399-8d63-48b2-8b1d-3f63b0af74d2.png">

The team members receive an email that includes a button to "log in for more details." If you click this event, you get the error page and stuck in a loop where it is impossible to view any page on QA. 
<img width="607" alt="Screen Shot 2022-10-27 at 4 31 45 PM" src="https://user-images.githubusercontent.com/516764/198391946-3f44c20a-c609-4174-a891-c032c5e8151c.png">

The specific error is here:
<img width="1205" alt="Screen Shot 2022-10-21 at 11 17 34 AM" src="https://user-images.githubusercontent.com/516764/198392668-07fdb641-5261-4ced-bf9a-c3d7ae0bbe44.png">

This card should cover fixing the error so that we can see where the link in the email takes students.

dboyer:
The link in the email may be trying to direct to /student/regional_pitch_events which also causes an error and gets the user stuck in a loop when trying to access it from the button at /student/dashboard#/events

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/3690?planning-poker) with Zenhub @shaunxp20 @viviancan

shaunxp20:
@dboyer Are you by chance impersonating when running into this error?

When I was impersonating, I ran into an error (although it wasn't a polymorphic route error), but when i opened the link (in the invite email) in a fresh window/tab it worked as expected. I tried with three separate teams and it's working as expected. If it's still an issue, I'll need the steps and criteria, maybe I'm doing something different or it's a specific team or team member that's causing an issue? 🤔

dboyer:
I get the error when I follow this series of events:

1. Login as the Deb Chile ChA account (deb+cha1028...)
2. Go to /chapter_ambassador/events#/ and select the Chile Chapter Celebration event. Click the "manage teams" icon and then the add teams button.
3. Select "maze madness" team and then click Done. Then click the "save selected teams" button.

[...]

PR #3968, opened 2023-05-01
Added missing column name configuration for judge recusal counter cache

viviancan:
Refs #3690 

The judge recusal count was incorrect. It appears that all judge recusals per judge were displayed in the count column regardless of the season. This seems to have been an issue since 2021 (actually one of the first things I added! 👀  😢 ). The judge `recusal_scores_count` column was missing the [column_name configuration](recusal_scores_count), which is needed. This change adds the needed configuration. 

QA Steps
1. Pull prod data
2. Check the recusal count on the judges tab for season 2023 - sort by descending 
3. The top few counts are the following
```
83 
82 
74 
18 
* note values may be a little off depending on when you pull prod data
```
4. Run the `fix_counts` rake task
5. Check the judge recusal count for season 2023 - sort by descending 
```
4
3
3
3
```
I also manually checked a few of these judge profiles to verify the number of recusals and it seems to be correct.

### DIFF
--- a/app/models/submission_score.rb
+++ b/app/models/submission_score.rb
@@ -161,6 +161,14 @@ class SubmissionScore < ActiveRecord::Base
     if score.current_season? && score.incomplete? && score.judge_recusal?
       "recusal_scores_count"
     end
+  },
+  column_names: {
+    [
+      "? = ANY (submission_scores.seasons) AND " +
+        "submission_scores.completed_at IS NULL AND " +
+        "submission_scores.judge_recusal = TRUE",
+      Season.current.year.to_s
+    ] => "recusal_scores_count"
   }
 
   scope :complete, -> { where("submission_scores.completed_at IS NOT NULL") }


viviancan:
> I tested the `fix_counts` rake task locally and it seemed to work as expected for me. But that doesn't seem to test the changes in this PR though? 🤔 Anywho, I guess if `fix_counts` is working as expected then if there are issues w/ this PR, that rake task can always be run again. 👍 🙌 :shipit:

I forgot to mention the reason this was needed! If you try to run `fix_counts` without this change you get the following error:
`Must provide :column_names option for relation [:judge_profile] when :column_name is a Proc; you may skip this relation with :skip_unsupported => true`
Then I saw the column_name configuration in the docs and in the other columns where we used the counter cache.

shaun-technovation:
> I forgot to mention the reason this was needed! If you try to run `fix_counts` without this change you get the following error: `Must provide :column_names option for relation [:judge_profile] when :column_name is a Proc; you may skip this relation with :skip_unsupported => true` Then I saw the column_name configuration in the docs and in the other columns where we used the counter cache.

Ah, okay! Thanks for the clarification, makes sense to me! 👍 💯

Issue #3967, opened 2023-05-01
Scores labeled as suspicious have no button to approve the score


dboyer:
Suspicious scores are listed in a table at the top of /admin/scores#/. The score monitors can click the "open score" button by each score to view details. At the bottom of that detail page are two buttons to approve or delete the score. That functionality is working as expected.

However, the main way we monitor scores is by exporting them from /admin/score_exports and looking for "yes" in the suspicious column. Currently scores labeled with a yes in this column do not have the approve button on the score details page, preventing the score monitors from approving the score and getting it off the suspicious list.

Examples include:

/admin/scores/56456#/
/admin/scores/56247#/
/admin/scores/56545#/

Could someone look at the database to confirm if those scores are still listed as suspicious? And if so, why the approve button is not appearing?

viviancan:
@dboyer As far as I can tell, scores listed as just as suspicious for the reason "completed_too_fast" (and I think it is just the first offense) do not show up on the scores page list at the top or have the approve button on the score details. The only scores that show up on the top of the page and also have the approve button are scores that have the reason "Completed too fast (repeat offense)" and "seems too low" and if they have not been manually approved. Scores that are the just "completed_too_fast" are automatically approved (like regular scores). 

It appears that the scores with the following IDs were manually approved. I have included the timestamp. Anything already manually approved will not have the "approve" button on the score details.
56305 - APPROVED AT 2023-04-28 20:57:51.518582
56456 - APPROVED AT 2023-05-01 14:06:48.736423
56508 - APPROVED AT 2023-04-28 20:56:20.0813

The scores export and the "approve" button functionality seem to work as intended. 
The listed examples do not have the approve button because of the following:
56456 - manually approved
56247 - completed too fast (has an automatic approved_at timestamp)
56545 - completed too fast (has an automatic approved_at timestamp)

I made a PR to include the "approved_at" timestamp for the scores export. That is one way to check if it has been manually approved or has the automatic timestamp. If the approved_at timestamp is the same as "created_at" and "updated_at" then it was automatic. If the "approved_at" timestamp is different it is a good indicator that the score has been manually approved.

dboyer:
@viviancan I see the new approved_at timestamp column when I export from QA. However, it is blank for some scores and I don't see a pattern for that. Some scores not labeled as suspicious have an approved time stamp and others don't. Is this correct behavior or should the approved at column always be filled in unless the score is suspicious for something other than completed too fast?

[...]

PR #3969, opened 2023-05-01
Added approved at column to scores grid

viviancan:
Refs #3967 

Adds `approved_at` column to the scores grid. Would score monitoring peeps find this helpful? I did when investigating the suspicious scores bug!


dboyer:
I also agree that this would be helpful!

shaun-technovation (approved):
Looks good/straight forward to me! Good idea, good work! 🙌 🎉 🥳

PR #3972, opened 2023-05-03
Update/fix clicked video tracking

shaun-technovation:
For some reason our existing video tracking isn't working, I have a hunch it's b/c the pop-up modal is intercepting/blocking the click that makes the HTTP request to the back-end to update the video tracking fields in the database.

Anywho, I added some new functionality that doesn't require an extra HTTP request but instead hooks into the "embeds" controller, and looks at what is being embedded (either the pitch or demo/technical video) and updates the tracking field (for the specific submission score) in the database accordingly. I also removed the old video tracking functionality.

### DIFF
--- a/app/controllers/judge/embed_codes_controller.rb
+++ b/app/controllers/judge/embed_codes_controller.rb
@@ -1,5 +1,20 @@
 module Judge
   class EmbedCodesController < JudgeController
     include EmbedCodeController
+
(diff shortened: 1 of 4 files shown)


viviancan (approved):
Tried this locally and both values for demo and technical turned to "true" when I clicked on the link to open the video! 💯 👍

Issue #3996, opened 2023-05-18
Long team names impact layout of judge dashboard


dboyer:
If the team name is long on the judge dashboard, it messes up the alignment of the left column

<img width="1223" alt="Screen Shot 2023-05-18 at 12 38 57 PM" src="https://github.com/Iridescent-CM/technovation-app/assets/516764/90139832-2026-410c-9472-437066a61b72">

shaun-technovation:
I have a fix for long submission (and team) names, making it so it will wrap to the next line.

in screenshot with the misalignment of the left column, this seems to be independent of long submission names, I can see the misalignment with shorter submission names, the left column seems to break when the screen size is a certain size.

shaun-technovation:
> in screenshot with the misalignment of the left column, this seems to be independent of long submission names, I can see the misalignment with shorter submission names, the left column seems to break when the screen size is a certain size.

I fixed this too, so the nav items shouldn't wrap like in the screenshot. 👍

shaun-technovation:
This one is on QA. 👍

PR #4005, opened 2023-05-25
Remove nowrap on judge score table cells

shaun-technovation:
This will remove the "nowrap" option for submission and team name table cells, allowing the content to wrap to the next line instead of pushing the content and table row potentially off the screen.

I tested it by using the browser dev tools to make a submission name longer.

I also updated the judge side nav items so they won't wrap to the next line (like the screenshot in https://github.com/Iridescent-CM/technovation-app/issues/3996#issue-1715943733).

#3996


viviancan (approved):
Did not try locally, but looks pretty straightforward! 💯

Issue #3986, opened 2023-05-10
Set up super admin user group


dboyer:
~~[ ] Create super admin table~~
~~[ ] Seed super admin table with ID of first person~~

viviancan:
We added a `super_admin` boolean on the `admin_profiles` table. This seemed like the most straightforward approach to set up the super admin permissions.

viviancan:
Confirming this is on QA. Currently, all admins are set to regular admins.

viviancan:
2 admin accounts have been updated to be super admin

dboyer:
I see two super admins at /admin/admins!

PR #4006, opened 2023-05-25
Adds admin_type to admin_profiles table to implement "super admin" permissions

viviancan:
This is a potential solution for adding super_admin permissions (#3986)

When I first started working on this, I discovered 3 paths:

1. Add a super_admin_profiles table similar to the admin_profiles table. As I started proving out some of the other pieces of functionality (locking down features, logic to display/hide, etc) some other questions popped up. For example, could a super admin also be a regular admin, if not, then would we need to share some of the logic/functionality that already exists for admin with super_admin.... would we need to pull admin data from 2 separate tables... etc. It seems like a possible route, but I'm not sure if it is the best route.

2. As I worked through the first approach, I discovered a lot of logic for the admin portal that references the `admin_status` which is a field on the accounts table. An admin can have 1 of 3 different statuses (not_admin, temporary_password, full_admin). I tried the idea of adding a 4th status of `super_admin`. It seemed to sort of work, but did not feel right. As I continued to investigate this approach, it seems like `admin_status` is really describing the state of an admin profile, like if they have updated their password. So this led me to the current approach

3. I thought we keep the admin_profiles table and everyone is admin, but we add an `admin_type` (open to naming suggestions here). Admin_type represents the level of permissions. Right now there is just admin and super_admin. So far this seems to be working as expected. I also think that in the future if there were other types of admin, they could be added to the enum. This could also prevent us from having to make additional admin profile tables. 

I could totally be off base, so let me know what you think! I will have this PR and a 2nd PR as a demonstration of locking down the "add a new admin" feature to just super_admins (#3996)

### DIFF
--- a/app/models/admin_profile.rb
+++ b/app/models/admin_profile.rb
@@ -7,6 +7,11 @@ class AdminProfile < ActiveRecord::Base
 
   has_many :exports, as: :owner, dependent: :destroy
 
+  enum admin_type: %w{
+    admin
+    super_admin
+  }
+
   def authenticated?
     true
   end
--- a/db/migrate/20230525160029_add_admin_type_to_admin_profiles.rb
+++ b/db/migrate/20230525160029_add_admin_type_to_admin_profiles.rb
@@ -1,0 +1,5 @@
+class AddAdminTypeToAdminProfiles < ActiveRecord::Migration[6.1]
+  def change
+    add_column :admin_profiles, :admin_type, :integer, default: 0, null: false
+  end
+end
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -191,7 +191,8 @@ CREATE TABLE public.admin_profiles (
     id integer NOT NULL,
     account_id integer NOT NULL,
     created_at timestamp without time zone NOT NULL,
-    updated_at timestamp without time zone NOT NULL
+    updated_at timestamp without time zone NOT NULL,
+    admin_type integer DEFAULT 0 NOT NULL
 );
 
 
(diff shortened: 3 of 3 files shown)


viviancan:
@shaun-technovation I did not see this pop up in dev-robots so tagging you!

viviancan:
test

app/models/admin_profile.rb
```diff
@@ -7,6 +7,11 @@ class AdminProfile < ActiveRecord::Base
 
   has_many :exports, as: :owner, dependent: :destroy
 
+  enum admin_type: %w{
+    admin
+    super_admin
+  }
```
shaun-technovation, line 13: It would be nice to map these to the database values, it would also be nice to use enums in the database too, then it would be really clear. Anywho, definitely not a blocker.

db/migrate/20230525160029_add_admin_type_to_admin_profiles.rb
```diff
@@ -0,0 +1,5 @@
+class AddAdminTypeToAdminProfiles < ActiveRecord::Migration[6.1]
+  def change
+    add_column :admin_profiles, :admin_type, :integer, default: 0, null: false
```
shaun-technovation, line 3: Nice @ setting up a default!

shaun-technovation:
Nice write up!

I thinking keeping it simple is the best approach. I could even see `super_admin` being just a boolean field on the `admin_profiles` table.. Adding `admin_type` makes it slightly murky (I think it already is w/ `admin_status`) regarding how an admin is defined - we have an existing `is_admin?` method, and this will add another `admin?` method, I think there should just be one source of truth for if someone is and admin. Anywho, def not a blocker, this covers the functionality! 👍

viviancan:
Closing this in favor of #4008

PR #4010, opened 2023-05-31
Display pending email confirmation page on judge dashboard

shaun-technovation:
This will display the "Confirm your email" page on the judge dashboard when their email address hasn't been confirmed yet, which is the same behavior as for students.


viviancan (approved):
Did not try locally, but looks good to me! 💯

PR #4011, opened 2023-06-01
Add unpublish submission functionality

shaun-technovation:
This will give super admins the ability to unpublish a submission.


viviancan (approved):
I tried locally and was able to unpublish as a super admin only. Great work! This is going to be so useful next season 💯 🚀

Issue #3999, opened 2023-05-22
Only allow super admins to create and manage admin accounts


dboyer:
Currently, any admin user can invite another user to be an admin. We should restrict this so that only super admin users can create and manage admin accounts. This task covers:

- [x] Only super admins see the "set up a new admin button" at /admin/admins
- [x] Only super admins can see the "delete" button next to the name of an admin

viviancan:
Waiting on confirmation of the approach before I begin the delete functionality

viviancan:
The "create" functionality is on QA. If you login as an admin you should not see the "add an admin" button

dboyer:
- Logged in as a super admin and saw the button to add an admin
- Sent an invite to a new admin and that admin was able to create an account and was listed as an admin
- Logged in as the new admin and did not see the button to add a new admin
- Asked Kiki (an existing admin) to log in and she confirmed that she does not have the button to add an admin

viviancan:
Confirming delete functionality is in on QA and working as expected

dboyer:
Confirming that I can delete an admin when logged in as a super admin but do not have the delete button when logged in as an admin.

PR #4012, opened 2023-06-02
Added admin delete functionality for super admin only 

viviancan:
Refs #3999 

This change allows only super admin to delete other admin.

I encountered some issues with the `Only super admins can delete admins` test. I kept getting an error only on CircleCI, where the delete link cannot be found. I have also had trouble trying to capture a screenshot to figure out what is happening (since locally it is running). I am going to revisit that test on a separate branch.


shaun-technovation (approved):
Looks good/straight forward to me! Good work, nice functionality addition! 🙌 🎉 🥳 :shipit:

PR #4017, opened 2023-06-05
Update mentor types

shaun-technovation:
This PR will:
- Update mentor type "Past Technovation student" to "Technovation alumnae"
- Add mentor type "Postsecondary student"
- Add a new route/endpoint to get mentor types
- Refactor the mentor registration flow to use this endpoint to make the mentor types list, instead of hard coding it


viviancan (approved):
Nice updates here!
I registered as a new mentor and saw the updated mentor types.
I was able to update my mentor type in the mentor profile and I was able to update the mentor type through the admin profile.

Issue #4000, opened 2023-05-23
Add ability for super admins to add new super admins


dboyer:
The initial super admin will be seeded in the database as part of #3986. There are two possible approaches:

1. On the/admin/admins page, super admins should see a button to add a new admin and also a button to "add a new super admin." Clicking this button loads the invite form to send someone an email saying they are invited to be a super admin

2. Next to the delete button for any admin should be a new button stating "change to super admin" so that admins can be given a new super admin role if necessary

There should only be a few super admins so we should go whichever route is the most straightforward. Note: many super admins may be admins already when we deploy this feature.

viviancan:
Confirming this is on QA

dboyer:
Confirming on QA that:

-  I see the button "make super admin" next to existing admins
- The "make super admin" button does not display next to people who are already super admins
- I was able to update an admin to a super admin

viviancan:
Confirming I do not see the "make super admin" button on production with my admin account

dboyer:
And I do see the "make super admin button" on production with my super admin account!

PR #4018, opened 2023-06-06
Added ability for super admin to make other super admin

viviancan:
Refs #4000 

This change will add a "make super admin" button available to super admins. Super admins are the only admin that can make other super admin

CleanShot 2023-06-06 at 09 31 44@2x


shaun-technovation:
Also, just noting, I tested it locally and it was working as expected for me!

PR #4020, opened 2023-06-06
Updated delete admin specs

viviancan:
Refs #3999 

I did not solve the artifact issue, but did solve the issue of the test failing on CircleCI. I ended up using `print page.html` and was able to identify why the link was failing.


shaun-technovation (approved):
Great detective work on this! Good tip about print page.html too!!! 👍 💯 🙌

Issue #4001, opened 2023-05-23
Only allow super admins to edit content settings


dboyer:
Admins should be able to view the content and settings at /admin/season_schedule_settings/ but only super admins should be able to edit the content and settings.

dboyer:
If we choose to only show admins the summary page of the settings, we should go tab by tab to confirm that all information is included on that summary page.

viviancan:
This wasn't initially pointed since I pulled it in at the end of the iteration. I added an estimate of 5 pts.

viviancan:
Confirming this is on QA

dboyer:
On QA, I logged in as a super admin and I was able to change and save content settings. I then logged in as an admin and was able to view the content settings but all the editing options were greyed out. Clicking the "review button" led to the summary page where the only option was to cancel, not save.

dboyer:
Kiki also reviewed and confirmed that she could not edit any content settings with her admin account.

viviancan:
Confirming that I cannot edit the settings with my admin account on prod

dboyer:
And confirming that I see the option to edit content settings as a super admin on prod.

PR #4026, opened 2023-06-13
Makes content and settings only editable by super admin

viviancan:
Refs #4001 

This change will make the content and setting toggles only editable by super admin. Regular admin can click through all of the tabs, but the form inputs are disabled.

It was a little challenging to pass the super admin boolean from rails to the vue.js but found a similar pattern in the judge scores app. I passed the value as a data attribute and then committed it to the vue store so it was accessible in all of the child components. 

Note - the linter auto-updated a lot of the code style in the js tests

### DIFF
--- a/app/javascript/admin/content-settings/components/AdminContentSettings.vue
+++ b/app/javascript/admin/content-settings/components/AdminContentSettings.vue
@@ -131,17 +131,17 @@
         :to="{ name: 'review' }"
       >Review</router-link>
       <button
-        v-else
+        v-else-if="isSuperAdmin"
         ref="submitButton"
         type="submit"
         class="button primary"
         @click.prevent="isProduction() ? confirmSaveSettings() : saveSettings()"
       >Save these settings</button>
-      or
       <a
         ref="cancelButton"
+        class="button secondary"
         :href="cancelButtonUrl"
-      >cancel</a>
+      >Cancel</a>
     </div>
 
     <div ref="formData"></div>
@@ -190,6 +190,7 @@ export default {
     ...mapGetters([
       'judgingEnabled',
       'formData',
+      'isSuperAdmin'
     ]),
 
     currentRoute() {
--- a/app/javascript/admin/content-settings/components/Review.vue
+++ b/app/javascript/admin/content-settings/components/Review.vue
@@ -171,7 +171,10 @@
       </div>
     </div>
 
-    <div class="notice warning">
+    <div
+      v-if="isSuperAdmin"
+      class="notice warning"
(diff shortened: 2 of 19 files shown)


app/javascript/admin/content-settings/components/AdminContentSettings.vue
```diff
@@ -131,17 +131,17 @@
         :to="{ name: 'review' }"
       >Review</router-link>
       <button
-        v-else
+        v-else-if="isSuperAdmin"
```
viviancan, line 134: The submit button is only accessible for super admin

app/javascript/admin/content-settings/components/Review.vue
```diff
@@ -171,7 +171,10 @@
       </div>
     </div>
 
-    <div class="notice warning">
+    <div
+      v-if="isSuperAdmin"
```
viviancan, line 175: The environment notice only displays to super admin

app/views/admin/season_schedule_settings/edit.html.erb
```diff
@@ -2,8 +2,15 @@
 <h3>
   Toggle user features and content throughout the season
 </h3>
+<h5>Note: Only available to super admin</h5>
 
 <div id="admin-content-settings">
+  <div
+    ref="isSuperAdmin"
+    data-super-admin="<%= current_admin.super_admin? %>"
```
viviancan, line 10: Passing the super_admin value to the vue app

app/javascript/admin/content-settings/components/Review.vue
shaun-technovation: Just a random thought, maybe the notice/reminder should be changed to something like "Viewing settings for Production/QA", anywho FWIW. 😄

app/views/admin/season_schedule_settings/edit.html.erb
```diff
@@ -2,8 +2,15 @@
 <h3>
   Toggle user features and content throughout the season
 </h3>
+<h5>Note: Only available to super admin</h5>
```
shaun-technovation, line 5: When I tried this locally I didn't notice this message right away, maybe it needs to stand out more, could also be fine blending in too, just a random thought. 😄

shaun-technovation (approved):
I tried this locally, the minor thing popping out to me was having the Review and Cancel buttons still there for regular admins (more so the Cancel one), anywho, overall looks good! I could only view the settings as an admin! Great work on this super admin stuff! 💯 🥳 🎉 🎆 :shipit:

PR #4027, opened 2023-06-13
Allow mentors to select multiple mentor types

shaun-technovation:
This will allow mentors to select multiple mentor types!

There were quite a few changes that needed to be made to make this happen, @viviancan if you have any questions I'm open for a 1-1 since there's a lot in this PR. 😊 😅


shaun-technovation:
> Tried locally and works as expected!
> 
>     * As an existing mentor I was able to update my mentor type
> 
>     * As a new mentor (went through registration flow) I was able to select mentor types
> 
>     * Mentor types displayed on admin portal and on mentor profile
> 
>     * As an admin I was able to update a mentor's email address
> 
> 
> I did not test converting a student to mentor (I actually don't know how to do this 😅 ) but can check once on QA
> 
> We may want verification from PT if there are any other places mentor types display! Great work on this, Shaun! Especially fixing some of those super old bugs 💯 🚀

Thanks so much for looking at this @vivan! 🙌 🎉 👏 

RE: converting students to mentors, I think that option is only available to students who are older than 18, anywho, i I tested it and it worked as expected, and of course it will get tested more once it lands on QA, so I think we're good for now! 😄

PR #4028, opened 2023-06-13
Rebrand initial submission honor code for mentors and students

viviancan:
Refs #4022 

This change rebrands the initial student submission honor code. The rebranded honor code will also display for mentors if they are the participant that starts the submission.

CleanShot 2023-06-13 at 14 54 42

### DIFF
--- a/app/views/team_submissions/_honor_code.en.html.erb
+++ b/app/views/team_submissions/_honor_code.en.html.erb
@@ -1,4 +1,4 @@
-<p>
+<p class="margin--b-large">
   Technovation Girls’ mission is to educate and empower 
   <%= link_to "young women", "https://www.technovation.org/diversity-equity-inclusion-statement/" %>
   to make a difference in their community using technology. In order to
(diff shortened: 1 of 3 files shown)


app/views/team_submissions/_honor_code.en.html.erb
```diff
@@ -1,4 +1,4 @@
-<p>
+<p class="margin--b-large">
```
viviancan, line 1: I opted to go with original classes since this honor code partial is not rebranded. I also tried the route of creating a 2nd honor code partial that was the "tailwind" version but it seemed like unnecessary duplication and something that we would need to upkeep x2 for future seasons - especially since the only used style was bottom margin. As we complete the rebrand process for the mentor experience these classes could be swapped for tailwind classes. Open to other suggestions.

shaun-technovation (approved):
I didn't try it locally, but looks good/straight forward to me! 👍 🥳 :shipit:

Issue #4021, opened 2023-06-07
Integrate Filestack for team photo upload in submission process


dboyer:
Depends on completing #3222 

- Integrate Filestack for the team photo upload - likely can be similar to the code on the screenshots in the submission process
- Confirm that any uploaded photo is displayed on the team summary page on the user profile
- If user has already uploaded a team photo on their team summary page on the user profile, the photo should display on the team photo section of the submission process

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/4021?planning-poker) with Zenhub @shaun-technovation @viviancan

viviancan:
Confirming this is on QA

Issue #3222, opened 2022-01-11
Upload team photo as part of submission process


rsgonzal:
Clicking team photo from the submission page should take you directly to the summary section of "My Team" 

after saving your photo, it should bring you back to that same page

![Screen Shot 2022-01-10 at 5.13.41 PM.png](https://images.zenhubusercontent.com/60527169cde2772242d0de95/bd3d3a3f-e5a1-4ba8-889f-7aee3bd87a8a)

dboyer:
We decided to integrate the team photo upload into the submission process rather than move the user between submissions and their dashboard to upload the photo. 

We will need to add a step in the submission process for uploading the photo. There is currently no section for "Start" which includes the honor code and team photo.

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/3222?planning-poker) with Zenhub @shaun-technovation @viviancan

dboyer:
Edit: We have discovered this page exists. This ticket now covers having students go directly to the "start" section of the submission process.

viviancan:
Confirming this is on QA

dboyer:
On QA, I tested:

- Changed the team photo on an existing complete submission. The photo updated in the submission, on the project page, and on the team summary page on the dashboard
- Started a new submission. I was taken to the honor code and then to the overview page for the start section of the submission. I was able to select, edit, and upload an image using Filestack. The team photo part was checked off as complete.
- I went back to the profile and the uploaded photo was showing on the team summary photo page.
- Logged in as a new student and created a team. Uploaded a photo on the team summary page on my profile. Started a submission and my team photo was already there. I was able to update the team photo.

PR #4032, opened 2023-06-14
Added team photo page to the student submission process

viviancan:
Refs #3222 

When a student clicks on the "team photo" link in the submission side menu bar a new team photo page will display. In #4021 filestack will be integrated on this page to upload the team photo. Also after the honor code is signed, students will be  redirected to the "Start" section.

### DIFF
--- a/spec/features/student/submission_piece_spec.rb
+++ b/spec/features/student/submission_piece_spec.rb
@@ -39,12 +39,11 @@ RSpec.feature "Students edit submission pieces" do
     end
   end
 
-  # scenario "Set the team photo" do
-  #   click_link "Team photo"
-  #   click_link "Summary"
-  #   expect(page).to have_selector("#filestack-team-photo-form", visible: false)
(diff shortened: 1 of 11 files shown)


spec/features/student/submission_piece_spec.rb
```diff
@@ -39,12 +39,12 @@
...
-    click_link "Summary"
-    expect(page).to have_selector("#filestack-team-photo-form", visible: false)
-    expect(page).to have_button "Change Image"
-  end
+  # scenario "Set the team photo" do
+  #   click_link "Team photo"
+  #   click_link "Summary"
+  #   expect(page).to have_selector("#filestack-team-photo-form", visible: false)
+  #   expect(page).to have_button "Change Image"
+  # end
```
shaun-technovation, line 47: Wondering if this is supposed to be commented out? If it is, I think using the `xscenario` style would be better so that way when the test suite is run it will be marked as a pending test and we'd know we'd have to address. 😄

PR #4033, opened 2023-06-15
Add missing migration for mentor types work

shaun-technovation:
I forgot to add this migration which is to remove `mentor_type` from the `signup_attempts` table.


viviancan:
LGTM! Did not try locally but seems straightforward!

viviancan (approved):

PR #4034, opened 2023-06-15
Remove mentor/ChA merge functionality

shaun-technovation:
This will remove the mentor/ChA merge functionality, it's really old, isn't being used and is currently broken! Anywho, it'll be here in case we want to restore/revisit it though!


viviancan (approved):
Nice cleanup! 👍

PR #4035, opened 2023-06-16
Update team submission friendly id

shaun-technovation:
Brief overview of the changes in this PR:
- Making it so that if the submission name or team name changes, it will regenerate the slug
- Restoring the default slug name, so it will be "app-name-by-team-name"
- Updating the FriendlyId gem
  - [Changelog](https://github.com/norman/friendly_id/blob/master/Changelog.md)

### DIFF
--- a/app/models/team.rb
+++ b/app/models/team.rb
@@ -10,6 +10,8 @@ class Team < ActiveRecord::Base
 
   include PublicActivity::Common
 
+  after_commit :regenerate_team_submission_friendly_id
+
   after_commit -> {
     return false if destroyed?
 
@@ -410,6 +412,12 @@ class Team < ActiveRecord::Base
   end
 
   private
+
+  def regenerate_team_submission_friendly_id
+    if saved_change_to_name? && !submission.is_a?(NullTeamSubmission)
+      submission.update(slug: nil)
+    end
+  end
 
   def default_team_photo_url
     if Season.current.year >= 2023 && current?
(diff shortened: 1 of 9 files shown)


app/models/team.rb
```diff
@@ -411,6 +413,12 @@ def assigned_judge_names
 
   private
 
+  def regenerate_team_submission_friendly_id
+    if saved_change_to_name? && !submission.is_a?(NullTeamSubmission)
+      submission.update(slug: nil)
```
shaun-technovation, line 418: I randomly stumbled upon this blurb that says if you set the `slug` to nil, it will regenerate the slug, so that's basically what this functionality is doing - if the team name changes, then regenerate the slug.

Here's the issue I found this blurb in:
https://github.com/norman/friendly_id/issues/968

app/models/team_submission.rb
```diff
@@ -641,9 +642,11 @@ def standardize_url(url)
   end
 
   def team_name_and_app_name
-    [
-      [:app_name, :team_name, :slug]
-    ]
+    "#{app_name} by #{team_name}"
+  end
```
shaun-technovation, line 646: Just for context, this functionality was changed in 2021:
https://github.com/Iridescent-CM/technovation-app/commit/70ba70bbee6b5dbd0e45e56d6de3a8890c1206d9#diff-17e1127ee3fe2922174d5e3016dd1889382ade65685ec684af0d8174e8233bb3L566-R568

app/models/team.rb
viviancan: 👍 Nice find!

viviancan (approved):
💯 Nice Fix!!

- I edited a preexisting project and the slug was regenerated as expected
- As a student I created a new submission. When I set the project name the slug updated as expected

PR #4051, opened 2023-06-23
Add ability for ChAs to delete a team invite

shaun-technovation:
This will allow ChAs to delete team invites.


viviancan (approved):
Tested locally and was able to delete pending invites as a ChA! Great work with this 💯

PR #4052, opened 2023-06-23
LearnWorlds tests

shaun-technovation:
This will add some very basic/general test coverage for the LearnWorlds functionality.

There are definitely more specific tests that could be written. These alone were quite the slog to write, maybe if more LearnWorlds changes need to happen in the future then more tests could be added at that point, but this is a start at least!


viviancan (approved):
💯 LGTM! Nice work here!

Issue #3771, opened 2023-01-05
Remove requirement that parent email be different than student email for all divisions


shaun-technovation:
If a beginner student returns as a junior student, they will be unable to get agree to the data terms because as a junior student their email address can't match their parent/guardian's email address (which is allowed in the beginners division), this validation check fails and causes the "Uh oh, something went wrong" page.

dboyer:
Note: we have three examples of this issue reported by a ChA from India. I confirmed on QA (account with the email deb+parent0811@). I was able to log into my account and was taken to the data terms page. When I selected the box agreeing to the terms and clicked the submit button, the data terms page just reloaded again.

<img width="1435" alt="Screen Shot 2023-01-06 at 4 37 56 PM" src="https://user-images.githubusercontent.com/516764/211104362-2d414b0d-a571-418e-ad18-86348c84bc32.png">

dboyer:
Possible solutions:

1. Allow junior division user emails to also match their parent emails
2. Add logic allowing only students who move from beginner division to junior division in a year to have user emails that match parent emails (although this just kicks the can down the road a year)
3. Add a page that if the validation fails for this reason, we take the student to a page where they have to update either their user email or their parents email so they don't match

dboyer:
Program team agrees that we should remove the requirement that the parent email be different than the user email. This can apply to all divisions.

dboyer:
Validation is enforced in the junior/senior registration workflow and on the parent email page of the junior/senior dashboard.

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/3771?planning-poker) with Zenhub @shaun-technovation @viviancan

dboyer:
Tested on QA:

- Registered a new junior account and used the same address for parent email and student email during registration. Did not receive any error message when trying to register. I received both the welcome email that goes to students and the parent consent email. After signing the forms, I also received both the student notification that my parents had signed the permission forms and the parent notification with copies of the signed consent forms. 

- Registered a new senior account and used the same address for parent email and student email during registration. Did not receive any error message when trying to register. I received both the welcome email that goes to students and the parent consent email. After signing the forms, I also received both the student notification that my parents had signed the permission forms and the parent notification with copies of the signed consent forms. 

[...]

PR #4058, opened 2023-06-26
Remove different email requirement for parent/guardian and student

viviancan:
Refs #3771 

This PR removes the validation that required student emails to differ from their parent/guardian email. Students in all divisions can now have the same account email as their parent/guardian email.


shaun-technovation (approved):
Didn't try it locally, but the changes seems straight forward to me! 👏 🎉 🥳 🚀

PR #4060, opened 2023-06-27
Only show delete invite button for pending invites

shaun-technovation:
This will only show the delete invite button if the invite is still in a "pending" status. I also added tests to make sure the delete button doesn't display for accepted or declined invites.


viviancan (approved):
Nice improvement and tests! 💯

Issue #4056, opened 2023-06-26
Bug: No error message when trying to upload too large of an image on team photo page


dboyer:
While testing #3222, I found that if I upload an image over 2mb on the team summary page (i.e. /student/teams/486), it will fail due to image size but not give me an error message. Instead it just reloads the screen with the prompt that I should upload an image.

I confirmed that I do see the error message when uploading a large image in the submission pages. I was able to upload an image less than 2mb on the team page without an issue.

<img width="902" alt="Screen Shot 2023-06-26 at 12 10 51 PM" src="https://github.com/Iridescent-CM/technovation-app/assets/516764/ca3c7945-3d97-4954-8b12-d71b5ab20db9">

viviancan:
I was unable to recreate the bug. I believe the error message is always displayed when uploading an image larger than 2MB, but depending on where you are on the screen you may not see it. If you scroll to the bottom of the page and then upload an image through the picker, you cannot see the error message since it is at the top of the page. Since this is filestack default styling, I don't think there is a way to lower the error message. I did add custom text to the picker adding the 2MB limit. I am currently reviewing documentation to determine if we can change the location of the error message or add it to the picker. 

![Screen Shot 2023-06-27 at 4.05.29 PM.png](https://images.zenhubusercontent.com/606b0f4fd13d53dff3d14fa5/63b75c04-47c9-4d9d-ba0b-991e6a4faa40)

dboyer:
Confirming that I see the updated text.

PR #4061, opened 2023-06-27
Added custom message in filestack picker

viviancan:
Refs #4056 

This PR adds custom text to the filestack picker. I was unable to recreate the bug. The error message is always displayed when uploading an image larger than 2MB. If you scroll to the bottom of the page and then upload an image through the picker, you cannot see the error message since it is at the top of the page. Since this is filestack default styling, I don't think there is a way to lower the error message. 

CleanShot 2023-06-27 at 16 06 07@2x


shaun-technovation (approved):
Didn't try it locally, but the change looks simple enough! 👍 💯 :shipit:

PR #4062, opened 2023-06-27
Update rake task that generates certs in batches

shaun-technovation:
I ran into an error running this rake task:
```
ArgumentError: comparison of String with 0 failed
```

I found that `batch_size` needs to be converted to an integer for it to work.


viviancan (approved):
LGTM!

PR #4066, opened 2023-06-28
Remove lingering student/parent email validation

shaun-technovation:
This validation was causing the issue w/ junior and senior students not being able to create a new team.


viviancan (approved):
I tried locally and created a team as a junior with the same email address as my parent/guardian. I tired a few other scenarios too! It would be great to get extra eyes on this! 💯

Issue #4030, opened 2023-06-14
Add "mentor types" field on admin participants page


shaun-technovation:
The participants page in the admin needs to be updated to allow admins to select multiple mentor types and view the corresponding mentors of those types.

**Background info:**
Mentors are now allowed to select multiple mentor types (#4013). From this work, the mentor type selection on the participants page in the admin was temporarily removed.

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/4030?planning-poker) with Zenhub @shaun-technovation @viviancan

dboyer:
Confirmed with VET that this should be an OR type search. Adding multiple mentor types to a search should return results that meet any of the search terms NOT only those that meet all search terms.

dboyer:
Confirmed on QA that I can add mentor types as an additional column. Added several different types of mentors and confirmed the search returned those that matched any of the selected search terms.

dboyer:
@viviancan One potential wrinkle based on testing from VET. Rae's testing found:

1. Search for two mentor types (i.e. postsecondary student and alumna)
2. Results return people who have either (or both) of those mentor types. In the mentor type column, it shows postsecondary student, alumna, or postsecondary student and alumna
3. In the export of search results, the mentor type column includes postsecondary student, alumna, or postsecondary student and alumna

The issue is that if a mentor is a postsecondary student and/or alumna but also has another mentor type (i.e educator in the case of /admin/participants/1087#/), that educator mentor type does not appear anywhere in the search result grid or the export. VET would thus have inaccurate information on the full list of mentor types for that user.

Ideally, all mentor types for the mentor would be in the mentor type column and exported. So search by the selected mentor types but display all mentor types for the mentor.

viviancan:
Moving this back into code review! [This PR](https://github.com/Iridescent-CM/technovation-app/pull/4093) addresses the issue

viviancan:
Confirming the update is on QA

dboyer:
Confirming that I also now see all mentor types for an individual (not just the ones searched) and all types are included in the export.

PR #4075, opened 2023-06-29
Adds mentor types filter to the participant admin grid

viviancan:
Refs #4030 

This PR will add a filter select option to the participant grid in the admin portal. It also fixes the mentor type display column

CleanShot 2023-06-29 at 17 23 29@2x

### DIFF
--- a/app/data_grids/accounts_grid.rb
+++ b/app/data_grids/accounts_grid.rb
@@ -19,9 +19,9 @@ class AccountsGrid
   column :last_name, mandatory: true
   column :email, mandatory: true
 
-  column :mentor_type do
+  column :mentor_types do
     if mentor_profile.present?
-      mentor_profile.type_names.join(", ")
+      mentor_profile.mentor_types.pluck(:name).join(",")
     else
       "-"
     end
@@ -514,7 +514,7 @@ class AccountsGrid
   filter :mentor_types,
     :enum,
     header: "Mentor Type",
-    select: proc { MentorType.all.map { |c| [c.name, c.id] } },
+    select: proc { MentorType.all.map { |m| [m.name, m.id] } },
     filter_group: "more-specific",
     html: {
       class: "and-or-field"
--- a/app/data_grids/accounts_grid.rb
+++ b/app/data_grids/accounts_grid.rb
@@ -21,7 +21,7 @@ class AccountsGrid
 
   column :mentor_type do
     if mentor_profile.present?
-      mentor_profile.mentor_type
+      mentor_profile.type_names.join(", ")
     else
       "-"
     end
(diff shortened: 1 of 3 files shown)


app/data_grids/accounts_grid.rb
```diff
@@ -524,6 +511,19 @@ class AccountsGrid
       .where(clauses.join(' OR '))
     end
 
+  filter :mentor_types,
+    :enum,
+    header: "Mentor Type",
+    select: proc { MentorType.all.map { |c| [c.name, c.id] } },
```
viviancan, line 517: I wasn't sure if this was necesssary, but found reference in the [documentation here](https://github.com/bogdan/datagrid/wiki/Filters#enumerable)

> IMPORTANT: Always wrap dynamic :select option with proc, so that datagrid fetch it from database each time when it renders the form.

app/data_grids/accounts_grid.rb
```diff
@@ -21,7 +21,7 @@ class AccountsGrid
 
   column :mentor_type do
```
shaun-technovation, line 22: I think this column name should change to `:mentor_types` (which I thought I did in the original work, but I guess not! 😊 😅). 😄

app/data_grids/accounts_grid.rb
```diff
@@ -21,7 +21,7 @@ class AccountsGrid
...
     if mentor_profile.present?
-      mentor_profile.mentor_type
+      mentor_profile.type_names.join(", ")
```
shaun-technovation, line 24: This can also be done inline w/ something like:
 `mentor_profile.mentor_types.pluck(:name).join(",")`, or if it's a method call, just imho, it might be clearer as `mentor_type_names` vs `type_names`. Anywho, not a blocker, just random feedback. 😄

app/data_grids/accounts_grid.rb
shaun-technovation: Excellent! Good find! Just wondering, what does `c` stand for here?

shaun-technovation (approved):
Awesome! LGTM! I tried it locally and it works as expected! Great work! Simple changes too! I don't think it would have been as easy for me to figure this one out! Nice work! 👍 💯 🥳 :shipit:

app/data_grids/accounts_grid.rb
viviancan: ahh I used the example in the documentation (it was category lol). I just updated it!

app/data_grids/accounts_grid.rb
viviancan: 👍

app/data_grids/accounts_grid.rb
viviancan: I ended up updating this inline! Thanks for the feedback!

PR #4076, opened 2023-07-03
Update admin impersonation specs

shaun-technovation:
I removed the ChA impersonation specs, these specs were commented out back in 2017!
https://github.com/Iridescent-CM/technovation-app/commit/35efaf398fa0090099bcc43f5cbc40cee16bbfd3

I'm guessing ChAs used to be able to impersonate, but then at some point that functionality got removed. Anywho, the specs aren't necessary and we can always add them back if this functionality comes back.

I also did a little clean up for the admin impersonation specs and renamed this file.


viviancan (approved):
lgtm! 💯 nice cleanup

Issue #4078, opened 2023-07-03
Update Nokogiri gem


shaun-technovation:

shaun-technovation:
We currently have 100 security vulnerabilities! 😱

I found this Nokogiri upgrade which will address 12 of them! Nokogiri follows SemVer, so making a minor upgrade from 1.11.7 to 1.14.5 shouldn't have any breaking changes.

Here are the related security vulnerabilities:
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/154
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/153
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/77
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/76
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/53
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/51
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/50
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/38
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/24
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/145
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/93
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/52

PR #4079, opened 2023-07-03
Update Nokogiri gem

shaun-technovation:
This will update Nokogiri (and its dependencies) to address twelve security vulnerabilities! 😱

#4078 has a few more details.


viviancan (approved):
I tried a few things locally! I think it would be great to get another set of 👀 

Great work on these updates!! 💯

PR #4085, opened 2023-07-06
Update Thunkable URL format validation

shaun-technovation:
Pretty simple change here, just changing `products` to `projectPage` in the Thunkable URLs.


viviancan (approved):
Did not try locally, but looks pretty straight forward! Nice update 👍

Issue #4087, opened 2023-07-06
Update Rack gem


shaun-technovation:
Upgrade Rack from `2.2.4` to `2.2.7`.

Related security vulnerabilities:
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/140
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/122
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/141
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/124
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/123

PR #4088, opened 2023-07-06
Update Rack gem

shaun-technovation:
This will update the Rack gem from 2.2.4 to 2.2.7 to address the security vulnerabilities listed in #4087.


viviancan (approved):
Did not try locally, but seems like a smaller version change 💯

PR #4090, opened 2023-07-07
Update Luxon package

shaun-technovation:
This will address the following security vulnerability:
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/120

Luxon is used to during registration to format the division cutoff date, and for the age validation.


viviancan (approved):
Did not try locally, but seems like a small upgrade!

PR #4093, opened 2023-07-10
Updated mentor type logic to display all of mentor's mentor types not just the types that are filtered

viviancan:
Refs #4030 

Updated mentor type logic to display all of the mentor's mentor types, not just the types that are filtered. This is based on feedback from VET


shaun-technovation (approved):
LGTM! Nice fix! 👍 💯 🥳

Issue #4095, opened 2023-07-11
Update Rails gem


shaun-technovation:
Update Rails from `6.1.6.1` to `6.1.7.4`.

This should address the following security vulnerabilities:
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/155
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/139
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/137
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/155
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/139
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/137
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/142
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/138
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/144
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/121
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/125
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/149

Relevant release notes:
- https://github.com/rails/rails/releases/tag/v6.1.7.4
- https://github.com/rails/rails/releases/tag/v6.1.7.3
- https://github.com/rails/rails/releases/tag/v6.1.7.2
- https://github.com/rails/rails/releases/tag/v6.1.7.1
- https://github.com/rails/rails/releases/tag/v6.1.7

shaun-technovation:
I think this is a relatively safe upgrade, since Rails 6 (6.0 and 6.1) is only receiving security (6.1) and severe security (6.0 and 6.1) updates:
https://guides.rubyonrails.org/maintenance_policy.html#security-issues

PR #4096, opened 2023-07-11
Update Rails gem

shaun-technovation:
This will bump Rails from version 6.1.6.1 to version 6.1.7.4.

The related security issues it address are in #4095.

It also should be a pretty safe upgrade since Rails 6.x is basically only receiving security updates:
https://guides.rubyonrails.org/maintenance_policy.html#security-issues

But since it's a bigger upgrade, I won't merge it into QA until after we get a deploy out.


Gemfile
```diff
@@ -1,9 +1,7 @@
 source "https://rubygems.org"
 ruby "3.1.2"
 
-gem "rails", "~> 6.1.0"
-gem "rack", "~> 2.2.0"
-gem "rails-html-sanitizer", "~> 1.4.0"
```
shaun-technovation, line 6: The `rack` and `rails-html-sanitizer` updates shouldn't be necessary as standalone updates now, since if I'm looking at the Rails 6.1.7.4 dependencies correctly, these higher versions are already required by Rails:
https://rubygems.org/gems/rails/versions/6.1.7.4/dependencies

viviancan (approved):
Tried this locally and it seems to be working as expected! Would be great to get more 👀 once on QA!

PR #4101, opened 2023-07-12
Update PDFKit gem

shaun-technovation:
PDFKit is used for the parental consent form that students can download in PDF format.

Should be a pretty safe upgrade, nothing really popping out to me in the changelog:
- https://github.com/pdfkit/pdfkit/releases

This will resolve the following (critical) security vulnerability:
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/96


viviancan (approved):
Did not try locally, but seems like a small change 👍

PR #4104, opened 2023-07-14
Update loader-utils module

shaun-technovation:
This will update `loader-utils`, which has six security vulnerabilities (two that are critical), this **should** address:
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/98
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/97
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/103
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/102
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/100
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/99

I say **should** update, because... we have two version of `loader-utils` a `1.x.x` version and a `2.x.x` version, Dependabot says to update to `2.0.4` which I did, and I also updated our `1.x.x` version to `1.4.2` because Dependabot **also** says that this vulnerability is fixed in `1.4.2` (despite it saying to update to version `2.0.4`), this can be seen in one of the Dependabot issues:
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/98

To address this, first, I tried removing `yarn.lock` and the `node_modules` directory, and reinstalling the modules and rebuilding `yarn.lock` in order to update `loader-utils`, but kept running into issues because most of our modules haven't been updated in years, which means a lot of breaking changes.

I ended up updating `loader-utils` manually by updating `yarn.lock`, probably not ideal, but in this case it seems simple enough and I ran across a couple of posts suggesting it:
https://stackoverflow.com/a/58142690
https://dev.to/ayc0/yarn-lock-how-to-update-it-1fa2

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -6858,18 +6858,18 @@ loader-runner@^2.4.0:
   integrity sha512-Jsmr89RcXGIwivFY21FcRrisYZfvLMTWx5kOLc+JTxtpBOG6xML0vzbc6SEQG2FO9/4Fc3wW4LVcB5DmGflaRw==
 
 loader-utils@^1.0.2, loader-utils@^1.1.0, loader-utils@^1.2.3, loader-utils@^1.4.0:
-  version "1.4.0"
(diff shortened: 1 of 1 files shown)


viviancan (approved):
LGTM 👍

PR #4116, opened 2023-07-21
Update Geocoder gem

shaun-technovation:
This will update Geocoder from version `1.4.9` to version `1.6.7`, which will address the following security vulnerability:
https://github.com/Iridescent-CM/technovation-app/security/dependabot/2

Version [`1.5.0`](https://github.com/alexreisner/geocoder/blob/master/CHANGELOG.md#150-2018-jul-31) had two changes I needed to address:
> - Change default street address lookup from :google to :nominatim.
> - Test lookup fixtures should now return coordinates and NOT latitude/longitude attributes (see `#1258`). This may break some people's tests.


config/initializers/geocoder.rb
```diff
@@ -1,4 +1,6 @@
 Geocoder.configure({
+  lookup: :google,
+
```
shaun-technovation, line 3: This simple line sets the default provider to Google, without this line it would use Geocoder's default provider which is Nominatim as of version `1.5.0`.

spec/geocoder_helper.rb
```diff
@@ -8,24 +8,22 @@
...
       {
-        'latitude'     => 34.052363,
-        'longitude'    => -118.256551,
-        'address'      => 'Los Angeles, CA, USA',
-        'state'        => 'California',
-        'city'         => 'Los Angeles',
-        'state_code'   => 'CA',
-        'country'      => 'United States',
-        'country_code' => 'US',
+        "coordinates" => [34.052363, -118.256551],
```
shaun-technovation, line 11: The pertinent changes in this file are removing `latitude` and `longitude` and replacing it w/ `coordinates`. The other changes were to just address all the formatting issues.

viviancan (approved):
Tried setting/changing a few locations locally and seems to be working as expected! It would be great to get more 👀 on this!

viviancan:
Woo! Awesome win with this update 💯 WTG Shaun 👍

PR #4123, opened 2023-07-26
Update sidekiq gem

shaun-technovation:
This will update sidekiq from version `6.2.1` to `6.4.2` to address the following security vulnerability:
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/61

Nothing popped out to me in the sidekiq changelog as far as breaking changes:
- https://github.com/sidekiq/sidekiq/blob/main/Changes.md


viviancan (approved):
Did not try locally. I agree that the changes don't seem too extreme! Nice upgrade 👍

PR #4130, opened 2023-07-31
Add ability for students to upload parental consent forms

shaun-technovation:
This will add functionality that will allow a student upload their parental consent form.

Overview of the work in this PR:
- Database changes
- New route and controller action
- New CarrierWave uploader
- New view with a form, and some styles for it
- Removing old/unused parental consent pages


shaun-technovation:
A random comment, open to naming changes, for example, on the back-end it's called the `uploaded_consent_form` and and it's also called `paper_consent_uploads` in places. 🤔 😕

app/views/student/paper_consent_uploads/_form.html.erb
```diff
@@ -0,0 +1,45 @@
...
+  <li>
+    <%= link_to "Parental Consent Form",
+      student_downloadable_parental_consent_path(format: :pdf),
+      target: :_blank,
+      class: "tw-link"
+    %>
+  </li>
+
+  <li>
+    <%= link_to "Media Consent Form",
```
viviancan, line 18: Should the media consent form be removed/commented out since the uploader is only for parental consent right now? Just a thought! 🤔

app/views/student/paper_consent_uploads/_form.html.erb
```diff
@@ -0,0 +1,45 @@
...
+  Please contact
+  <%= mail_to ENV.fetch("HELP_EMAIL"),
+    "support@technovation.org",
+    subject: "Parental consent signed form",
+    class:"tw-link"
+  %>
+  with any questions.
+</p>
+
+<%= form_for @parental_consent, url: student_paper_consent_upload_path, html: { multipart: true } do |f| %>
```
viviancan, line 40: I think it would be helpful to have a max file size!

viviancan (approved):
Nice addition and cleanup! I left a few comments, no blockers or anything like that! Woohoo, this is such a big win 👍

app/views/student/paper_consent_uploads/_form.html.erb
```diff
@@ -0,0 +1,45 @@
+<h3 class="mb-4">Uploading Consent Forms</h3>
```
viviancan, line 1: I also think it would be nice to include some sort of status here if they manually uploaded the file - like uploaded at or status is pending, something like that! 

CleanShot 2023-08-03 at 14 54 24@2x

shaun-technovation: Yup, I agree, I think I mentioned this on standup too. Anywho, I'll work on adding something after some of the other functionality gets ironed out. 👍

app/views/student/paper_consent_uploads/_form.html.erb
shaun-technovation: Good call out, I'll see about adding a reasonable max size, right now it's using the default we have set, which is 500MB!

Issue #3777, opened 2023-01-09
Rebrand page where returning students agree to data use terms


dboyer:
When a student who competed in a previous years logs in, they are taken to a page (/terms_agreement/edit#/) where they need to agree to the terms of use for the new year. That page is on the old branding and should be changed to the new branding.

<img width="1431" alt="Screen Shot 2023-01-09 at 9 56 24 AM" src="https://user-images.githubusercontent.com/516764/211337658-5812dad2-18b9-4304-aa86-21430aff42c1.png">

dboyer:
Confirming that I see the rebranded data terms when logging into QA as a returning participant.

PR #4136, opened 2023-08-03
Rebrand Data Terms interrupt screen for returning participants

viviancan:
Refs #3777 

This change rebrands the "interrupt" screen for returning participants. I experimented with two approaches: 
1. Creating the vue rebrand version of this component 
2. Reworking this functionality in Rails + rebrand

After speaking through it yesterday in Retro, I feel more comfortable with this approach, given how soon the season is opening. While I have a semi-working version (still struggling with passing tests) with approach 2, it feels too rushed. I prefer to keep that on its branch and use it as a dev improvement task or revisit it for later in the season. 

In this PR, I moved the energetic container and thick rule component to the application component directory since they are now used in multiple places. This also required me to update a few of the import paths in the judging experience. I tested the judging portal and registration flow locally - they look/function as expected. 

Screen Shot 2023-08-03 at 10 57 51 AM


shaun-technovation (approved):
Didn't try it locally, but LGTM and I'm sure anything off visually will be discovered on QA. 👍 🙌 :shipit: 🚀

PR #4141, opened 2023-08-08
Allow admins to reject uploaded parental consents

shaun-technovation:
This will add basic functionality for an admin to reject an uploaded parental consent form.

Basic overview of changes:
- New route/controller action
- New reject button/action on the datagrid
- New student mailer (to notify them about rejected parental consent)

### DIFF
--- a/app/controllers/admin/paper_parental_consents_controller.rb
+++ b/app/controllers/admin/paper_parental_consents_controller.rb
@@ -31,6 +31,18 @@ module Admin
         success: "You approved the parental consent for #{parental_consent.student_profile_full_name}."
     end
 
+    def reject
+      parental_consent = ParentalConsent.find(params[:paper_parental_consent_id])
+
+      parental_consent.update(
+        upload_rejected_at: Time.now,
+        upload_approval_status: ParentalConsent::PAPER_CONSENT_UPLOAD_STATUSES[:rejected]
+      )
+
+      redirect_to admin_paper_parental_consents_path,
+        success: "You rejected the parental consent for #{parental_consent.student_profile_full_name}."
+    end
+
     private
 
     def grid_params
--- a/config/locales/mailers.en.yml
+++ b/config/locales/mailers.en.yml
@@ -39,6 +39,10 @@ en:
     welcome_returning_judge:
       subject: "Welcome to Technovation %{season_year}!"
 
+  student_mailer:
+    parental_consent_rejected:
+      subject: Issue with parent/guardian consent form
+
   team_mailer:
     invite_member:
       greeting:
--- a/app/data_grids/parental_consents_grid.rb
+++ b/app/data_grids/parental_consents_grid.rb
@@ -45,8 +45,18 @@ class ParentalConsentsGrid
       }
     )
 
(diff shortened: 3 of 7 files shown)


app/controllers/admin/paper_parental_consents_controller.rb
```diff
@@ -29,6 +29,17 @@ def approve
 
       redirect_to admin_paper_parental_consents_path,
         success: "You approved the parental consent for #{parental_consent.student_profile_full_name}."
+
```
viviancan, line 32: Missing `end` keyword for the `approve` method

config/locales/mailers.en.yml
```diff
@@ -39,6 +39,10 @@ en:
     welcome_returning_judge:
       subject: "Welcome to Technovation %{season_year}!"
 
+  student_mailer:
+    parental_consent_rejected:
+      subject: Issue with parent/guardian consent form
```
viviancan, line 44: Should this subject string be in quotes?

viviancan (approved):
Tried locally, and with the fixed `end` syntax it works as expected! We should discuss what happens if the paper consent is rejected. On the student side, after my parental consent was rejected, I could upload again. I uploaded a different file, but then in the admin table, it still showed as rejected. Nothing major, but something to work out I think! Maybe if it is rejected, they aren't able to upload through the student dashboard again? 🤔 Great work!

app/controllers/admin/paper_parental_consents_controller.rb
shaun-technovation: Ah, great catch on this! I botched this when I was resolving diff conflicts! 😊 🤦 😂

config/locales/mailers.en.yml
shaun-technovation: Should be fine w/o quotes, I'm pretty sure the quotes are needed when there is string interpolation. 👍

shaun-technovation:
> Tried locally, and with the fixed `end` syntax it works as expected! We should discuss what happens if the paper consent is rejected. On the student side, after my parental consent was rejected, I could upload again. I uploaded a different file, but then in the admin table, it still showed as rejected. Nothing major, but something to work out I think! Maybe if it is rejected, they aren't able to upload through the student dashboard again? 🤔 Great work!

Ah, great call-out on rejecting/re-uploading! I think for now I'll just make a quick fix so that any re-upload will reset it to a pending status (so it can be approved or rejected again in the admin).

shaun-technovation:
> > Tried locally, and with the fixed `end` syntax it works as expected! We should discuss what happens if the paper consent is rejected. On the student side, after my parental consent was rejected, I could upload again. I uploaded a different file, but then in the admin table, it still showed as rejected. Nothing major, but something to work out I think! Maybe if it is rejected, they aren't able to upload through the student dashboard again? 🤔 Great work!
> 
> Ah, great call-out on rejecting/re-uploading! I think for now I'll just make a quick fix so that any re-upload will reset it to a pending status (so it can be approved or rejected again in the admin).

Just FYI, I made this update in #4142. 👍

PR #4145, opened 2023-08-09
Add "additional questions" for 2024 season

shaun-technovation:
This will use the same questions for the 2024 season as 2023 one; this will fix the bug, we can update the questions later if necessary.


viviancan (approved):
Did not try locally, but looks straight forward!

PR #4164, opened 2023-08-17
Various work related to uploaded consent forms

shaun-technovation:
This PR includes various work related to uploaded consent forms:
- Ability to view uploaded media consent forms (new admin datagrid)
- Ability to approve and reject media consents
- Hiding the upload section if one has been uploaded already
- Fixing the issues where if no consent form file is selected when the form is submitted it causes an error
- Some styling tweaks for the "View" button in the admin
- Some refactoring - moving shared constants to `ConsentForms` constants


viviancan (commented):
I was able to upload the media consent and the parental consent form at the same time. I could see both uploads in their tables! 

For media consents and parental consents, I could click on the action buttons. However, the status did not seem to update. For example, when I clicked "approve with consent," the status displayed as `pending` on the grid. When I checked the DB, the `upload_approval_status` status was still 0 and the `upload_approved_at` timestamp was still null. When I returned to the individual participant detail page, I saw that media consent was approved. It was similar behavior for parental consents. 

Also not sure if we are getting rid of the "grant permission" / "do not grant permission" functionality for media consents on the detailed participant page, but when I tried to click one of those buttons, I got the following error 
```
NameError (uninitialized constant MediaConsent::ELECTRONIC_SIGNATURE_FOR_A_PAPER_MEDIA_CONSENT):
  
app/controllers/admin/paper_media_consents_controller.rb:13:in `create'
```

I think we should also finalize the naming of the action buttons so they can be consistent. On the detailed participant page, the action buttons are `grant permission` and `do not grant permission`. For the same action on the media consent table, the values are `approve with consent` and `approve without consent`.

viviancan:
Great work on this!! Excited to see it all come together. I know it's going to be such a big help for the admin portal/program team. I think it's pretty much ready to get  👀 on it, besides the status issue mentioned above ^^ 💯 🚀

shaun-technovation:
> For media consents and parental consents, I could click on the action buttons. However, the status did not seem to update. For example, when I clicked "approve with consent," the status displayed as `pending` on the grid. When I checked the DB, the `upload_approval_status` status was still 0 and the `upload_approved_at` timestamp was still null. When I returned to the individual participant detail page, I saw that media consent was approved. It was similar behavior for parental consents.

Good catch on this, this should be fixed now!

> Also not sure if we are getting rid of the "grant permission" / "do not grant permission" functionality for media consents on the detailed participant page, but when I tried to click one of those buttons, I got the following error
> 
> ```
> NameError (uninitialized constant MediaConsent::ELECTRONIC_SIGNATURE_FOR_A_PAPER_MEDIA_CONSENT):
>   
> app/controllers/admin/paper_media_consents_controller.rb:13:in `create'
> ```

Good catch on this one too! It should also be fixed! 👍 

> I think we should also finalize the naming of the action buttons so they can be consistent. On the detailed participant page, the action buttons are `grant permission` and `do not grant permission`. For the same action on the media consent table, the values are `approve with consent` and `approve without consent`.

Good call out! I agree, we should definitely make them consistent. 💯

PR #4166, opened 2023-08-18
Update Axios module

shaun-technovation:
This will update Axios to address this security vulnerability:
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/74


viviancan (approved):
Did not try locally, but seems like a small change 💯

PR #4168, opened 2023-08-18
Update Moment.js module

shaun-technovation:
This will update Moment.js to address the following security vulnerabilities:
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/48
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/86


viviancan (approved):
Did not try locally but seems like a small upgrade!

Issue #4149, opened 2023-08-11
Check technical additions as complete even if someone does not provide the optional Thunkable email


dboyer:
Currently a student does not have to provide an email to accompany their Thunkable link as it is an optional field. However, the technical section is not marked with a green check mark as completed unless they fill in the optional field.

Is there a way to make this section marked as completed if only an optional field is not filled in?

<img width="1079" alt="Screen Shot 2023-08-11 at 10 04 15 AM" src="https://github.com/Iridescent-CM/technovation-app/assets/516764/a500728b-9e4f-4431-97ba-c42f2f0f02eb">

dboyer:
Confirmed on QA that the technical elements section is marked as complete even if I did not add the optional thunkable email

PR #4171, opened 2023-08-22
Updated logic for completed technical elements section for optional gmail/email field

viviancan:
Refs #4149 

This change will update the conditions for the completed green check mark for the technical elements sections. The updated logic checks if the email field is blank and if it's not blank check the errors for the field. This change will allow for the green checkmark and the section to display as complete if the optional email fields are not included.


shaun-technovation (approved):
LGTM! I didn't try it locally, but I trust it works! 👍 🚀 🚢

PR #4174, opened 2023-08-25
Verify returning students age

shaun-technovation:
This will add a `before_action` hook for students to verify that they will be 18 or younger by the division cut-off date. If they're going to be older, we'll display the new mentor conversion page with the option to become a mentor.

### DIFF
--- a/app/controllers/concerns/verify_student_age_concern.rb
+++ b/app/controllers/concerns/verify_student_age_concern.rb
@@ -1,0 +1,24 @@
+module VerifyStudentAgeConcern
+  extend ActiveSupport::Concern
+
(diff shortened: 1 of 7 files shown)


viviancan (approved):
Nice addition!! Everything worked as expected. I only noticed that the interrupt screen with the container was off-center. 
CleanShot 2023-08-29 at 13 29 48
💯 🚀

shaun-technovation:
> I only noticed that the interrupt screen with the container was off-center. CleanShot 2023-08-29 at 13 29 48 💯 🚀

I just centered it, nice suggestion! 👍

Issue #4155, opened 2023-08-11
Admin invitation emails should go to the new registration flow


shaun-technovation:
When an admin invites someone to register, the invite email should have a link (with an invite code) to the new registration flow.

When the invited person clicks on the invite link (in the email), they will be taken the first page of the registration system where the invited profile type will be pre-selected, and the other profile types will be disabled.

Invite codes can only be used once.
If registration is closed for a particular profile type, all associated invites for that profile type shouldn't work.

dboyer:
I like the idea of having them go to the first page with the profile type already selected as a reminder of what they are signing up for.

shaun-technovation:
This one is on QA! Admin invites to student, mentors, and judges should be using the new registration system now. 👍

dboyer:
Tested on QA:

- Successfully invited a judge with a new email account. I received the email and was sent to the new registration system where I registered as a new judge.
- Successfully invited a mentor and was able to register as that mentor. I changed the email address from the one the invite was sent to. I received an error message when I tried to select an address already in the system but was able to change to a different address. NOTE: on the admin page, the invite to that mentor says "registered" so I'm guessing the invite is tied to the token not the email address. That feels correct.
- Successfully invited a student and the link in the email directed to the new registration page. However, it was only selecting the 13-18 year old option and did not allow me to switch to the 8-12 year old option.

I tried to enter a birthdate that would make the student 11 and the validation check flagged that the student was too young. It prompted me to go back to the previous screen and select the other student type which I was not able to do.

@shaun-technovation Is there a way for the student invite to default to the age 13-18 but allow selecting age 8-12 too? If not, maybe we need to add two student invite types - one for 8-12 and one for 13-18.

shaun-technovation:
> @shaun-technovation Is there a way for the student invite to default to the age 13-18 but allow selecting age 8-12 too? If not, maybe we need to add two student invite types - one for 8-12 and one for 13-18.

Hmm, I can see at least two options:
- A student invite could enable both the student and parent profile type
- Adding a new parent invite that would enable the parent type

Also, having a message like "If you would like to register as a different participant type, please click here" could help too, but we need to get #4189 fixed for this functionality.

dboyer:
I don't have a strong preference between the options if one of them is technically easier than the other.

shaun-technovation:
> I don't have a strong preference between the options if one of them is technically easier than the other.

I created #4215 so an admin can invite a parent. 👍

PR #4178, opened 2023-08-30
Integrate admin invite codes with new registration flow

shaun-technovation:
This will make it so that after someone registers with an admin invite code (aka `admin_permission_token`), the associated invitation will get updated with their newly created account and be marked as "registered", this will also be reflected in the invitations area in the admin.

### DIFF
--- a/app/controllers/new_registration_controller.rb
+++ b/app/controllers/new_registration_controller.rb
@@ -1,4 +1,6 @@
 class NewRegistrationController < ApplicationController
+  after_action :update_invitation, only: :create
+
   layout "new_registration"
 
   def show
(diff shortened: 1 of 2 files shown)


app/controllers/new_registration_controller.rb
```diff
@@ -114,4 +117,13 @@ def registration_params
       mentorTypes: []
     )
   end
+
+  def update_invitation
+    invitation = UserInvitation.find_by(admin_permission_token: params[:inviteCode])
```
shaun-technovation, line 122: I decided to refer to the `admin_permission_token` as an "invite code". It will still be referred to as `admin_permission_token` on the back-end though. And ideally, we'll update the URL parameter in #4155, so it says `invite_code` instead of `admin_permission_token` in the address bar in the browser.

viviancan (approved):
LGTM 💯 🚀  I tried it locally and saw the updated invitation status in the admin portal

PR #4182, opened 2023-08-30
Various consent form upload improvements

shaun-technovation:
Bundling together a few improvements/fixes related to consent form uploads:
- Ensuring a consent forms exist when impersonating a student
- Only allowing a media consent to be uploaded if a parental consent is also being uploaded
- Text changes for paper media consents (related to the comment here: https://github.com/Iridescent-CM/technovation-app/pull/4164#pullrequestreview-1585073082)
- A couple HTML tweaks, mismatched header tags, etc. (good catch on those)

### DIFF
--- a/app/controllers/student/paper_consent_uploads_controller.rb
+++ b/app/controllers/student/paper_consent_uploads_controller.rb
@@ -4,9 +4,13 @@ module Student
       @parental_consent = current_student.parental_consent
       @media_consent = current_student.media_consent
 
-      if params[:parental_consent].blank?
-        redirect_to student_dashboard_path, error: "You forgot to upload a consent form. Please try again."
-      elsif @parental_consent.update(parental_consent_params) && @media_consent.update(media_consent_params)
+      if parental_consent_params[:uploaded_consent_form].blank?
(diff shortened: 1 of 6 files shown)


viviancan (approved):
Everything worked as expected! Just 2 points
- Maybe there should be some sort of helper text saying that the media consent can only be uploaded when the parental consent is uploaded
- The media consent didn't have a max file size. I was able to upload a file size bigger than 3MB
CleanShot 2023-08-31 at 10 49 38
Great work on this!!

viviancan:
test

shaun-technovation:
> * Maybe there should be some sort of helper text saying that the media consent can only be uploaded when the parental consent is uploaded
> * The media consent didn't have a max file size. I was able to upload a file size bigger than 3MB

Hmm, I couldn't reproduce the error. I didn't change the validations so they should still be working. I think what could have happened is that it didn't allow the file to be uploaded, but also didn't display an error, just silently failing. Anywho, I made some more changes and tested again, and it's working as expected for me, guess we'll just keep an eye on it. And the text suggestion won't be necessary now, I made it so that the consent forms can be uploaded separately or together in any order. :+1:

PR #4186, opened 2023-09-01
Update simple-get module

shaun-technovation:
A simple update to address the following high security vulnerability:
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/66


viviancan (approved):
Did not try locally, but seems like a small bump!

PR #4188, opened 2023-09-07
Make admin invites use new registration system

shaun-technovation:
This will make it so that when an admin invites a judge, mentor, or student it will go to the new registration system; chapter ambassadors will still use the old/existing registration system.

Basic overview of the work here:
- Update invite emails
- Rename `admin_permission_token` to `invite_code`
- A new back-end API endpoint that will get invite details
- Front-end will use new invite details endpoint, and:
  - Pre-select invited profile type
  - Disable non-invited profile types

### DIFF
--- a/app/controllers/api/registration_invites_controller.rb
+++ b/app/controllers/api/registration_invites_controller.rb
@@ -1,0 +1,13 @@
+module Api
+  class RegistrationInvitesController < ApplicationController
+    def show
+      invite = RegistrationInviteValidator.new(invite_code: params[:id]).call
+
+      render json: {
+        isValid: invite.valid?,
+        profileType: invite.profile_type,
+        friendlyProfileType: invite.friendly_profile_type
+      }
+    end
+  end
+end
--- a/app/javascript/new_registration/components/FormWrapper.vue
+++ b/app/javascript/new_registration/components/FormWrapper.vue
@@ -97,7 +97,7 @@ export default {
     },
     async submitHandler (data) {
       const csrfTokenMetaTag = document.querySelector('meta[name="csrf-token"]')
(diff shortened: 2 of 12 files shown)


viviancan (approved):
Great work, everything looks good to me!!

The only thing that stands out to me (and it may be a nonissue) is that the current invite system auto-assigns the invited email address to the account email address. I verified this with a ChA on QA and with a test mentor on production. The language in the mailer also leads me to believe this 
`Please confirm your email address....` and then when I click the link for the old system, you can see the email address is already assigned.
Screen Shot 2023-09-13 at 5 31 20 PM

I'm unsure if this matters since they can update their email after registration, but I wanted to note it! 

Besides that, everything worked as expected 💯 🚀

shaun-technovation:
> The only thing that stands out to me (and it may be a nonissue) is that the current invite system auto-assigns the invited email address to the account email address. I verified this with a ChA on QA and with a test mentor on production. The language in the mailer also leads me to believe this `Please confirm your email address....` and then when I click the link for the old system, you can see the email address is already assigned. Screen Shot 2023-09-13 at 5 31 20 PM

Thanks for taking a look at this one! I mentioned the email address pre-populating functionality in one of our meeting, and I thought we were okay with leaving that out; for now at least. Anywho, we can add that at a later time too! The invite emails though, good catch! I think the text in those could use a refresh too, I'll create a ticket for that. I'll get this one merged in so we can iterate and build off of it. 👍

shaun-technovation:
I created #4200 to address the admin invite emails. 👍

shaun-technovation:
> We can keep the pre-populated email address. PT is good with that approach.

Cool, I can add another issue to work on the later. 👍

PR #4197, opened 2023-09-13
PT2 - Checkr International Background Check Integration - Mostly backend work

viviancan:
This is PT2 of the international bg check integration

This PR contains the 
- API client for making manual requests to Checkr 
- enum updates in BackgroundCheck
- Added INDIA to logic that determines when a bg is required

### DIFF
--- a/app/models/background_check.rb
+++ b/app/models/background_check.rb
@@ -1,6 +1,7 @@
 class BackgroundCheck < ActiveRecord::Base
   # If these change, you will need to update dataclips
-  enum status: %i[pending clear consider suspended canceled]
+  enum status: %i[pending clear consider suspended canceled invitation_sent]
+  enum invitation_status: %i[pending completed expired], _prefix: :invitation
 
   belongs_to :account
 
--- a/app/models/background_check.rb
+++ b/app/models/background_check.rb
@@ -15,6 +15,10 @@ class BackgroundCheck < ActiveRecord::Base
     pending?
   end
 
+  def invitation_sent?
+    candidate_id.present? && invitation_id.present?
+  end
+
   class << self
     def get(resource, id)
       Checkr.request(:get, "/v1/#{resource}/#{id}")
--- a/app/models/mentor_profile.rb
+++ b/app/models/mentor_profile.rb
@@ -165,6 +165,8 @@ class MentorProfile < ActiveRecord::Base
   delegate :submitted?,
            :candidate_id,
            :report_id,
+           :invitation_id,
+           :invitation_status,
     to: :background_check,
     prefix: true,
     allow_nil: true
@@ -271,6 +273,12 @@ class MentorProfile < ActiveRecord::Base
     !!background_check_candidate_id and !!background_check_report_id
   end
 
+  def background_check_invitation_sent?
+    !!background_check_candidate_id and
+      !!background_check_invitation_id and
+        background_check.invitation_pending?
+  end
+
   def enable_searchability
     self.searchable = can_enable_searchable?
 
@@ -300,7 +308,7 @@ class MentorProfile < ActiveRecord::Base
 
   def requires_background_check?
     (account.valid? and account.age >= 18) and
-      account.country_code == "US" and
+      (account.country_code == "US" or account.country_code == "IN") and
         not (background_check.present? and background_check.clear?)
   end
 
--- a/app/models/mentor_profile.rb
+++ b/app/models/mentor_profile.rb
@@ -166,7 +166,6 @@ class MentorProfile < ActiveRecord::Base
            :candidate_id,
            :report_id,
            :invitation_id,
-           :invitation_status,
     to: :background_check,
     prefix: true,
     allow_nil: true
@@ -271,12 +270,6 @@ class MentorProfile < ActiveRecord::Base
 
   def background_check_submitted?
     !!background_check_candidate_id and !!background_check_report_id
-  end
-
-  def background_check_invitation_sent?
-    !!background_check_candidate_id and
-      !!background_check_invitation_id and
-        background_check.invitation_pending?
   end
 
   def enable_searchability
--- a/app/services/custom_checkr/api_client.rb
+++ b/app/services/custom_checkr/api_client.rb
@@ -1,0 +1,81 @@
+module CustomCheckr
+  class ApiClient
+    def initialize(
+      client_id: ENV.fetch("CHECKR_API_KEY"),
+      base_url: ENV.fetch("CHECKR_API_BASE", "https://api.checkr.com"),
+      logger: Rails.logger,
+      error_notifier: Airbrake
+    )
+
+      @connection = Faraday.new(
+        url: base_url,
+        headers: {"Content-Type" => "application/json"}
+      ) do |conn|
+        conn.request :authorization, :basic, client_id, ""
+      end
+
+      @logger = logger
+      @error_notifier = error_notifier
+    end
+
+    def request_checkr_invitation(resource, profile:)
+      candidate_response = connection.post("/v1/#{resource}") do |req|
+        req.body = {
+          email: profile.account.email,
+          first_name: profile.account.first_name,
+          last_name: profile.account.last_name,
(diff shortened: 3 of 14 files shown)


app/models/mentor_profile.rb
```diff
@@ -300,7 +308,7 @@ def background_check_complete?
 
   def requires_background_check?
     (account.valid? and account.age >= 18) and
-      account.country_code == "US" and
+      (account.country_code == "US" or account.country_code == "IN") and
```
viviancan, line 311: This will eventually be updated to pull from ENV variable

app/services/custom_checkr/api_client.rb
```diff
@@ -0,0 +1,81 @@
+module CustomCheckr
+  class ApiClient
+    def initialize(
```
viviancan, line 3: I followed the approach used for the learn worlds API client

app/services/custom_checkr/api_client.rb
```diff
@@ -0,0 +1,81 @@
...
+            work_locations: [{country: profile.account.country_code}]
+          }.to_json
+        end
+
+        invitation_response_body = JSON.parse(invitation_response.body, symbolize_names: true)
+
+        if invitation_response.status == 201
+          profile.account.update(background_check_attributes: {
+            candidate_id: candidate_response_body[:id],
+            report_id: "international",
```
viviancan, line 47: I am going to remove the requirement for a report_id and then this will not be needed

app/models/background_check.rb
```diff
@@ -1,6 +1,7 @@
 class BackgroundCheck < ActiveRecord::Base
   # If these change, you will need to update dataclips
-  enum status: %i[pending clear consider suspended canceled]
+  enum status: %i[pending clear consider suspended canceled invitation_sent]
+  enum invitation_status: %i[pending completed expired], _prefix: :invitation
```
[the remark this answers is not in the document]
shaun-technovation: Does `_prefix: invitation` mean this will be accessed as `some_background_check.invitation_invitation_status`?

app/models/mentor_profile.rb
```diff
@@ -271,6 +273,12 @@ def background_check_submitted?
     !!background_check_candidate_id and !!background_check_report_id
   end
 
+  def background_check_invitation_sent?
+    !!background_check_candidate_id and
+      !!background_check_invitation_id and
+        background_check.invitation_pending?
```
shaun-technovation, line 279: A few things on this one:
- I've been slowing trying to replace the existing `and`s with `&&`, a few notes on this:
  - https://stackoverflow.com/a/2083630
  - https://rubystyle.guide/#and-or-flow
- Instead of the `!!`, I think using `present?` would read better, but no big deal

And the naming seems to be a little off:  `background_check_invitation_sent?` ... `background_check.invitation_pending?`, if the background check is sent, I would think that would be independent of the status. 🤔

app/models/mentor_profile.rb
```diff
@@ -165,6 +165,8 @@ class MentorProfile < ActiveRecord::Base
   delegate :submitted?,
            :candidate_id,
            :report_id,
+           :invitation_id,
+           :invitation_status,
```
shaun-technovation, line 169: Is this being used? I can't seem to find where. 🤔 👀

app/services/custom_checkr/api_client.rb
```diff
@@ -0,0 +1,81 @@
...
+      base_url: ENV.fetch("CHECKR_API_BASE", "https://api.checkr.com"),
+      logger: Rails.logger,
+      error_notifier: Airbrake
+    )
+
+      @connection = Faraday.new(
+        url: base_url,
+        headers: {"Content-Type" => "application/json"}
+      ) do |conn|
+        conn.request :authorization, :basic, client_id, ""
```
shaun-technovation, line 14: Wondering, what is the `""` is for? Can it be removed or replaced with `_` (indicating not used) instead. Not a big deal, just curious.

app/services/custom_checkr/api_client.rb
```diff
@@ -0,0 +1,81 @@
...
+      end
+
+      @logger = logger
+      @error_notifier = error_notifier
+    end
+
+    def request_checkr_invitation(resource, profile:)
+      candidate_response = connection.post("/v1/#{resource}") do |req|
```
shaun-technovation, line 22: This takes `resource`, but can it be hard-coded to `candidates` instead? Is there a benefit to having it dynamic? The method suggests this would only be used for invitations and it's being assigned to `candidate_response` suggesting it's already for a `candidates`. 🤔

app/services/custom_checkr/api_client.rb
```diff
@@ -0,0 +1,81 @@
...
+        headers: {"Content-Type" => "application/json"}
+      ) do |conn|
+        conn.request :authorization, :basic, client_id, ""
+      end
+
+      @logger = logger
+      @error_notifier = error_notifier
+    end
+
+    def request_checkr_invitation(resource, profile:)
```
shaun-technovation, line 21: This might line-up better if `profile:` was `candidate:`, and then it would be called w/ something like `candidate: current_mentor` or something like that, since this method is entering the Checkr world where they're referred to as candidates, anywho, just a suggestion, not a blocker.

app/services/custom_checkr/api_client.rb
```diff
@@ -0,0 +1,81 @@
+module CustomCheckr
```
shaun-technovation, line 1: Can this be renamed to just `Checkr` or does that clash w/ the gem? If that's the case, I might suggest renaming it to `CheckrApiClient`vs `CustomCheckr::ApiClient`, anywho, another not a big deal one! 👍

app/services/custom_checkr/api_client.rb
```diff
@@ -0,0 +1,81 @@
...
+            package: "international_basic_plus",
+            candidate_id: candidate_response_body[:id],
+            work_locations: [{country: profile.account.country_code}]
+          }.to_json
+        end
+
+        invitation_response_body = JSON.parse(invitation_response.body, symbolize_names: true)
+
+        if invitation_response.status == 201
+          profile.account.update(background_check_attributes: {
```
shaun-technovation, line 45: Just another one I'm curious on, using `profile.account.update` vs `profile.background_check.update`?

shaun-technovation (approved):
I left some curiosity questions and a few suggestions, I didn't see any blockers. Overall, looks like some nice functionality additions to me! Great work! Having this functionality to request an invite is a huge win! 💯 👏 🙌 🎉

app/models/background_check.rb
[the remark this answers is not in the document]
viviancan: It can be accessed like `some_background_check.invitation_status`

The prefix was needed since there was already a "pending" value defined in this model. So it can be used like this `some_background_check.invitation_pending?` to check the invitation_status vs `some_background_check.pending`  for the report status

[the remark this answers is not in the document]
shaun-technovation: Ah yes, that makes sense, thank you! I remember now I had to do that for the consent forms, for upload status, since regular "status" was already taken there too. 👍

app/services/custom_checkr/api_client.rb
viviancan: I tried using the `_` and received this error `NameError (undefined local variable or method `_' for #<CheckrApiClient::ApiClient:0x000000010e14b950>`

If I don't include it all, I get a bad authentication error from Checkr. [The Checkr docs](https://docs.checkr.com/#section/Getting-Started/Authenticate-with-Checkr) state 
`Use HTTP Basic authentication to authenticate with Checkr, with your API key as the username and an empty password.` I didn't see anything else in the Faraday docs, but I can keep searching!

app/services/custom_checkr/api_client.rb
viviancan: 👍

app/services/custom_checkr/api_client.rb
viviancan: Updated in #4208

PR #4199, opened 2023-09-13
Fix webdrivers

shaun-technovation:
To fix this locally I ran:
> bundle update selenium-webdriver
> bundle update capybara

but, I think since I updated the Gemfile.lock, a simple `bundle` should work too.

Anyhow, after I updated both of those gems, I was able to run `:js` specs again.

Also, here are some comments on how I discovered we need to update both selenium and capybara:
https://github.com/titusfortner/webdrivers/issues/247#issuecomment-1650315937
https://github.com/titusfortner/webdrivers/issues/247#issuecomment-1648154088
https://github.com/titusfortner/webdrivers/issues/247#issuecomment-1654522509


viviancan (approved):
Nice fix 💯 great find on the comments too!!

PR #4196, opened 2023-09-13
PT1 - Checkr International Background Check Integration

viviancan:
Refs #4081 

This PR consists of a few additions
- Route/controller action for international bg check
- Frontend updates 
   -  Updates to the little green checkmark logic for background checks in mentor profile
   -  2 new partials for the main background check view `us_background_check` & `international_background_check`
   - A few additions of logic in the views depending on if an invitation is pending or not
- DB addition - `invitation_status` and `invitation_id` added to `background_check` table

Note* This PR is currently failing in circle ci which is to be expected because some of the code needed is in PT2 PR

PR #4203, opened 2023-09-14
PT3 - Checkr International Background Check Integration - Invitation Status updates

viviancan:
Refs #4180 

This PR is part 3 of the background check work. This update was needed in order to capture the invitation status and the background check report ID from checkr.

Current flow for requesting an international bg check invitation:
1. Mentors in certain countries are required to complete a bg check
2. Mentor clicks button "request bg check invitation"
3. Candidate is created, then invitation using candidate id is created
4. This invitation is sent to mentor's email address
5. Mentor will click on the link in the email from checkr which will lead them to the Checkr portal
6. They will complete the bg check in the the checkr portal
7. When they log back into technovation platform there is an option to check the invitation status. This is where this PR comes into place. There is new logic that will request the invitation from checkr. 
8. We can save the updated status ( options are pending, completed, expired) and the now available `report_id`
9. Then the flow continues with report status check

### DIFF
--- a/app/controllers/concerns/background_check_controller.rb
+++ b/app/controllers/concerns/background_check_controller.rb
@@ -8,7 +8,13 @@ module BackgroundCheckController
   end
 
   def show
-    BackgroundChecking.new(current_profile.background_check).execute
+    if current_profile.background_check.invitation_id.present?
+      InvitationChecking.new(current_profile.background_check).execute
+    end
+
+    if current_profile.background_check.report_id.present?
+      BackgroundChecking.new(current_profile.background_check).execute
+    end
     @background_check = current_profile.background_check
   end
 
--- a/app/services/custom_checkr/api_client.rb
+++ b/app/services/custom_checkr/api_client.rb
@@ -44,7 +44,6 @@ module CustomCheckr
         if invitation_response.status == 201
           profile.account.update(background_check_attributes: {
             candidate_id: candidate_response_body[:id],
-            report_id: "international",
             invitation_id: invitation_response_body[:id],
             invitation_status: invitation_response_body[:status],
             status: :invitation_sent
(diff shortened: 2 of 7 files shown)


app/controllers/concerns/background_check_controller.rb
```diff
@@ -8,7 +8,13 @@ def new
   end
 
   def show
-    BackgroundChecking.new(current_profile.background_check).execute
+    if current_profile.background_check.invitation_id.present?
```
shaun-technovation, line 11: In [#4196](https://github.com/Iridescent-CM/technovation-app/pull/4196/files#r1326425043), I made a comment suggesting a new helper method `#international_background_check?`, but now seeing this, I think something like `#invited_background_check?` or `#background_check_invite?` (or something like that) would make more sense - if "invited background check" then do `InvitationChecking` or if "invited background check" then display background invite page/form... etc.

app/services/custom_checkr/api_client.rb
```diff
@@ -68,14 +67,26 @@ def request_checkr_invitation(resource, profile:)
...
 
-    # todo: get request for invitation status update from checkr
-    # def get_checkr_invitation()
-    # end
+    def retrieve_invitation(invitation_id)
+      invitation_response = connection.get("/v1/invitations/#{invitation_id}")
+
+      invitation_response_body = JSON.parse(invitation_response.body, symbolize_names: true)
+
+      if invitation_response.status == 200
```
shaun-technovation, line 75: Ah, I was wondering when I looked at #4197 if Farady has helper method for this, and I just checked and it looks like there is one - `#success?`.

https://stackoverflow.com/a/60425238

app/technovation/invitation_checking.rb
```diff
@@ -0,0 +1,28 @@
+class InvitationChecking
```
shaun-technovation, line 1: Ah, looks like this was copy/pasted from `BackgroundChecking`, I see that file and this could use some improvements, like using keyword arguments instead of an options hash, accessing methods instead of instance variables, updating the default logger, it looks like `#retrieve_invitation` returns an object, but the properties are being accessed like a hash... anywho, all just FWIW stuff, looks like this will get the job done. 👍

app/views/background_checks/_international_background_check.html.erb
```diff
@@ -90,7 +90,9 @@
   <div class="grid__col-sm-5">
     <div class="panel">
       <h3><%= t("views.background_checks.new.international_title") %></h3>
-      <%= link_to "Request background check invitation", mentor_international_background_check_url %>
+      <%= link_to "Request background check invitation",
+                  mentor_international_background_check_url,
```
shaun-technovation, line 94: Ah, related to my random comments on standup, this might be more versatile as "background check invite url" vs "international background check url", of course that would mean route/controller/etc updates.  Anywho, another FWIW. 😄

shaun-technovation (approved):
LGTM! Just left some FWIW type comments, but no blockers. I didn't test any of it locally, but looking forward to seeing it all come together on QA! 👍 🙌 🥳 🎉 👏 :shipit:

PR #4206, opened 2023-09-15
Update Sidekiq gem

shaun-technovation:
A high security vulnerability came in yesterday, the changes in this PR will update Sidekiq to address this:
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/164

I also had to make a few other updates to account for this upgrade (we upgraded from 6.x to 7.x):
- Addressing this deprecation warning:
  - DEPRECATION WARNING: default_worker_options= is deprecated and will be removed from Sidekiq 7.0 (use default_job_options= instead)
- Requiring "redis" to address this error:
  - NameError (uninitialized constant Redis)

I also included a couple of code style updates.

I tested this by viewing and updating "Content & Settings", I was running into issues at first^ and then addressed those, and now it's working for me locally.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -54,7 +54,7 @@ gem "friendly_id", "~> 5.5"
 
 gem "bcrypt", "~> 3.1"
 
-gem "sidekiq", "~> 6.4.0"
+gem "sidekiq", "~> 7.0"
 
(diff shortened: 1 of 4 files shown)


viviancan (approved):
LGTM! I tested locally and changed the content settings. I also registered as a new judge!

PR #4207, opened 2023-09-15
Admin invites for Chapter Ambassadors using new registration system

shaun-technovation:
This will add the basic functionality so that when an admin invites a chapter ambassador:
- the invite email will have a link with an invite code that will go to the new registration system
- a new chapter ambassador profile (which needs a new icon) will be pre-selected
  - this new profile type should only be shown with a valid chapter ambassador invite code
- after the new chapter ambassador profile has been created, they will be put back into the regular/existing "pending" chapter ambassador flow


viviancan (approved):
It worked as expected when I invited a ChA locally! 💯 This is a great addition ✅ 🚀

Issue #4208, opened 2023-09-18
International bg check code review feedback


viviancan:
Creating a ticket to tackle code review feedback for international bg check

viviancan:
- [ ] Review the method name ["international_background_check"](https://github.com/Iridescent-CM/technovation-app/pull/4196#discussion_r1326421577)
- [ ] Rename `CustomCheckr` to [`CheckrApiClient`](https://github.com/Iridescent-CM/technovation-app/pull/4197/files#r1326341063)
- [ ] Update `profile` to `candidate` in the checkr [API request checkr invite](https://github.com/Iridescent-CM/technovation-app/pull/4197/files#r1326337349)
- [ ] Review [faraday success method](https://github.com/Iridescent-CM/technovation-app/pull/4203#discussion_r1327708053)

PR #4210, opened 2023-09-19
International Background Check code review feedback

viviancan:
Refs #4208 
This PR makes a few changes based on code review feedback

- Rename `CustomCheckr` to [`CheckrApiClient`](https://github.com/Iridescent-CM/technovation-app/pull/4197/files#r1326341063)
- Update `profile` to `candidate` in the checkr [API request checkr invite](https://github.com/Iridescent-CM/technovation-app/pull/4197/files#r1326337349)
- Review [faraday success method](https://github.com/Iridescent-CM/technovation-app/pull/4203#discussion_r1327708053)
- Review approach with [InvitationChecking](https://github.com/Iridescent-CM/technovation-app/pull/4203#discussion_r1327715328)

I also added the following
- Add DOB to API call for international bg check
- Review the conditional logic for text when the invitation is pending vs. complete

### DIFF
--- a/app/controllers/concerns/background_check_controller.rb
+++ b/app/controllers/concerns/background_check_controller.rb
@@ -19,7 +19,7 @@ module BackgroundCheckController
   end
 
   def international_background_check
-    checkr_result = CustomCheckr::ApiClient.new.request_checkr_invitation("candidates", profile: current_profile)
+    checkr_result = CheckrApiClient::ApiClient.new.request_checkr_invitation(candidate: current_profile)
 
     if checkr_result.success?
       redirect_to send("#{current_scope}_dashboard_path"),
--- a/app/models/background_check.rb
+++ b/app/models/background_check.rb
@@ -15,10 +15,6 @@ class BackgroundCheck < ActiveRecord::Base
     pending?
   end
 
-  def invitation_sent?
-    candidate_id.present? && invitation_id.present?
-  end
-
   class << self
     def get(resource, id)
       Checkr.request(:get, "/v1/#{resource}/#{id}")
--- a/app/services/checkr_api_client/api_client.rb
+++ b/app/services/checkr_api_client/api_client.rb
@@ -31,7 +31,7 @@ module CheckrApiClient
 
       candidate_response_body = JSON.parse(candidate_response.body, symbolize_names: true)
 
-      if candidate_response.status == 201
+      if candidate_response.success?
         invitation_response = connection.post("/v1/invitations") do |req|
           req.body = {
             package: "international_basic_plus",
@@ -42,8 +42,8 @@ module CheckrApiClient
 
         invitation_response_body = JSON.parse(invitation_response.body, symbolize_names: true)
 
-        if invitation_response.status == 201
-          profile.account.update(background_check_attributes: {
+        if invitation_response.success?
+          candidate.account.update(background_check_attributes: {
             candidate_id: candidate_response_body[:id],
             invitation_id: invitation_response_body[:id],
             invitation_status: invitation_response_body[:status],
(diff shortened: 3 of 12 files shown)


app/models/background_check.rb
```diff
@@ -15,10 +15,6 @@ def submitted?
     pending?
   end
 
-  def invitation_sent?
```
viviancan, line 18: You made a comment about the naming and why it depended on the status in a [previous PR](https://github.com/Iridescent-CM/technovation-app/pull/4197#discussion_r1326313730)

> And the naming seems to be a little off: background_check_invitation_sent? ... background_check.invitation_pending?, if the background check is sent, I would think that would be independent of the status. 🤔

Which had me thinking why I had included the status. When I revisited the logic, I needed the status because I only wanted that text to display if the invitation status was pending (vs complete or expired). So I decided to remove this and just check the `invitation_status`

app/views/background_checks/_us_background_check.html.erb
```diff
@@ -81,7 +81,7 @@
           <p>
             You can use the
             <%= link_to "Checkr Applicant Portal",
-                        "//applicant.custom_checkr.com",
+                        "//applicant.checkr.com",
```
viviancan, line 84: fixed typo

app/services/checkr_api_client/api_client.rb
```diff
@@ -18,31 +18,32 @@ def initialize(
...
+      candidate_response = connection.post("/v1/candidates") do |req|
         req.body = {
-          email: profile.account.email,
-          first_name: profile.account.first_name,
-          last_name: profile.account.last_name,
-          work_locations: [{country: profile.account.country_code}]
+          email: candidate.account.email,
+          first_name: candidate.account.first_name,
+          last_name: candidate.account.last_name,
+          dob: candidate.account.date_of_birth,
```
viviancan, line 27: Added DOB

app/services/checkr_api_client/api_client.rb
```diff
@@ -1,4 +1,4 @@
-module CustomCheckr
+module CheckrApiClient
```
viviancan, line 1: Renamed module

app/models/background_check.rb
shaun-technovation: Ah, okay, good call! 👍

app/services/checkr_api_client/api_client.rb
shaun-technovation: 👍

app/controllers/concerns/background_check_controller.rb
```diff
@@ -19,7 +19,7 @@ def show
   end
 
   def international_background_check
-    checkr_result = CustomCheckr::ApiClient.new.request_checkr_invitation("candidates", profile: current_profile)
+    checkr_result = CheckrApiClient::ApiClient.new.request_checkr_invitation(candidate: current_profile)
```
shaun-technovation, line 22: Ah, I think the extra `ApiClient` is unnecessary here - `CheckrApiClient::ApiClient.new`, it could just be `CheckrApiClient.new`, not a blocker,  just IMHO. 👍

app/views/completion_steps/_background_check.html.erb
```diff
@@ -4,9 +4,9 @@
 
 <% if profile.requires_background_check? %>
   <p>
-    <% if profile.background_check.invitation_sent? %>
+    <% if profile.background_check.invitation_pending? %>
```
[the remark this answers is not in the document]
shaun-technovation: Nice update!

shaun-technovation (approved):
LGTM! Some nice updates here! 👍 💯 🎉 :shipit:

PR #4211, opened 2023-09-19
Remove old registration related tests

shaun-technovation:
Removing some old specs related to the "signup" flow and the "signup attempt" flow, neither of which we're using now.


viviancan (approved):
LGTM 💯 Nice cleanup!

PR #4212, opened 2023-09-20
WIP: Registration invite specs

shaun-technovation:


spec/system/admin/registration_invites_spec.rb
```diff
@@ -0,0 +1,34 @@
+require "rails_helper"
```
shaun-technovation, line 1: These look like new specs, but they're just the ones from `invite_users_spec.rb`, I just renamed the file and I made a few tweaks.

viviancan (approved):
LGTM! 💯

PR #4213, opened 2023-09-21
Chapter ambassador judge invites

shaun-technovation:
This will make it so that when a chapter ambassador invites a judge:
- it will use the new registration system
- the judge will be associated to the relevant RPEs (just restoring/moving this functionality to a new file)


app/jobs/update_registration_invite_job.rb
```diff
@@ -0,0 +1,26 @@
+class UpdateRegistrationInviteJob < ActiveJob::Base
```
shaun-technovation, line 1: This looks like a new file, but it's just the `attach_user_invitation_job.rb` file that I renamed, I also updated a few things, like finding the invitation by `admin_permission_token` instead of `account.email`, and made some minor renaming changes.

spec/system/registration/invites_spec.rb
```diff
@@ -3,7 +3,6 @@
 RSpec.describe "Registration invites", :js do
   {
     student: "student",
-    mentor: "mentor",
```
shaun-technovation, line 6: This is related to the spec that's currently failing sporadically on CircleCI, just removing it for now.

viviancan (approved):
💯 I was able to invite a judge from the ChA events page! When I registered the judge option was already preselected. Great job! 🚀

PR #4214, opened 2023-09-21
Add "Club Ambassador" mentor type

shaun-technovation:


viviancan (approved):
I was able to register and select the club ambassador option! I also saw this option as a mentor type filter option in the admin portal 💯 🚀

PR #4217, opened 2023-09-22
Fix mentor type validation during registration

shaun-technovation:
This will make the "As a mentor you may call me a..." field required, making it so that the "Next" button should be disabled if a mentor type isn't selected/picked.


viviancan (approved):
Worked as expected! If I did not select a mentor type I was not able to click the "next" button 💯

Issue #4193, opened 2023-09-13
Add env variable for international background check countries 


viviancan:

viviancan:
Confirming this is on QA and working as expected
I added the following value to heroku QA environment config 
`BACKGROUND_CHECK_COUNTRY_CODES` `IN,CA,US`

I tested a mentor from India and went through the entire process and successfully passed my bg check.
I also logged in as a mentor in the US and Canada and saw the form/button to complete the bg check. When I logged in as a mentor from another country that was not US, IN, or CA I did not have an option to complete a bg check.

So far it seems to be working as expected!! 🎉

note - I had to manually engage the IN candidate through the checkr staging portal

dboyer:
Checked on QA and confirmed that I was prompted for a background check when registered as a mentor from Canada, India, and the United States. I changed the location to France and then did not have to background check.

PR #4220, opened 2023-09-22
Added checkr background check country codes to environment variables

viviancan:
Refs #4193 

Added env variables for background check countries. You will need to add the following to your `.env` file
`BACKGROUND_CHECK_COUNTRY_CODES="US,IN,CA"`. I will also need to add this variable to QA and PROD env variables

With this update, anytime we want to add another country for background checks, we will need to add the 2 letter ISO country code this variable. 

This change also affects the green checkmark for background checks in the side mentor profile menu. It will not be checked if they mentor's account is one of these countries.

### DIFF
--- a/app/models/mentor_profile.rb
+++ b/app/models/mentor_profile.rb
@@ -300,9 +300,14 @@ class MentorProfile < ActiveRecord::Base
   end
 
   def requires_background_check?
-    (account.valid? and account.age >= 18) and
-      (account.country_code == "US" or account.country_code == "IN") and
-        not (background_check.present? and background_check.clear?)
+    (account.valid? && account.age >= 18) &&
+      in_background_check_country? &&
+      !(background_check.present? && background_check.clear?)
+  end
+
+  def in_background_check_country?
+    country_codes = ENV.fetch("BACKGROUND_CHECK_COUNTRY_CODES", "").split(",")
+    country_codes.include?(account.country_code)
   end
 
   def bio_complete?
--- a/app/views/background_checks/_new.html.erb
+++ b/app/views/background_checks/_new.html.erb
@@ -1,5 +1,5 @@
 <% if current_account.country_code == "US" %>
   <%= render "background_checks/us_background_check", url: url %>
-<% elsif current_account.country_code == "IN" %>
+<% else %>
   <%= render "background_checks/international_background_check" %>
 <% end %>
--- a/app/views/mentor/dashboards/show.html.erb
+++ b/app/views/mentor/dashboards/show.html.erb
@@ -9,7 +9,10 @@
       current_teams: TeamSerializer.new(current_mentor.current_teams).serialized_json,
       consent_waiver: ConsentSerializer.new(current_mentor.consent_waiver).serialized_json,
       background_check: BackgroundCheckSerializer.new(current_mentor.background_check).serialized_json,
-      settings: { canDisplayScores: SeasonToggles.display_scores? }.to_json
+      settings: {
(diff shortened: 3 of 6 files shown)


app/views/background_checks/_new.html.erb
```diff
@@ -1,5 +1,5 @@
 <% if current_account.country_code == "US" %>
   <%= render "background_checks/us_background_check", url: url %>
-<% elsif current_account.country_code == "IN" %>
```
viviancan, line 3: Removed this because this view only ever displays if a background check is required which is determined by the `requires_background_check` method. Right now the only condition that decides which version of the background check displays is if they are US or Not

spec/javascript/mentor/store/getters.spec.js
```diff
@@ -1,4 +1,6 @@
-import getters from 'mentor/store/getters'
+import getters from "mentor/store/getters";
+
+const backgroundCheckCountryCodes = "IN, CA, US";
```
viviancan, line 3: I wasn't sure if there was a way to fetch ENV variables here. I searched a bit but didn't want to hold up the code review. Open to suggestions! I will also continue to research.

app/models/mentor_profile.rb
```diff
@@ -300,9 +300,14 @@ def background_check_complete?
...
-    (account.valid? and account.age >= 18) and
-      (account.country_code == "US" or account.country_code == "IN") and
-        not (background_check.present? and background_check.clear?)
+    (account.valid? && account.age >= 18) &&
+      in_background_check_country? &&
+      !(background_check.present? && background_check.clear?)
+  end
+
+  def in_background_check_country?
+    country_codes = ENV.fetch("BACKGROUND_CHECK_COUNTRY_CODES", "").split(",")
```
viviancan, line 309: Decided to go with `BACKGROUND_CHECK_COUNTRY_CODES` which includes US,IN,CA

spec/javascript/mentor/store/getters.spec.js
shaun-technovation: Try `process.env. BACKGROUND_CHECK_COUNTRY_CODES`, but isn't the line above (in `app/views/mentor/dashboards/show.html.erb`) setting the country codes so they can be accessed in the JS world?

app/views/mentor/dashboards/show.html.erb
```diff
@@ -9,7 +9,10 @@
       current_teams: TeamSerializer.new(current_mentor.current_teams).serialized_json,
       consent_waiver: ConsentSerializer.new(current_mentor.consent_waiver).serialized_json,
       background_check: BackgroundCheckSerializer.new(current_mentor.background_check).serialized_json,
-      settings: { canDisplayScores: SeasonToggles.display_scores? }.to_json
+      settings: {
+        canDisplayScores: SeasonToggles.display_scores?,
+        backgroundCheckCountryCodes: ENV.fetch("BACKGROUND_CHECK_COUNTRY_CODES")
```
shaun-technovation, line 14: Isn't this setting the country codes for the JS world? 🤔

shaun-technovation (approved):
LGTM! 🙌 🎉 🥳 👏 :shipit:

spec/javascript/mentor/store/getters.spec.js
viviancan: Yes, that line (`app/views/mentor/dashboards/show.html.erb`) sets the country codes in JS, but this test failed until I added this const. I don't remember the exact error (I can recreate it), but it was something with the backgroundCheckCountryCodes not found.

Issue #4222, opened 2023-09-25
Update international bg check to use sidekiq


viviancan:

viviancan:
The `perform_later` method is on QA

The redirect is almost instant now! I need to update the logic in this view. With the sidekiq background job, I received the email from checkr about 40 seconds later.


[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBckRpIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--5b6df3a90f0c290b0e36d337f952f52f464b598f/CleanShot%202023-09-26%20at%2013.49.59%202.mp4)

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBckhpIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--90d381eff796502a9ae00ecdd8d6a47f73e418dd/CleanShot%202023-09-26%20at%2013.51.20.png)

PR #4227, opened 2023-09-27
Updated background check background check + background check dashboard view

viviancan:
Refs #4222 

This PR has a few changes
1. Added code to create a `Job` record before the`request_checkr_background_check_invitation_job.rb` is run. This was needed to conditionally update the view in `_background_check.html.erb` after an international mentor requests a background check invitation
2. Added new conditional to check for a `queued` job for the current profile. If there is a `queued` job then show the "invitation_job_queued" text. They should receive the background check email from Checkr within 1 min or so.
3. Cleaned up `_background_check.html.erb` by separating the content into partials 

It was a little difficult to test this locally since I haven't been able to get the "perform_later" background job behavior to mimic QA. I did test for the following:
1. US based Mentor
2. India based Mentor
3. Us based ChA
4. India based ChA


shaun-technovation (approved):
Interesting approach! i didn't know we'd be manually creating and manipulating a jobs record, if that was the case it might have been better to create the background check record beforehand, that way all the background check status related stuff would be in one place and it would be consistent with the other background check checking stuff. Anywho, not a blocker or big deal, just side commentary! Didn't try locally, but looks like this will work! 👍 🙌 🚀 :shipit:

app/controllers/mentor/dashboards_controller.rb
```diff
@@ -3,6 +3,7 @@ class DashboardsController < MentorController
     include LocationStorageController
 
     def show
+      @job = current_mentor.jobs.queued.first
```
shaun-technovation, line 6: Random question, is this guaranteed to be the background check invite job?

PR #4229, opened 2023-09-27
Anytime registration invites 

shaun-technovation:
This will allow admins to send anytime invites. And it will allow someone register with an anytime invite (which should also work even if registration is closed).


viviancan (approved):
I turned off all of the registration toggles and invited a ChA, mentor, judge, and student using the "register" at any time checkbox. Everything seems to be working as expected! 💯 🚀 this is a great addition!

PR #4235, opened 2023-10-03
Specs for registration steps and fields 

shaun-technovation:
This will add some basic specs to test registration steps and fields for students, parents, mentors, judges, and chapter ambassadors, the specs will:
- Select the appropriate profile type, click next
- Make sure the appropriate page/step is displayed
  - Fill in the required fields for the selected profile type, click next
- Make sure the data terms page/step is displayed, click next
- Make sure the email, password page/step is displayed

There is no step after this, but I think there should be, like expecting a welcome message or ensuring the account got created. But for some reason the form isn't being submitted to the backend, I verified this by putting some `put` statements in the controller action, but they're not getting called, which afaict it means it's something with the form, Vue Formulate, which I can't figure out, all the fields are being filled out afaict. If you feel like taking a look at this, feel free! Anywho, the specs cover the majority of the registration flow, except for the final submit step.

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -11,7 +11,7 @@ Rails.application.configure do
   config.cache_classes = false
 
   # Do not eager load code on boot.
-  config.eager_load = false
+  config.eager_load = true
 
   # Show full error reports.
   config.consider_all_requests_local = true
--- a/spec/system/registration/invites_spec.rb
+++ b/spec/system/registration/invites_spec.rb
@@ -3,6 +3,7 @@ require "rails_helper"
 RSpec.describe "Using registration invite codes", :js do
   {
     student: "student",
+    mentor: "mentor",
     judge: "judge",
     chapter_ambassador: "chapter ambassador"
   }.each do |profile_type, friendly_profile_type|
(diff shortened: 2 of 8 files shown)


spec/system/registration/invites_spec.rb
```diff
@@ -3,6 +3,7 @@
 RSpec.describe "Registration invites", :js do
   {
     student: "student",
+    mentor: "mentor",
```
shaun-technovation, line 6: Re-enabling these specs, I temporary removed them recently (in e2f1e2c) b/c they were failing randomly on CircleCI, but I have a fix for that now. 😄

config/environments/development.rb
```diff
@@ -11,7 +11,7 @@
   config.cache_classes = false
 
   # Do not eager load code on boot.
-  config.eager_load = false
+  config.eager_load = true
```
shaun-technovation, line 14: This is to address the `Unable to autoload constant` errors when running the specs. More context on this:
https://stackoverflow.com/a/58805153

viviancan (approved):
I am also looking into the issue you mentioned in the main PR but I don't want to hold it up either! Feel free to merge and maybe we will discover something else with the submit step! 💯

PR #4236, opened 2023-10-03
Remove pending team invites to mentors

shaun-technovation:
This will:
* Remove all existing team invites to mentors sent before July 1, 2023
* Update the season reset task to remove all pending team invites to mentors (which doesn't need a date/where clause, since it'll remove all the current/existing pending invites up to the point when the season reset is run)


viviancan (approved):
Did not try locally but LGTM!! 💯

Issue #4247, opened 2023-10-11
Update background check flow so that processing invite logic does not depend on the jobs table


dboyer:
@viviancan to add more info

viviancan:
Confirming this is on QA and working as expected.

I need to review the logic in the mentor debugging section. Something weird is going on with the display values for the invitation status <img width="1017" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNXpwQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--deb137d2eb938e8754a43a3bd653299131fabfa9/Screen%20Shot%202024-01-22%20at%202.30.10%20PM.png" alt="Screen Shot 2024-01-22 at 2.30.10 PM.png" />

dboyer:
I went through the background check process with several mentors and it worked as expected.

PR #4253, opened 2023-10-13
Update httparty gem

shaun-technovation:
This will address this security vulnerability:

- https://github.com/Iridescent-CM/technovation-app/security/dependabot/117

Nothing popped out to me in the [changelog](https://github.com/jnunemaker/httparty/blob/master/Changelog.md), so I think there's not too much risk with this upgrade. Also, it looks like we're only using httparty in one spot, which I don't think is being used now, so we might be able to remove this gem completely, I added this to #4247.


viviancan (approved):
Did not try locally, but seems pretty small! 💯

PR #4256, opened 2023-10-16
WIP: Registration invite updates

shaun-technovation:
A few general updates here related to registration invites:
- Adding the profile type of who sent the invite on the registration invites datagrid
- Adding more info/columns to the registration invites datagrid export CSV
- Add additional/help text for name field on admin registration invites section
- Address "null" name when a ChA removes a judge from an RPE


viviancan (approved):
Did not try locally! Nice improvements 👍

PR #4257, opened 2023-10-17
Added invitation_url to background checks for easier access to checkr portal

viviancan:
CleanShot 2023-10-16 at 20 33 31@2x

This PR adds the `invitation_url` to the background checks table and also adds a link on the background check status page
so that participants may access the checkr portal more directly. They will still receive the email from checkr. Open to wording changes! Refs #4259


shaun-technovation (approved):
LGTM, nice little addition! 👏 🎉 🥳 🚀

Issue #4246, opened 2023-10-10
Registration/invite flow cleanup and removal


shaun-technovation:
- Remove `controllers/signups_controller.rb` and associated cookies
- Check if `email_validations_controller.rb` is being used and remove if necessary
  - Remove httparty and mailgun if we're not using them
- Remove rake task functionality for inviting a chapter ambassador
- Remove unused registration routes/controllers
- Refactor mentor registration API calls for expertises list and mentor types list
  - Move to `API` namespace
  - Rename expertises -> mentor expertises
  - Remove extra `attributes` property for expertises
- Refactor registration settings and invites API calls
  - Rename and move to `api/registration` directory
  - Inherit from APIs controller

## DEPLOYMENT NOTES

For QA, remove Heroku setting/ENV for:
- [x] `COOKIES_ADMIN_PERMISSION_TOKEN`
- [x] `COOKIES_SIGNUP_TOKEN`
- [x] `COOKIES_SIGNUP_WIZARD_MODE`
- [x] `MAILGUN_PRIVATE_KEY`

For Production remove Heroku setting/ENV for:
- [x] `COOKIES_ADMIN_PERMISSION_TOKEN`
- [x] `COOKIES_SIGNUP_TOKEN`
- [x] `COOKIES_SIGNUP_WIZARD_MODE`
- [x] `MAILGUN_PRIVATE_KEY`

shaun-technovation:
This one is one QA. 

Some things to look out for:
- There were some changes to the current registration system regarding API calls for registration settings, invites, mentor types, and mentor skills & interests
- More old signup/invite flow/functionality was removed, shouldn't have any impact, but also something to keep an eye out for

dboyer:
Tested on QA and I was able to:

- Create a mentor with three different mentor types in a country that required a background check
- Create a mentor with different mentor skills in a country that didn't require a background check
- Create a student account and request to join a team
- Received the student invite to join my team and approved it
- As a student on a team, sent an invite to a mentor to create an account and mentor my team
- Received the mentor invite and was able to log in and create an account and saw the invite from the team
- As a ChA, invite an existing judge to judge an event
- As that judge, received the invite and logged in and saw the event once I completed training and signed the consent waiver
- As a ChA, invite a new judge to an event
- As that judge, get an email to register for an account and see the event once registered
- As an admin, send an invite for a mentor to register

viviancan:
Confirming I deleted the following in production
[...]

PR #4260, opened 2023-10-17
Remove signup attempts

shaun-technovation:
This will remove ~99% of the old signup attempts functionality, and the associated functionality in:
- models
- services
- serializers
- the database
- controllers
- routes
- mailers
- locales
- views
- specs
- factories

The last bit of the signup attempts functionality will be removed in #4246.


viviancan (approved):
Wow 🤩 Awesome cleanup here!!

Issue #4198, opened 2023-09-13
Add functionality to handle expired background check invitations 


viviancan:

viviancan:
QA Steps
1. Login to an account that has an expired invitation. These are the accounts on QA that should have an expired bg check invite
```
ophelie+m12023@technovation.org
vivian+testmentorindia927@technovation.org
deb+mentor1124@technovation.org
```
2. Check your background check status
3. You should see text stating that the background check invitation is expired and a button that allows you to request a new background check

dboyer:
On QA, I logged in as deb+mentor1124. I didn't see anything that said I had an expired background check. I did have an option to request a background check and received the email from Checkr right away.

When I tried with the vivian test account, I did see the page saying my invite had expired and I sent a request for a new one.

On the participants page, I added the background check and invitation status columns. It worked when I searched for mentors in India, but I got an error page when using the following parameters: Profile type - mentor, Country - Canada, additional columns - background check, invitation status

viviancan:
@dboyer The participants grid is now updated! You should be able to filter by additional countries with the invitation status!

dboyer:
Confirming on QA that I can now filter by country after adding the background check and invitation status columns. Also confirmed that all data was included in the export.

PR #4263, opened 2023-10-18
Added functionality to handle expired checkr background check invitations 

viviancan:
Refs #4198 

This PR was needed to handle expired checkr background check invitations. If a mentor's invitation is expired there is new button ( "Request a new invitation") they can click to request a new checkr invitation

Main changes include:
- Adding a more general `background_check_invitation_controller` with `background_check_invitation` method
- Removed `international_background_check_method`
- General cleanup of `CheckrApiClient`
      - Added `create_checkr_candidate` and `create_checkr_invitation` method
- The main  `request_checkr_invitation` method will now check for a candidate_id before creating a new checkr candidate. This was added so that if a mentor has an expired invite an new invite can be requested without creating another candidate on the checkr platform
- Added the background check status for all background check country participants in the admin participants grid 

CleanShot 2023-10-17 at 13 08 04@2x

### DIFF
--- a/app/controllers/concerns/background_check_invitation_controller.rb
+++ b/app/controllers/concerns/background_check_invitation_controller.rb
@@ -1,0 +1,8 @@
+module BackgroundCheckInvitationController
+  def background_check_invitation
+    RequestCheckrBackgroundCheckInvitationJob.perform_later(candidate: current_profile)
+
+    redirect_to send("#{current_scope}_dashboard_path"),
+      success: t("controllers.background_checks.invite.success")
+  end
+end
--- a/app/controllers/concerns/background_check_invitation_controller.rb
+++ b/app/controllers/concerns/background_check_invitation_controller.rb
@@ -1,8 +1,24 @@
 module BackgroundCheckInvitationController
-  def background_check_invitation
-    RequestCheckrBackgroundCheckInvitationJob.perform_later(candidate: current_profile)
+  extend ActiveSupport::Concern
+
+  included do
+    before_action :verify_background_check_invitation_required, only: [:create_invitation]
+  end
+
+  def create_invitation
+    RequestCheckrBackgroundCheckInvitationJob.perform_later(account_id: current_account.id)
 
     redirect_to send("#{current_scope}_dashboard_path"),
       success: t("controllers.background_checks.invite.success")
+  end
+
+  private
+
+  def verify_background_check_invitation_required
+    unless current_profile.in_background_check_invitation_country? &&
+        (current_profile.background_check.blank? || current_profile.background_check.invitation_expired?)
+      redirect_to send("#{current_scope}_dashboard_path"),
+                  alert: t("controllers.application.unauthorized")
+    end
   end
 end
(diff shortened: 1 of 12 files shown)


app/services/checkr_api_client/api_client.rb
```diff
@@ -1,6 +1,7 @@
 module CheckrApiClient
   class ApiClient
     def initialize(
+      candidate,
```
shaun-technovation, line 4: Interesting, the `create_checkr_candidate` method takes a `candidate` keyword argument, and this one doesn't. Random suggestion to make them both keyword arguments or positional arguments. 😄

app/services/checkr_api_client/api_client.rb
```diff
@@ -16,55 +17,42 @@ def initialize(
...
-        req.body = {
-          email: candidate.account.email,
-          first_name: candidate.account.first_name,
-          last_name: candidate.account.last_name,
-          dob: candidate.account.date_of_birth,
-          work_locations: [{country: candidate.account.country_code}]
-        }.to_json
+    def request_checkr_invitation
+      if candidate_id.blank?
+        create_checkr_candidate(candidate: candidate)
```
shaun-technovation, line 26: Ah, if I'm understanding/reading this code correctly, it looks like `candidate` is an instance method/reader, in that case it wouldn't need to be passed in as an argument, it can just be referenced as `candidate` in the `create_checkr_candidate`. 🤔

shaun-technovation (approved):
I left some feedback, didn't feel like any blockers though, I think it's fine as-is. Anywho some nice additions here! 👏 🎉 :shipit:

app/services/checkr_api_client/api_client.rb
```diff
@@ -16,55 +17,42 @@ def initialize(
 
       @logger = logger
       @error_notifier = error_notifier
+      @candidate = candidate
+      @candidate_id = candidate.account.background_check.candidate_id
```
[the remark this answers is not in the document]
viviancan: 👍 Made this update! Nice catch

app/services/checkr_api_client/api_client.rb
viviancan: 👍 Made them all keyword arguments!

app/services/checkr_api_client/api_client.rb
viviancan: updated!

app/controllers/concerns/background_check_invitation_controller.rb
```diff
@@ -0,0 +1,8 @@
+module BackgroundCheckInvitationController
+  def background_check_invitation
+    RequestCheckrBackgroundCheckInvitationJob.perform_later(candidate: current_profile)
```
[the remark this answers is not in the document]
viviancan: Didn't know this! I updated to pass the account ID and then in the job found the account by id! Ty 💯

app/controllers/concerns/background_check_invitation_controller.rb
```diff
@@ -0,0 +1,8 @@
+module BackgroundCheckInvitationController
+  def background_check_invitation
```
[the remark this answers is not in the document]
viviancan: 👍 Updated!

Issue #4265, opened 2023-10-19
SUPPORT 10/11/23 - 10/24/23


viviancan:

viviancan:
We received a [notice from Rae](https://iridescentlearning.slack.com/archives/C06CHBQ4E/p1697721214434789) that there were 4 bg check invites shown in the production chekr portal. This was surprising since the only country listed on production that require a bg check right now is the US. After investigating further, we discovered it was possible to manually visit the invitation URL on our platform to request an invitation. A PR has been made to fix this issue. There is also a new rake task to delete background checks by ID so we can delete the background check records for those individuals.

dboyer:
Testing on QA:
- Went to an existing account for a mentor from Nigeria and the dashboard said no background check was needed.
- When logged in as the same mentor, went to mentor/international_background_check and was not able to access the page
- Created a new account for a mentor in Italy and went to mentor/international_background_check and was not able to access the page
- Created a new account for a mentor from India and went to mentor/international_background_check and got a message that I now had an invite from Checkr. Was able to start the background check using the link in the email.
- Created a new account for a mentor from the US and my dashboard showed the need for a background check
- On the same US account, went to mentor/international_background_check and received a message that I could not access that page

viviancan:
This ^^ is on Prod but moving ticket back to backlog until the end of the spring

dboyer:
Closing as access to background checks is now limited to select countries.

PR #4267, opened 2023-10-19
Updates to checkr invitation method so that only accounts that need an invitation can request one

viviancan:
Refs #4265 

We discovered an issue where background invitations could be requested by navigating to the specific invitation route. More [information here](https://iridescentlearning.slack.com/archives/C06CHBQ4E/p1697721214434789)

This PR includes 2 main changes:
- `remove_pending_background_checks_by_id` rake task to remove the bg checks that were requested by accounts that should not have bg checks. 
     - This rake task needs a list of bg check IDs to delete
 - Added a `before_action` for the `international_background_check` method that verifies the current account actually requires a bg check invitation.

Conditions for a background check invitation 
1. The account is in a country the requires invitations (essentially non US countries that we bg check)
2. And that they currently don't have a bg check OR their current invitation is expired 

[QA Steps are here](https://iridescentlearning.slack.com/archives/C02J1GU8VRA/p1697743743936599)

### DIFF
--- a/app/controllers/concerns/background_check_controller.rb
+++ b/app/controllers/concerns/background_check_controller.rb
@@ -1,4 +1,10 @@
 module BackgroundCheckController
+  extend ActiveSupport::Concern
+
+  included do
+    before_action :verify_background_check_invitation_required, only: [:international_background_check]
+  end
+
   def new
     if current_profile.background_check_candidate_id.present?
       redirect_to send("#{current_scope}_background_check_path", id: current_profile.background_check_candidate_id)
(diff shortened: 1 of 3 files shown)


app/models/mentor_profile.rb
```diff
@@ -310,6 +310,12 @@ def in_background_check_country?
     country_codes.include?(account.country_code)
   end
 
+  def in_background_check_invitation_country?
```
viviancan, line 313: Open to suggestions with this method. I needed a way to identify the countries that require the invitation since there is just 1 env variable right now that groups all bg check countries together.

app/controllers/concerns/background_check_controller.rb
```diff
@@ -1,4 +1,10 @@
 module BackgroundCheckController
+  extend ActiveSupport::Concern
+
+  included do
+    before_action :verify_background_check_invitation_required, only: [:international_background_check]
```
shaun-technovation, line 5: I could see this functionality living in the Checkr service too, that way the service itself would be responsible for who can/can't use the service and it's all encapsulated in there; but this works too! 😄

shaun-technovation (approved):
LGTM! 👍 🙌 🚢

PR #4268, opened 2023-10-19
Restore account created activity

shaun-technovation:
This will restore the account created activity that chapter ambassadors used to see when someone in there area registered.


viviancan (approved):
Did not try locally, but looks straightforward! Nice find and readdition to the activity notices!

PR #4272, opened 2023-10-20
WIP: Update follow-redirects module

shaun-technovation:
This will update the follow-redirects module to address these security vulnerabilities:

- https://github.com/Iridescent-CM/technovation-app/security/dependabot/26
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/36


viviancan (approved):
LGTM! Nice updates on all of these!

PR #4273, opened 2023-10-20
Remove old regional ambassador code

shaun-technovation:
This will remove these old redirects (for when we renamed regional ambassador to chapter ambassador back in 2020) that we don't need now.


viviancan (approved):
LGTM 💯

PR #4275, opened 2023-10-23
Setup Pundit and create new parental consents pollicy

shaun-technovation:
This is some work that I did that on Pundit that will:
- Install Pundit and setup the groundwork
- Create a new policy for viewing parental consents
  - Ensuring students can only view their own parental consent


viviancan (approved):
Nice groundwork on this! 

I tried it locally and could not access the parental_consents. I saw this locally `Pundit::NotAuthorizedError (not allowed to show? this ParentalConsent):` Really nice addition 💯

app/controllers/application_controller.rb
```diff
@@ -40,6 +42,10 @@ def current_account
     end
   end
 
+  def pundit_user
+    current_account
+  end
```
shaun-technovation, line 47: This is needed because by default Pundit will look for and use `current_user` if that doesn't exist (like it does for us) then you have to define what `current_user` should be. So far, `current_account` seems to work, but maybe once we write a few more policies this might need to be updated, we'll see. 😄 

Pundit documentation on `pudit_user`:
https://github.com/varvet/pundit#customize-pundit-user

spec/policies/parental_consent_policy_spec.rb
```diff
@@ -0,0 +1,45 @@
+require "rails_helper"
```
shaun-technovation, line 1: I basically followed the examples from the Pundit documentation for these specs:
https://github.com/varvet/pundit#rspec

There is a gem to provide some nice matchers, I tried it, but I couldn't get them working, maybe something to revisit:
https://github.com/punditcommunity/pundit-matchers

PR #4280, opened 2023-10-23
Update Dalli gem

shaun-technovation:
This will update Dalli to version 3.2.6 to address this security vulnerability:

- https://github.com/Iridescent-CM/technovation-app/security/dependabot/101

We're upgrading from 2.x.x to 3.x.x, I looked over the [changelog](https://github.com/petergoldstein/dalli/blob/main/CHANGELOG.md) and saw some breaking changes, but I don't think they concern us. I searched our codebase for `dalli` and I didn't see setting any configuration options.

From what I can tell, Dalli/memcache is being used via `Rails.cache`, so for example, I noticed the submission "percentage complete" functionality uses `Rails.cache`, so to test this locally I updated a submission (add/removing required fields) and saw the percentage calculating/working as expected.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -58,7 +58,7 @@ gem "sidekiq", "~> 7.0"
 
 gem "airbrake", "~> 11.0.3"
 
-gem "dalli", "~> 2.7"
+gem "dalli", "~> 3.0"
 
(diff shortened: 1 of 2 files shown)


viviancan (approved):
💯

PR #4282, opened 2023-10-23
Update Rails to latest version

shaun-technovation:
This will bump Rails to `6.1.7.6` to get the latest fixes, including a security fix for:

- https://github.com/Iridescent-CM/technovation-app/security/dependabot/162

I also had to update `lodash-rails` to resolve some dependency issues.


viviancan (approved):
LGTM! Tried a few things locally 💯

PR #4288, opened 2023-10-27
Add new policy for judge scores

shaun-technovation:
This will add a new policy to ensure that when judges start a new score or resume a score (which both go to the `new` action) that they can only view/access their own scores.

I also refactored the parental consent policy and specs so that the specs are testing the policy against:
- students
- mentors
- judges
- chapter ambassadors
- admins
- guests

And the conditions/rules of the policy will be in the same order^.

### DIFF
--- a/app/controllers/judge/scores_controller.rb
+++ b/app/controllers/judge/scores_controller.rb
@@ -29,6 +29,10 @@ module Judge
     end
 
     def new
+      if params[:score_id].present? && params[:score_id].to_i > 0
+        authorize SubmissionScore.find(params[:score_id]), policy_class: Judge::SubmissionScorePolicy
+      end
+
       respond_to do |f|
         f.html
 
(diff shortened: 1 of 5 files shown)


app/controllers/judge/scores_controller.rb
```diff
@@ -29,6 +29,10 @@ def index
     end
 
     def new
+      if params[:score_id].present? && params[:score_id].to_i > 0
```
shaun-technovation, line 32: This will only enforce this new policy if there is a `score_id` and it's a number.

Sometimes the `score_id` parameter will be the string `null`, and in those cases we don't wan to use this new policy; and sometimes there isn't a `score_id` (like when staring a new score) and those cases we also don't want to use the new policy.

viviancan (approved):
On a separate branch without this update I was able to update the id and get a new submission to score. On this branch, when I tried to update the score ID I got our default error page and saw the pundit `not authorized error` 💯 🚀

PR #4290, opened 2023-10-27
Update Elliptic module

shaun-technovation:
This will update Elliptic (and its dependencies) from version `6.5.3` to version `6.5.4` to address this security vulnerability:

- https://github.com/Iridescent-CM/technovation-app/security/dependabot/3


viviancan (approved):
LGTM!

Issue #4284, opened 2023-10-24
Front-end registration updates for team invites


shaun-technovation:
Background:
- #4283 will make some back-end changes to validate a team invite
- #4245 will make some changes to the API response to include the team invite info

The front-end flow should be updated so that if the team invite is valid, the corresponding registration type (student or mentor) should be enabled and allow the invited person to register.

shaun-technovation:
For mentors who are invited by a student, the mentor profile type will be pre-selected (like an admin registration invite), but if a student is invited, nothing will be pre-selected, there isn't a way to guaranteed way (that I can think of) that will determine whether the person registering would be a parent or student; maybe if we add that (parent/student) as an option when students are sending out team invites, then we could do the pre-selecting.

shaun-technovation:
Another note, when a student or mentor registers via a team invite, the team invite will only get updated if they register with the same email address as the email address the invite was sent to. This is the way admin registration invites used to work too. I was going to address this in my current PR (#4293), but it's already pretty big (over 1000+ lines of code added), and it's standalone enough, so I created this ticket for it: #4308.

shaun-technovation:
The ability to use invites from students to join their team until submission close date is now on QA. The official date range of when they can be used is season start date until submission close date. Heroku config settings would need to be changed to test the date ranges (I think the submission close date is the only important one, I added a beginning date to prevent potential loopholes about using them during another season).

Side note, accepting the team invite as another student is not very intuitive, it's the existing functionality, nothing has changed with it, but maybe something we want to revisit.

Also, I made a bunch of changes in #4245, creating a new service that combines the admin registration settings, admin registration invites, and these new team invites, so if anything is off or needs changing regarding anything registration related, let me know!

dboyer:
@shaun-technovation I'm not sure if any of the above depends on the Heroku variables being correct. I think they updated the dates on the platform but I don't know how to confirm that.

Issue #4245, opened 2023-10-10
Refactor registration/invite API calls


shaun-technovation:
- Right now, we make two separate API calls for registration settings and to validate an invite code. Refactor this code so we only make one API call. The back-end endpoint will do both (validate an invite and return registration settings factoring that in).
    - This is so we can have more dynamic error messages, like mentioned in https://github.com/Iridescent-CM/technovation-app/issues/4157#issuecomment-1748027342.
- Incorporate the new team invite validator service into the API response

PR #4293, opened 2023-11-01
Registration/invite refactor, team invite integration

shaun-technovation:
This is a bundle of work for #4245, #4284, and a bug fix, #4296. Overview of some of the changes:
- Allowing students and mentors to register (via a student team invite) when registration is closed, before the submission deadline
- Updating our two exisiting invite code validation services:
  - `RegistrationInviteValidator` renamed to `RegistrationInviteCodeValidator`
  - `TeamInviteValidator` renamed to `TeamInviteCodeValidator`
  - Both services will now return a success and error message
- A new service called `RegistrationSettingsAggregator` (open to changing the name for this) that will combine the results from:
   - `SeasonToggles` registration settings
   - `RegistrationInviteCodeValidator`
   - `TeamInviteCodeValidator`
 - The new aggregator service does a lot of the validation that the front-end used to do
 - Removed the `Api::Registration::InvitesController` and route, this endpoint and functionality aren't needed now; now we just make one API call to the aggregator that will do all the work


app/controllers/admin/user_invitations_controller.rb
```diff
@@ -11,6 +11,10 @@ def new
     end
 
     def create
+      if user_invitation_params[:profile_type] == "chapter_ambassador"
+        user_invitation_params.merge(register_at_any_time: true)
+      end
```
shaun-technovation, line 16: This is for #4296, on the admin side if someone sends an invite to a chapter ambassador, the "register at any time" checkbox will get automatically checked and disabled; but disabled form elements won't get sent to the back-end, doh! So this change will handle that. 👍

viviancan (approved):
Nice improvements and updates 💯 I think its good to get this on QA since we will have extra testing soon 🚀

PR #4298, opened 2023-11-03
Remove rake task to remove pending mentor invites

shaun-technovation:
This was a run-once rake task that we don't need now, so I'm removing it.


viviancan (approved):
Thanks for remembering this! Nice to remove unneeded things 💯

PR #4300, opened 2023-11-03
Update browserify-sign package

shaun-technovation:
This will update the browserify-sign package and its dependencies to address this security vulnerability:

- https://github.com/Iridescent-CM/technovation-app/security/dependabot/172


viviancan (approved):
💯 👍

PR #4302, opened 2023-11-03
Update tmpl module

shaun-technovation:
This will update the tmpl module to address this security vulnerability:

- https://github.com/Iridescent-CM/technovation-app/security/dependabot/19


viviancan (approved):
Did not try locally but seems like a small upgrade 💯

Issue #4294, opened 2023-11-03
Convert old css to tailwind in submission rebrand


viviancan:

dboyer:
I was editing submissions, etc on the platform and did not see any CSS issues

Issue #4067, opened 2023-06-29
Create `submission_rebrand` layout for student submission process


viviancan:
[Mural boards are here](https://app.mural.co/t/technovationgirls4378/m/technovationgirls4378/1687281291476/2230bb934649c1ec9ee5658134d9079029e01529?sender=u9adab98cc36af2c4665e9477)

- [ ] Create "submission_rebrand" layout file based on the application_rebrand layout. This layout should mimic the submission layout. We still need the submission layout for the mentor experience.
- [ ] Create "helper text" css or reusable component - the light blue background copy from judge portal or honor code page

This ticket is the foundation of all the other work. Once the base layout is complete, we are able to start updating the individual sections/pieces and the side menu bar. 

**Helper text**
![Screen Shot 2023-06-29 at 2.24.57 PM.png](https://images.zenhubusercontent.com/606b0f4fd13d53dff3d14fa5/7623e92f-6e68-4bf9-8a79-9ea491ca3e4e)

PR #4305, opened 2023-11-03
PT2 submission rebrand update menu and sections

viviancan:
PT2 Of #4067 Submission Rebrand Work

This PR includes rebrand for all of the "sections". A section is the overview page like "Ideation", "Pitch", "Start" etc
This PR also includes a few updates to the left menu bar which is the "Submission Checklist". You may notice a few old styles in some of the files. I will convert those styles to tailwind in #4294


shaun-technovation (approved):
Looks good, straight forward to me! Just gotta get the tests to pass, then should be good to go! 👍 🙌 🥳

PR #4306, opened 2023-11-06
Update cross-fetch module

shaun-technovation:
This will update the cross-fetch module to address this security vulnerability:

- https://github.com/Iridescent-CM/technovation-app/security/dependabot/75


viviancan (approved):
👍

Issue #4311, opened 2023-11-08
Add content setting for adding teams to RPE


dboyer:
Add a content setting in the Events section for adding teams to an RPE. 

- ChAs should only be able to add teams to an event if the toggle is on
- If toggle is off, ChAs should be able to remove existing teams since teams drop up to the day of the event and those teams need to go back into the virtual judging pool

Related to #4140. Perhaps need to discuss if these two toggles interact at all.

shaun-technovation:
Hope you don't mind, I created a new ticket for the admin controls (#4312), that work is standalone enough and a good chunk of work on its own I think.

shaun-technovation:
I think this one and #4140 would work similarly to the registration flow - admins enable/disable registration via "Content & Settings" controls, there is an API to get these values that will control the registration profile types. Something similar would need to happen to get the admin settings for the new RPE controls (both registration and RPEs are using Vue.js).

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/4311?planning-poker) with Zenhub @shaun-technovation @viviancan

shaun-technovation:
Well, this is funny, we have a (poorly named) ENV setting, called `MANAGE_EVENTS` (which can be managed in Heroku to see how it works), that will show/hide the Manage Teams button based on this setting. One thing I don't particularly like about this functionality is that you can't see the current teams associated to the event (that I can find) when this setting is off; and I also don't like how it's a hidden, easy to forget about setting! Anywho, I vote we remove this functionality in another ticket. I'l proceed with the work here, I'll make the "Add Teams" button dynamically shown/hidden vs the entire section like the ENV does.

shaun-technovation:
Also, regarding `MANAGE_EVENTS`, someone couldn't remove a team (part of the work in this issue) when this setting is off. Another reason why I think we should remove it.

shaun-technovation:
I did the work for this one in the [PR](https://github.com/Iridescent-CM/technovation-app/pull/4322) for #4140, there was too much overlap for two separate PRs.

shaun-technovation:
This one is on QA, if the corresponding admin control to prevent new teams from being added to an RPE (in #4312) is disabled, then the "ADD TEAMS" button should be hidden.

dboyer:
Testing on QA:

[...]

Issue #4140, opened 2023-08-08
Toggle ability for ChAs to create an RPE based on admin setting


dboyer:
A new admin setting to enable creating RPEs will be added in #4312, use this new setting so that:
- ChAs should only be able to create events if the toggle is on
- If toggle is off, ChAs should be able to still edit existing events
- If toggle is off, ChAs should be able to delete existing events (this is not necessary if it's technically challenging)

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/4140?planning-poker) with Zenhub @shaun-technovation @viviancan

shaun-technovation:
This one is on QA, if the corresponding admin control (done in #4312) to prevent new RPES from being created is disabled, then the "Add an event" button should be hidden.

dboyer:
Testing on QA and confirmed:

- When the toggle was on, I was able to create a new event, edit an existing event details, add teams to an existing event
- When the toggle was off, I was not able to create a new event, edit an existing event, add teams to an existing event, delete an existing event. The button to add an event was not displayed.

dboyer:
Tested by PT and no concerns found

PR #4322, opened 2023-11-09
Dynamically display buttons for RPE functionality

shaun-technovation:
This is work for both #4140 and #4311, which will:
- Dynamically display the "Add an event" button based on the new/corresponding admin "Content & Settings" control
- Dynamically display the "ADD TEAMS" button based on the new/corresponding admin "Content & Settings" control

The other open PR, #4314 is for the admin controls which the work here relies on.


viviancan (approved):
💯 Awesome work on this!

PR #4324, opened 2023-11-14
Add team name to email mentors receive when one of their team is removed from an RPE

shaun-technovation:
This will make it so that when a mentor's team is removed from an RPE, the email the mentor receives will include the team name in it.


viviancan (approved):
Did not try locally, but looks pretty straightforward! Nice addition 💯

PR #4326, opened 2023-11-15
Remove global invitations

shaun-technovation:
This was an old piece of functionality that was used for the old signup flow, from what I can tell, anyone who had a global invite code (there were just two of them defined) could register at anytime (using the old signup flow).


viviancan (approved):
💯 Lgtm! Nice cleanup 👍

PR #4330, opened 2023-11-16
Limit number of judge recusals

shaun-technovation:
A new config var `JUDGE_MAXIMUM_NUMBER_OF_RECUSALS` will be used to limit the number of recusals a judge can make. If they go over this limit, a popup will be displayed with an info message.


viviancan (approved):
Nice update 💯 Worked as expected once I had 15 recusals!

PR #4332, opened 2023-11-17
Updated old css to tailwind in submissions files

viviancan:
Refs #4294 Updated the remaining CSS classes in the submissions flow to tailwind

### DIFF
--- a/app/javascript/stylesheets/layout.css
+++ b/app/javascript/stylesheets/layout.css
@@ -49,3 +49,11 @@ html{
 .tw-flag-magenta{
     @apply text-white block p-2 text-right text-base font-bold shadow-sm bg-tg-magenta
 }
+
+.tw-hint {
+    @apply mb-4 italic text-base;
+}
+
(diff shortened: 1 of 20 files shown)


app/javascript/stylesheets/layout.css
```diff
@@ -48,4 +48,12 @@ html{
 
 .tw-flag-magenta{
     @apply text-white block p-2 text-right text-base font-bold shadow-sm bg-tg-magenta
+}
+
+.tw-hint {
```
shaun-technovation, line 53: Random comment on the `tw` prefixes and naming things in general, it's good practice to hideaway the implementation details. Our libraries change out over time and if the functionality is wrapped generically, the wrapper doesn't need to change when our libraries change. It's not likely Tailwind will change out anytime soon (I'm sure that was the thought when Bootstrap was implemented too!), but good practice anytime we're using/wrapping 3rd party libraries. 😃

PR #4338, opened 2023-11-21
Add judge recusal count to recusal popup

shaun-technovation:
This will add the recusal count to the recusal popup in the footer, displaying text like: 1 out of 15 recusals used

I also updated the recusal conditional logic, the logic before would let a judge use 16 recusals (maximum +1), this will cut them off at 15 (or whatever is defined as the maximum).


viviancan (approved):
Did not try locally! Nice addition to the recusals 💯

PR #4341, opened 2023-11-21
Update beginner scoring scale from 1-3 to 1-5

shaun-technovation:
This will update the beginner division scoring scale, so it goes from 1-5, instead of 1-3.

This will also update the score descriptions for the beginner division, so they match the Jr/Sr division score descriptions.


viviancan (approved):
Did not try locally, but LGTM! 💯 🚀

PR #4342, opened 2023-11-22
RPE scores should not be marked as suspicious

shaun-technovation:
This will make RPE scores (aka "live" scores) be marked as non-suspicious scores.


viviancan (approved):
LGTM! Tried locally and works as expected. RPE scores would not be marked completed too fast. Virtual scores would still be marked as too fast and after 2 suspicious scores the score would show up in suspicious scores.

PR #4343, opened 2023-11-22
Add event type filter to scores export datagrid

shaun-technovation:
This will add an event type filter (either "live" or "virtual") to the scores export datagrid, allowing someone to view/download only "live" or only "virtual" scores.


viviancan (approved):
LGTM 💯 🚀

PR #4348, opened 2023-11-22
Add score details link to scores export datagrid

shaun-technovation:
This will add a link to the scores detail page on the scores export datagrid.

I couldn't test the link on the the CSV export, that doesn't work for me locally, so we'll have to test that once it get to QA.

### DIFF
--- a/app/data_grids/scores_grid.rb
+++ b/app/data_grids/scores_grid.rb
@@ -136,7 +136,7 @@ class ScoresGrid
   end
 
   column :view, html: true do |submission_score|
-    link_to(
+    html = link_to(
       web_icon("list-ul", size: 16, remote: true),
       send("#{current_scope}_score_path", id: submission_score.id),
       data: {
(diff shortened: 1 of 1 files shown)


app/data_grids/scores_grid.rb
```diff
@@ -136,13 +136,29 @@ class ScoresGrid
...
       data: {
         turbolinks: false
       },
       class: "view-details"
     )
+
+    html += " "
+    html += link_to(
+      web_icon("list-alt", size: 16, remote: true),
+      send("#{current_scope}_score_detail_path", id: submission_score.team_submission.id),
```
viviancan, line 151: I was able to get the export from the s3 bucket and noticed the scores detail link was going to the team submission id and not the score id! I think this should be updated to the score id to see the score details 🙂 Besides that the new link did display in the export! 💯 👍 
CleanShot 2023-11-27 at 15 12 14

shaun-technovation: Awesome, thanks for this! Good catch! 💯 Saves me/us finding, fixing and deploying another time! 😊 👍 👏

PR #4349, opened 2023-11-28
Update avatars on registration page

shaun-technovation:
This will update the avatars on the registration page to the ones given to us.


app/javascript/new_registration/components/StepOne.vue
```diff
@@ -151,35 +151,35 @@ export default {
     },
     studentProfileType() {
       return {
-        label: `<img src="${require('signup/myTG-student.png')}" alt="" class="student"> <span class="student s1-label-text">I am registering myself and am 13-18 years old*</span>`,
+        label: `<img src="${require('signup/student.png')}" alt="" class="student mb-2"> <span class="student s1-label-text">I am registering myself and am 13-18 years old*</span>`,
```
viviancan, line 154: Should we add alt text for accessibility? Something like `student profile avatar` `judge profile avatar`

viviancan (approved):
Did not try locally, but LGTM! 💯 🚀

app/javascript/new_registration/components/StepOne.vue
shaun-technovation: Nice suggestion, there wasn't `alt` text there originally and I thought it was intentional!

From [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#alt):
> Setting this attribute to an empty string (alt="") indicates that this image is not a key part of the content (it's decoration or a tracking pixel), and that non-visual browsers may omit it from rendering. Visual browsers will also hide the broken image icon if the alt attribute is empty and the image failed to display.

PR #4356, opened 2023-12-01
Update color-string package

shaun-technovation:
This will update the `color-string` package to address this security vulnerability:

- https://github.com/Iridescent-CM/technovation-app/security/dependabot/15


viviancan (approved):
Did not try locally, but LGTM! Awesome work keeping up with all the security updates 💯

PR #4362, opened 2023-12-05
Update score details link

shaun-technovation:
This will update the score details link to use the team submission id (instead of the score id).


shaun-technovation:
Just a simple one-line update here, just merging this one straight into QA.

PR #4364, opened 2023-12-05
Update error text for uploading consent forms

shaun-technovation:
This will add the `max_file_size` that can be uploaded to the error message gets displayed when someone uploads a parental and media consent that are too large.


viviancan (approved):
Did not try locally, but LGTM!

Issue #4331, opened 2023-11-17
Update RPE slide uploads styles (rebrand)


viviancan:

dboyer:
I see the new branding on QA for slides and it worked as expected. I uploaded slides for a team who did not have any previously and was a bit loss as to what to do after I uploaded the slides. Any ideas? My only thought was to add a "Return to dashboard" button.

<img width="990" alt="Screen Shot 2023-12-03 at 9 11 56 PM" src="https://github.com/Iridescent-CM/technovation-app/assets/516764/d428d103-e4eb-4b00-9747-ba4773f89637">

dboyer:
Saw the random RPE page on submissions again! I had the content settings to judging off, submissions on, and adding teams to events on (but creating and selecting RPEs was toggled off). I clicked from my dashboard into my submission. It took me to this page. It would be good to remove the mention of no events due to covid since that is no longer true.

<img width="2000" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN1V2QWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--f0855f24dee58844f1c4ed12e513bd7613b77121/Screen%20Shot%202023-12-04%20at%209.26.52%20PM.png" alt="Screen Shot 2023-12-04 at 9.26.52 PM.png" />

viviancan:
@dboyer I removed COVID refs and also added a dashboard btn when submissions are not editable
<img width="1098" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNEkwQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--88a15f0007a081fcb16d1da6304487221cec4376/CleanShot%202023-12-05%20at%2012.23.25.png" alt="CleanShot 2023-12-05 at 12.23.25.png" />

dboyer:
I was not able to trigger this message on any of the submissions I tested.

PR #4365, opened 2023-12-05
Added back to dashboard btn for RPE slide uploads + removed covid references 

viviancan:
Refs #4331 

Updates based on QA feedback
- Added back to dashboard btn for RPE slide uploads 
- Removed covid references


shaun-technovation (approved):
👍

PR #4378, opened 2023-12-06
WIP: Allow students to send invites until submissions close

shaun-technovation:
This will:
- let students send invites to both students and mentors until submissions close
- let students view all their pending invites until submissions close


shaun-technovation:
Closing this PR since it's not needed now, we decided that sending invites should be tied to the "Forming teams allowed" admin setting; basically the way it behaves now.

PR #4380, opened 2023-12-08
Salesforce auth token refresh and caching

shaun-technovation:
These are restforce configuration changes will allow us to refresh and cache the Salesforce auth token.

If the `oauth_token` is expired, restforce will use the `refresh_token` to get another `oauth_token` which we will cache.

So, the flow is:
1. Use cached `oauth_token`
2. If expired, get a new one
3. Cache new `oauth_token`
4. Repeat

### DIFF
--- a/app/services/salesforce/api_client.rb
+++ b/app/services/salesforce/api_client.rb
@@ -1,10 +1,16 @@
 module Salesforce
   class ApiClient
     def initialize(
-      url: ENV.fetch("SALESFORCE_INSTANCE_URL"),
-      access_token: ENV.fetch("SALESFORCE_ACCESS_TOKEN"),
-      version: ENV.fetch("SALESFORCE_API_VERSION"),
       enabled: ENV.fetch("SALESFORCE_ENABLED", false),
+      instance_url: ENV.fetch("SALESFORCE_INSTANCE_URL"),
+      api_version: ENV.fetch("SALESFORCE_API_VERSION"),
+      client_id: ENV.fetch("SALESFORCE_CLIENT_ID"),
+      client_secret: ENV.fetch("SALESFORCE_CLIENT_SECRET"),
+      refresh_token: ENV.fetch("SALESFORCE_REFRESH_TOKEN"),
+      oauth_token: Rails.cache.fetch(:salesforce_access_token),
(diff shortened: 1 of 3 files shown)


app/services/salesforce/api_client.rb
```diff
@@ -1,22 +1,31 @@
...
+      authentication_callback: proc do |response|
+        Rails.cache.write(:salesforce_access_token, response["access_token"])
+      end,
       client_constructor: Restforce,
       logger: Rails.logger,
       error_notifier: Airbrake
     )
 
-      if enabled
+      if ActiveModel::Type::Boolean.new.cast(enabled)
```
shaun-technovation, line 19: This simple change will allow us to use `0` or `false` (for `SALESFORCE_ENABLED`) to disable the Salesforce integration.

https://stackoverflow.com/a/60769537

app/services/salesforce/api_client.rb
```diff
@@ -1,29 +1,33 @@
...
-          instance_url: url,
-          api_version: version
-        )
-
-        @logger = logger
-        @error_notifier = error_notifier
-      else
-        logger.info "[SALESFORCE DISABLED] Trying to initialize Salesforce API client"
-
-        raise "Salesforce is disabled"
```
shaun-technovation, line 25: I made a revision so that we'll just log an error instead of logging and raising an error, for when Salesforce is disabled.

viviancan (approved):
💯 Nice work on this!

app/services/salesforce/api_client.rb
```diff
@@ -1,35 +1,39 @@
...
+      @salesforce_enabled = ActiveModel::Type::Boolean.new.cast(enabled)
+      @logger = logger
+      @error_notifier = error_notifier
     end
 
     def add_contact(account:)
       salesforce_id = nil
 
-      handle_request do
+      handle_request "Adding account #{account.id}" do
```
shaun-technovation, line 36: @viviancan Just wanted to document here, that I added this functionality after you approved this PR. I copied this functionality from the Mailchimp integration. So we get better logging messages for when Salesforce is disabled, so it'll display something like:

`[SALESFORCE DISABLED] Adding account xyz`

Issue #4346, opened 2023-11-22
Add judge onboarding status to admin RPE page


dboyer:
Currently admins only see the judges that are part of an RPE, not their ready/onboarding status. Add a column showing judge ready/onboarding status to admin event details page (i.e. /admin/events/209)

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeUQrQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--23a4b5d4753c4cf20c2bed73563d4c0465f48109/Screen%20Shot%202023-11-22%20at%201.44.29%20PM.png)

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/4346?planning-poker) with Zenhub @shaun-technovation @viviancan

dboyer:
Program Team confirmed that this would be useful so they do not need to log in as the ChA to view the info. We can proceed with this feature and do not need to add the ability for admins to add or delete judges from the admin page. They will log in as the ChA if that is necessary.

viviancan:
Confirming this is on QA

dboyer:
I also see the judge status in the new table on the RPE admin pages and the status updates as expected.

Issue #4345, opened 2023-11-22
Add team status to admin view of RPEs


dboyer:
Currently admins only see the teams that are part of an RPE, not their submission status. Add a column showing team submission status to admin event details page (i.e. /admin/events/209)

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeDMrQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--f63724d0ee28fb0ded3f426e29eaa8aabf69387b/Screen%20Shot%202023-11-22%20at%201.42.17%20PM.png)

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/4345?planning-poker) with Zenhub @shaun-technovation @viviancan

dboyer:
Program Team confirmed that this would be useful so they do not need to log in as the ChA to view the info. We can proceed with this feature and do not need to add the ability for admins to add or delete teams from the admin page. They will log in as the ChA if that is necessary.

viviancan:
Confirming this is on QA

dboyer:
Tested on QA and confirming that I also see the team status in the new table.

<img width="1067" alt="Screen Shot 2023-12-11 at 11 25 07 AM" src="https://github.com/Iridescent-CM/technovation-app/assets/516764/c0bdb7a6-687f-4db0-9372-2c1404d83864">

PR #4381, opened 2023-12-08
Admin RPE/event details page updates

viviancan:
Refs #4345 and #4346 

This PR includes changes the Admin rpe/event details page
- Added a team and judge table with additional columns similar to the ChA rpe management page
- Added a new `pill` class for the team and judge status
- I also discovered a tiny bug where invited judges would not be displayed in the judge list so I updated that
- A few additional tests

CleanShot 2023-12-08 at 08 29 12@2x

CleanShot 2023-12-08 at 08 29 05@2x

CleanShot 2023-12-07 at 20 08 33@2x


shaun-technovation (approved):
Didn't try locally, but LGTM! 👍 🙌 💯 🚢

PR #4388, opened 2023-12-14
Add score range to scores export datagrid

shaun-technovation:
This will add the score range (max/min score difference) to the scores export datagrid, for both quarterfinal scores and semifinal scores.


viviancan (approved):
LGTM! 💯

PR #4389, opened 2023-12-16
PT2 Updated bg check flow to save and display (to admin) any error messages from Checkr 

viviancan:
Refs #4247 

This is pt 2 of bg check flow updates.

- This PR adds an error message field to the background checks table
- The error message is saved and is displayed in the admin portal under the mentor debugging section

CleanShot 2023-12-15 at 20 25 13@2x

I reviewed all of the errors for international BG checks on Checkr production portal. There are only 3 unique account errors. All 3 of them are related to the alpha count of the last name.

### DIFF
--- a/app/services/checkr_api_client/api_client.rb
+++ b/app/services/checkr_api_client/api_client.rb
@@ -23,7 +23,15 @@ module CheckrApiClient
 
     def request_checkr_invitation
       if candidate_id.blank?
-        create_checkr_candidate
+        candidate_resp = create_checkr_candidate
+        candidate_resp_body = JSON.parse(candidate_resp.body, symbolize_names: true)
+
+        if candidate_resp.success?
+          @candidate_id = candidate_resp_body[:id]
+        else
+          error = candidate_resp_body[:error].join(", ")
+          handle_error(error)
+        end
       end
 
       if candidate_id.present?
--- a/app/services/checkr_api_client/api_client.rb
+++ b/app/services/checkr_api_client/api_client.rb
@@ -27,36 +27,26 @@ module CheckrApiClient
       end
 
       if candidate_id.present?
-        invitation_response = create_checkr_invitation(
+        invitation_resp = create_checkr_invitation(
           candidate_country_code: candidate.country_code,
           candidate_id: candidate_id
         )
 
-        if invitation_response.success?
+        invitation_resp_body = JSON.parse(invitation_resp.body, symbolize_names: true)
+
+        if invitation_resp.success?
           candidate.background_check.update(
             candidate_id: candidate_id,
-            invitation_id: invitation_response.payload[:id],
-            invitation_status: invitation_response.payload[:status],
-            invitation_url: invitation_response.payload[:invitation_url],
-            internal_invitation_status: :invitation_sent
+            invitation_id: invitation_resp_body[:id],
+            invitation_status: invitation_resp_body[:status],
+            invitation_url: invitation_resp_body[:invitation_url],
+            internal_invitation_status: :invitation_sent,
+            error_message: nil
           )
-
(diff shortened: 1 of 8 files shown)


db/migrate/20231216014610_add_error_message_to_background_checks.rb
```diff
@@ -0,0 +1,5 @@
+class AddErrorMessageToBackgroundChecks < ActiveRecord::Migration[6.1]
```
viviancan, line 1: Open to feedback on this! Not sure if we would want to track all error messages kind of like a history of them (essentially adding an error message table?) Or is this approach ok... do we just need to know if there is currently an error with the bg check but that's all? 🤔 ❓

db/migrate/20231216014610_add_error_message_to_background_checks.rb
```diff
@@ -0,0 +1,5 @@
...
+  def change
+    add_column :background_checks, :error_message, :string
```
shaun-technovation, line 3: Hmm, I think `:string` is fine here, but we don't have any control of what Checkr would send us either. Also, most (all?) of our free-form style text fields are `:text` vs `:string` data types. Anywho, it's probably fine the way it is, but just flagging it just in case.

app/views/admin/participants/_mentor_debugging.html.erb
```diff
@@ -47,9 +47,14 @@
                   "exclamation-circle icon-red",
                   text: profile.background_check.invitation_status
                 ) %>
+        <% elsif  profile.background_check.error? %>
```
shaun-technovation, line 50: Minor, extra spacing^

app/services/checkr_api_client/api_client.rb
```diff
@@ -23,40 +23,38 @@ def initialize(
...
           candidate.background_check.update(
             candidate_id: candidate_id,
-            invitation_id: invitation_response.payload[:id],
-            invitation_status: invitation_response.payload[:status],
-            invitation_url: invitation_response.payload[:invitation_url],
-            state: :invitation_sent
+            invitation_id: invitation_resp_body[:id],
+            invitation_status: invitation_resp_body[:status],
+            invitation_url: invitation_resp_body[:invitation_url],
+            state: :invitation_sent,
```
shaun-technovation, line 51: Also, re: last PR and naming, reading this with just `state` makes me think it could be more descriptive. I think adding `invitation` would help to signify it's for invitations (and perhaps one of the values could be `sent`?). Also, it really feels like it's just a `status`, but of course `invitation_status` is taken, but if `invitation_id`, `invitation_status` and `invitation_url` are implied to be Checkr specific. This could be explicitly named to let us know this is our `internal_invitation_status`. Just random commentary. 😄

shaun-technovation (approved):
LGTM, didn't try it locally, left some FWIW feedback, no blockers, ultimately I think this will move us in the right direction and surfacing the error messages from Checkr will be a big win for us! 💯 🙌 🎉 🚢

app/services/checkr_api_client/api_client.rb
```diff
@@ -23,40 +23,38 @@ def initialize(
...
+
+        if candidate_resp.success?
+          @candidate_id = candidate_resp_body[:id]
+        else
+          error = candidate_resp_body[:error].join(", ")
+          handle_error(error)
+        end
       end
 
       if candidate_id.present?
```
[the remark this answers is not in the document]
viviancan: If the invitation is expired, the candidate_id would be present already. If the invitation is expired, they would have a candidate_id and an invitation_status of `expired`. In this situation an additional checkr candidate would not be created, only a new checkr invitation request would be made.

PR #4390, opened 2023-12-16
PT1: Updates to the background check flow so that the initial flow no longer depends on a running background job

viviancan:
Refs #4247 

The PR includes the updates so that the initial background check flow (when Checkr is processing the invitation) does not depend on the background jobs record.
Changes include
- Creating a "placeholder" background check record in the background check job before the API call happens
- Changed the `invitation_sent` status value to `invitation_required`
- Added `state` field to background checks table - State can be`requesting_invitation invitation_sent error`
- Added a background check invitation status section to the mentor debugging portion in the admin portal. This is also updated in pt 2 of the PR to display the actual error message from Checkr for participant support requests 
- Removed the not null candidate id constraint in the background check table
- A new "error" view with a "request new invitation" button

CleanShot 2023-12-15 at 20 25 03@2x


app/models/background_check.rb
```diff
@@ -1,7 +1,8 @@
 class BackgroundCheck < ActiveRecord::Base
   # If these change, you will need to update dataclips
-  enum status: %i[pending clear consider suspended canceled invitation_sent]
+  enum status: %i[pending clear consider suspended canceled invitation_required]
```
viviancan, line 3: I felt this value was necessary because this status field represents the background check status. When an invitation is being processed or completed by the participant, the background check hasn't really started yet. We also don't have a bg check report ID yet so the other values did not seem appropriate to describe the status of the bg check

app/services/checkr_api_client/api_client.rb
```diff
@@ -121,5 +125,11 @@ def post(resource, body)
         req.body = body.to_json
       end
     end
+
+    def update_state(state)
```
viviancan, line 129: This method is removed in pt 2 - I created a handle_error method instead

db/migrate/20231215163819_add_state_to_background_checks.rb
```diff
@@ -0,0 +1,5 @@
+class AddStateToBackgroundChecks < ActiveRecord::Migration[6.1]
```
viviancan, line 1: I needed to remove the not null here because when the placeholder bg check is created before the API call, there is not a candidate ID yet

app/jobs/request_checkr_background_check_invitation_job.rb
```diff
@@ -10,6 +10,12 @@ class RequestCheckrBackgroundCheckInvitationJob < ActiveJob::Base
       status: "queued",
       owner: profile
     )
+
+    bg_check = BackgroundCheck.find_or_create_by!(account: profile)
```
viviancan, line 14: This is where the placeholder record is created before the API call

app/models/background_check.rb
```diff
@@ -1,7 +1,8 @@
...
   enum invitation_status: %i[pending completed expired], _prefix: :invitation
+  enum state: %i[requesting_invitation invitation_sent error]
```
viviancan, line 5: Open to naming here! Initially I went with "stage" but that didn't feel right because it seemed that it represented a sequence of steps... 🤔 💭

shaun-technovation: Ah, I see that we're already using `invitation_status`, hmm, I could see this being `internal_invitation_status` 🤔 And perhaps to more explicit, renaming `invitation_status` to `checkr_invitation status`. Just thinking out loud here. 😄

app/jobs/request_checkr_background_check_invitation_job.rb
```diff
@@ -10,6 +10,12 @@ class RequestCheckrBackgroundCheckInvitationJob < ActiveJob::Base
...
+    bg_check.update_columns(
```
shaun-technovation, line 15: The naming here, `bg_check` vs `background_check`, sometimes I see `bg_check` and others it's `background_check`.

app/views/completion_steps/_background_check.html.erb
```diff
@@ -3,16 +3,16 @@
 </h1>
 
 <% if profile.requires_background_check? %>
-  <% if @job.present? && @job.status == "queued" %>
-    <%= render "background_checks/statuses/invitation_job_queued" %>
+  <% if profile.background_check.requesting_invitation? %>
+    <%= render "background_checks/statuses/requesting_invitation" %>
+  <% elsif profile.background_check.error? %>
+    <%= render "background_checks/statuses/invitation_error" %>
```
shaun-technovation, line 9: This conditional is interesting, if there is an `error?` display the `invitation_error` partial. If the errors are specific to invitations, or if they're just general errors I would think the naming should align. 🤔

db/migrate/20231211204753_remove_not_null_candidate_id_constraint_from_background_checks.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveNotNullCandidateIdConstraintFromBackgroundChecks < ActiveRecord::Migration[6.1]
+  def change
+    change_column_null :background_checks, :candidate_id, true
```
shaun-technovation, line 3: Neat @ `change_column_null`, that's new to me!

app/controllers/concerns/background_check_invitation_controller.rb
```diff
@@ -16,7 +16,7 @@ def create_invitation
 
   def verify_background_check_invitation_required
     unless current_profile.in_background_check_invitation_country? &&
-        (current_profile.background_check.blank? || current_profile.background_check.invitation_expired?)
+        (current_profile.background_check.blank? || current_profile.background_check.invitation_expired? || current_profile.background_check.error?)
```
shaun-technovation, line 19: This pops out to me as a potential refactoring opportunity, the basic conditional is "unless this and (this or this or this)", seems like there should be a better way to express that. 🤔 😅

app/null_objects/null_background_check.rb
```diff
@@ -10,5 +10,9 @@ def suspended?; false; end
   def paranoid?; false; end
   def invitation_status; nil; end
   def invitation_pending?; false; end
+  def requesting_invitation?; false; end
+  def invitation_completed?; false; end
+  def invitation_expired?; false; end
```
shaun-technovation, line 15: A very minor thing, but looks like the invitation status from Checkr can be pending, completed or expired, I think it would be nice to have those statuses grouped here ☝️ 😄

app/services/checkr_api_client/api_client.rb
```diff
@@ -33,26 +33,29 @@ def request_checkr_invitation
...
         if invitation_response.success?
-          candidate.update(background_check_attributes: {
+          candidate.background_check.update(
             candidate_id: candidate_id,
             invitation_id: invitation_response.payload[:id],
             invitation_status: invitation_response.payload[:status],
             invitation_url: invitation_response.payload[:invitation_url],
-            status: :invitation_sent
-          })
+            state: :invitation_sent
```
shaun-technovation, line 41: Just based on this line, and the ones below (like `update_state`), I think it would be clearer (re: naming) if at least it mentioned "invitation", since it's not a general state, it's very specific for invitations.

shaun-technovation (approved):
Left some renaming, food for thought type of feedback, but overall I think it looks good and will move us in the right direction, so looks good to me as is or w/ a few changes. Either way 👍 🙌 🚀 :shipit:

app/views/completion_steps/_background_check.html.erb
viviancan: Updated to just "error" in case we can capture other errors too

PR #4398, opened 2023-12-21
Code formatting cleanup using Standard

shaun-technovation:
I ran the command `standardrb --fix` on our codebase which will fix all the "safe" formatting violations that won't change any behavior.

From the [Standard README](https://github.com/standardrb/standard#fixing-errors):
> Because we're confident Standard's fixes won't change the behavior of our code...

We will still have 200+ violations that can't safely be fixed with the `--fix` flag, we'll have to fix those manually or use the `--fix-unsafely` flag.


shaun-technovation:
I merged this one in, since it's just (according to the README) code formatting changes and not behavior changes. 👍

PR #4400, opened 2024-01-08
Add "Partners" link to footer

shaun-technovation:
This will add "Partners" to our footer, only for our application rebrand (Rails and Vue.js), it doesn't update the old layout.


viviancan (approved):
Did not try locally! 💯

PR #4410, opened 2024-01-10
Track account activity

shaun-technovation:
This PR will:
- Update the `public_activity` gem
- Let us know the account id of the account that is impersonating
- Track who made changes to an account and what those changes were
- Add a new section to the participants show page in the admin to display the participant's account changes

Also, noting that this is just rough/proof of concept functionality and that we can definitely iterate and improve on it.


app/controllers/admin/participant_sessions_controller.rb
```diff
@@ -1,5 +1,8 @@
 module Admin
   class ParticipantSessionsController < AdminController
```
shaun-technovation, line 2: The changes in this file are for letting us know which account id is impersonating.

app/controllers/concerns/profile_controller.rb
```diff
@@ -6,12 +6,18 @@ module ProfileController
   end
 
   def update
+    account_performing_update = if session[:admin_account_id_performing_impersonation].present?
+      Account.find(session[:admin_account_id_performing_impersonation])
+    else
+      current_account
+    end
```
shaun-technovation, line 13: This will let us know which account is updating an account.

app/technovation/profile_updating.rb
```diff
@@ -1,22 +1,28 @@
 class ProfileUpdating
```
shaun-technovation, line 1: A few changes in this file are to address StandardRB formatting issues.

app/technovation/profile_updating.rb
```diff
@@ -1,22 +1,28 @@
...
-  private
-
-  attr_reader :profile, :account, :scope
-
-  public
-
-  def initialize(profile, scope = nil)
+  def initialize(profile, scope = nil, account_performing_update: profile.account)
```
shaun-technovation, line 2: `ProfileUpdating` needs to know who is performing the update, so we can track it the activity.

app/technovation/profile_updating.rb
```diff
@@ -1,22 +1,28 @@
...
-  def self.execute(profile, scope = nil, attrs)
-    new(profile, scope).update(attrs)
+  def self.execute(profile, scope = nil, attrs, account_performing_update: profile.account)
+    new(profile, scope, account_performing_update: account_performing_update).update(attrs)
   end
 
   def update(attributes)
     if profile.update(attributes)
+
+      profile.account.create_activity(
```
shaun-technovation, line 16: This is where we're tracking who made the changes and what changes were made. I had to move the `create_activity` call here because if it's elsewhere the `saved_changes` aren't available for us to access.

app/views/public_activity/account/_update.en.html.erb
```diff
@@ -1,33 +1,71 @@
...
-        send("#{current_scope}_participant_path", activity.trackable),
-        data: { turbolinks: false } %>
-    </td>
-  <% else %>
-    <td>–</td>
-  <% end %>
-
-  <td><%= time_ago_in_words(activity.created_at) %> ago</td>
-</tr>
+<% if current_account.is_admin? %>
```
shaun-technovation, line 1: This is how/where we're displaying the account update activity. I inlined all of the CSS, my thought was that when this switches over to Tailwind, the Tailwind classes can be used instead of the inline CSS. I also retained the old way of displaying this activity b/c I'm pretty sure ChAs use this partial too.

viviancan: Nice idea for our "vision platform!"

viviancan (approved):
Tried it locally and I think it is working as expected! I think it's a great start for this new functionality 💯 

The only feedback I have is to consider adding helper text when there isn't any tracked activity.

CleanShot 2024-01-12 at 10 47 27@2x

shaun-technovation:
> The only feedback I have is to consider adding helper text when there isn't any tracked activity.

Great suggestion! 💯 I just assumed everyone would have activity since it's already being tracked, anywho, good catch, good suggestion! I'll make this change to accommodate for this! 👍

PR #4413, opened 2024-01-11
Update team invite registration validation

shaun-technovation:
This will make it so that team invites to register will be valid when "Forming teams allowed" is enabled in the admin (instead of using the submission close date). And I added some very basic, rudimentary help text to the admin; and updated the JS specs accordingly.

### DIFF
--- a/app/services/team_invite_code_validator.rb
+++ b/app/services/team_invite_code_validator.rb
@@ -1,14 +1,13 @@
 class TeamInviteCodeValidator
-  def initialize(team_invite_code:, important_dates: ImportantDates)
+  def initialize(team_invite_code:)
     @team_invite_code = team_invite_code
-    @important_dates = important_dates
   end
 
   def call
     invite = TeamMemberInvite.find_by(invite_token: team_invite_code)
 
     if invite.present? && invite.pending? && invite.inviter_type == "StudentProfile" &&
-        (important_dates.official_start_of_season..important_dates.submission_deadline).cover?(Date.today)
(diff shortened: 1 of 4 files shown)


app/services/team_invite_code_validator.rb
```diff
@@ -1,14 +1,13 @@
...
     @team_invite_code = team_invite_code
-    @important_dates = important_dates
   end
 
   def call
     invite = TeamMemberInvite.find_by(invite_token: team_invite_code)
 
     if invite.present? && invite.pending? && invite.inviter_type == "StudentProfile" &&
-        (important_dates.official_start_of_season..important_dates.submission_deadline).cover?(Date.today)
+        SeasonToggles.team_building_enabled?
```
shaun-technovation, line 10: This is the main change - instead of using the submission deadline date to check if a team invite is valid, we'll now use the "team building" admin setting.

viviancan (approved):
Everything worked as expected! I was able to register as a mentor and as a student from a team invite even if registration was off (team forming had to be on!)

viviancan:
Noting when you register from a team invite, the email field in the registration form is "open" and not prefilled with the invited email address. On one of my trials, I entered a different email address and was able to register from the invite link, but did not see the team invite in my dashboard. I then updated my email to match the email from the invite and I still did not see the invite in my dashboard. As an admin, I could see the pending invite with the "not registered" text. I'm wondering if we need to have the email address prefilled and uneditable in the registration form when there are invites. Tagging @dboyer

PR #4416, opened 2024-01-12
Update Puma gem

shaun-technovation:
This will update the Puma gem from version `5.6.7` to `5.6.8` to address this security vulnerabilty:

- https://github.com/Iridescent-CM/technovation-app/security/dependabot/175


viviancan (approved):
Did not try locally, but seems like a small bump! 💯

PR #4417, opened 2024-01-12
Update postcss-svgo node module

shaun-technovation:
This will update `postcss-svgo` to address these security vulnerabilities:

- https://github.com/Iridescent-CM/technovation-app/security/dependabot/64
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/4

By updating `postcss-svgo`, it removes the `is-svg` dependency which addresses the security vulnerabilities.


viviancan (approved):
💯

Issue #4409, opened 2024-01-09
Admin ability to add ChA profile to a mentor account


shaun-technovation:
This ticket covers the basic core functionality of allowing PT to go to a mentors admin page and click a button to add a Chapter Ambassador profile to that mentor account. There will be additional tickets including:

- We need to send an email to the new ChA with information about their new account
- We need the ability to maintain a mentor mode for the new ChA
- We need to have an option to convert a ChA account back to just a mentor account

Notes: The most common use case is a mentor who becomes a Chapter Ambassador. For context, we will likely be adding a club ambassador role in the future and admins will also need to be able to convert that role to a ChA.

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/4409?planning-poker) with Zenhub @shaun-technovation @viviancan

viviancan:
Confirming this is on QA and working as expected. I was able to add a ChA profile to an account with a mentor profile. I the mentor profile did not have a filled in/completed bio I received an error with the bio must be filled out error.

dboyer:
Confirmed on QA:

- I see the button a mentor account to add a ChA profile
- Logged in as the new ChA and went to their mentor profile where I had a button to switch to ChA mode
<img width="356" alt="Screen Shot 2024-01-30 at 10 53 33 AM" src="https://github.com/Iridescent-CM/technovation-app/assets/516764/eb6f519a-3aac-43d4-bb75-b020a4f8872d">
- Was marked as pending and then when I approved the ChA account as an admin, I was able to see the full ChA account when I logged back in

Tested also on the preview site:

- Button to convert to a ChA was visible on the mentor's admin participant page
- Logged in as the new ChA and went straight to new ChA dashboard, with no interim pending step
- Was able to use the mentor mode button to switch back to the mentor dashboard

I see the button on a judge account if they also have a mentor profile but not if they have just a judge account. I think that's the right behavior. We'll need an option to remove the ChA profile for a user and return them to being just a mentor but that can be a different ticket.

viviancan:
Confirming i saw the button on a mentor profile in the admin portal

PR #4423, opened 2024-01-18
Added ability for admin to add a ChA profile to an account that already has a mentor profile

viviancan:
Refs #4409 

This PR allows admin to add a ChA profile to an account with a mentor profile. 

CleanShot 2024-01-18 at 16 55 40@2x

Noting for historical context, we discussed the specifics of "profile" vs. "account" wording/terminology. All of the other “conversions” use a similar naming convention, for example, student conversions or mentor to judge conversions.

The term "conversion" doesn't necessarily fit this scenario, because we are adding an additional profile type (the ChA profile) while the account still has a a mentor profile. For additional context, we may need to have the option to convert club ambassadors to ChA in the future. 

There still may be additional split-off items as listed in #4409


app/controllers/concerns/chapter_ambassador_profile_addition_controller.rb
```diff
@@ -0,0 +1,11 @@
+module ChapterAmbassadorProfileAdditionController
```
shaun-technovation, line 1: Hmm, this should only be used by admins, so I don't think it needs to be a `concern`. 🤔 Def not a blocker, but it adds an extra step/layer that I don't think is needed. 🤔

app/controllers/concerns/chapter_ambassador_profile_addition_controller.rb
```diff
@@ -0,0 +1,11 @@
...
+  extend ActiveSupport::Concern
+
+  def create
+    account = Account.find(params.fetch(:account_id))
+    chapter_ambassador_profile_addition_result = ChapterAmbassadorProfileAddition.new(account: account).call
+
+    redirect_to send("#{current_scope}_participant_path", id: account.to_param),
```
shaun-technovation, line 8: I think the `current_scope` can be hard-coded to `admin` since it looks like this is only used by admins. 🤔

spec/system/admin/add_chapter_ambassador_profile_to_account_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+      click_link "Add Chapter Ambassador profile"
+
+      account.reload
+
+      expect(account.chapter_ambassador_profile).to be_present
+    end
+  end
+
+  context "when a chapter ambassador profile cannot be added to an account" do
+    it "does not show an add chapter ambassador profile button" do
```
shaun-technovation, line 22: 👍

shaun-technovation (approved):
LGTM! Nice work on this one! Pretty straightforward! Nice addition! 👍 🎉 🙌 :shipit:

app/controllers/concerns/chapter_ambassador_profile_addition_controller.rb
viviancan: Removed and updated!

PR #4425, opened 2024-01-19
Added rake task to update bg check internal invitation status

viviancan:
Refs #4247

This rake task is needed to update the new `internal_invitation_status` field for background checks. It will get all background checks that have a value for `invitation_status` (meaning that an invitation has been sent) and then update those records to `invitation_sent`. 

Running locally: 
`bundle exec rake 'update_bg_check_internal_invitation_status[run]'`

There is also a dry mode
`bundle exec rake 'update_bg_check_internal_invitation_status[dry]'`

On QA:
dry first, then run
`heroku run bundle exec rake 'update_bg_check_internal_invitation_status[dry]' --app technovation-qa`

On PROD:
dry first, then run
`heroku run bundle exec rake 'update_bg_check_internal_invitation_status[dry]' --app technovation`

As of right now, Jan 19, 2024 @ 10:04CST there are 202 records on production that will be updated.

### DIFF
--- a/lib/tasks/update_bg_check_internal_invitation_status.rake
+++ b/lib/tasks/update_bg_check_internal_invitation_status.rake
@@ -1,0 +1,17 @@
+desc "Update background check internal invitation status"
+task :update_bg_check_internal_invitation_status, [:dry_run] => :environment do |t, args|
+  dry_run = args[:dry_run] != "run"
+
+  background_checks = BackgroundCheck.where.not(invitation_status: nil)
+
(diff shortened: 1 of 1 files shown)


shaun-technovation (approved):
LGTM! Nice usage of `find_each` and having dry-run functionality. I'm also liking how `internal_invitation_status` reads after looking at it w/ fresher eyes. 👍 all around! 🙌 🥳 🎉 🚢 :shipit:

PR #4429, opened 2024-01-19
Security: Update dns-packet node module

shaun-technovation:
This will update the dns-packet node module to address this security vulnerability:

- https://github.com/Iridescent-CM/technovation-app/security/dependabot/10


viviancan (approved):
👍

PR #4441, opened 2024-01-25
Allow chapter ambassadors to edit their own location

shaun-technovation:
This will make it so that a chapter ambassador can edit their own location.


viviancan (approved):
LGTM! The only thing I noticed was the "header" container title was missing the blue background.

CleanShot 2024-01-25 at 15 05 10@2x

shaun-technovation:
> LGTM! The only thing I noticed was the "header" container title was missing the blue background.

Good eye, great catch on this, thank you! I fixed this. 👍

PR #4448, opened 2024-01-25
Chapter model groundwork

shaun-technovation:
This will add a migration and model for chapters, and also create the association between chapter ambassadors and chapters.


app/models/chapter_ambassador_profile.rb
```diff
@@ -10,6 +10,7 @@ class ChapterAmbassadorProfile < ActiveRecord::Base
   }
 
   belongs_to :account
+  belongs_to :chapter, optional: true
```
shaun-technovation, line 13: I had to add the `optional: true` part because all the current chapter ambassadors aren't currently associated with chapters; but once they all are I think we can remove this `optional` bit of code.

app/models/chapter_ambassador_profile.rb
```diff
@@ -48,8 +49,8 @@ class ChapterAmbassadorProfile < ActiveRecord::Base
     prefix: true,
     allow_nil: true
 
-  def method_missing(method_name, *args)
-    account.public_send(method_name, *args)
+  def method_missing(method_name, *args) # standard:disable all
+    account.public_send(method_name, *args) # standard:disable all
```
shaun-technovation, line 53: StandardRB tries to format this and then it breaks, so for now I just added this to bit so StandardRB will just ignore these two lines of code.

db/migrate/20240125153743_add_chapters.rb
```diff
@@ -0,0 +1,12 @@
+class AddChapters < ActiveRecord::Migration[6.1]
+  def change
+    create_table :chapters do |t|
+      t.string :non_profit_name
```
viviancan, line 4: I think this name may be changing based on the wireframe feedback meeting! We can double check w/ Deb on Monday!

viviancan (approved):
👍 Looks good to me! Just left 1 comment based on the chapter wireframe meeting

db/migrate/20240125153743_add_chapters.rb
dboyer: Yes! Program Team suggested changing any mention of non-profit to "organization" as that word is more commonly understand internationally than non-profit.

PR #4450, opened 2024-01-26
Update browserslist node module

shaun-technovation:
This will update browserslist and its dependecies to address this security vulnerability:

- https://github.com/Iridescent-CM/technovation-app/security/dependabot/9


shaun-technovation:
Closing this one for now, since it seems to be having some unintended side effects.

PR #4458, opened 2024-01-29
WIP: Add ability to publish submissions only needing to submit

shaun-technovation:
This will let super admins publish all submissions that only need to submit.


app/views/admin/_navigation.html.erb
```diff
@@ -25,6 +25,13 @@
       data: { turbolinks: false },
       class: al(admin_team_submissions_path) %>
 
+    <% if current_profile.try(:super_admin?) %>
```
viviancan, line 28: Nice try method! I haven't come across this in the platform before 💯

viviancan (approved):
💯 Woohoo, tried locally and I think it is working as expected. The publish button published 4 submissions. I verified that they were published with a `published_at` timestamp! Awesome work - this is going to help us so much 👍

app/views/admin/_navigation.html.erb
shaun-technovation: This was an annoying last minute thing I had to fix. The "super admin inviting an admin" spec was failing; I think b/c the invited admin is registering for their account (and they're not logged in), but the navigation is trying to get displayed, and `super_admin?` isn't available for this situation (I was also trying to use `current_admin` which also isn't available), so I had add the `try` to address this. 😅

Issue #4393, opened 2023-12-20
Integrate Checkr background checks for all chapter ambassadors


dboyer:
The first step of the ChA onboarding process should be integrating a Checkr background check. This should follow a similar flow as the check for international mentors in that ChAs view some initial information on the platform and request an invite to start the process but otherwise do everything via Checkr. A few notes:

1. The background check should be the first step listed in the onboarding process. However, ChAs can access the training onboarding step even if they do not have their check completed. They cannot access the MOU or enter chapter profile information.

2. Ideally, all ChAs including US based ones will complete the process via Checkr. We do not want US based ChAs to go through the check process on the platform as it is currently available to US mentors. Instead they should go through the off-platform process.

3. We will likely want the check to be good for several years so we'll need to figure out how that works.
  
<img width="536" alt="Screen Shot 2023-12-19 at 9 46 57 PM" src="https://github.com/Iridescent-CM/technovation-app/assets/516764/ed443ffe-6588-4dc6-80c9-989ff7c4ed10">

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/4393?planning-poker) with Zenhub @shaun-technovation @viviancan

viviancan:
Noting that there is some weird styling in the side nav due to the "tw-active" class. This was captured in #4464

viviancan:
This is on preview-qa. Confirming I was able to request a bg check invitation as a US based and nonUS based chapter ambassador. I also tried requesting an invitation when the last name was 1 character and I could not request the invitation. I saw the error message directing me to update the last name.

viviancan:
@dboyer This is due to the new sidenav work that was recently merged into preview. I am going to update the background check link with the new sidenav and you should be able to access the background check as normal.

viviancan:
Actually, I just saw that Shaun is already working on that in #4495!

shaun-technovation:
>> I created two new ChA accounts on preview via invites. I was able to create the accounts and set locations in the US. When I click on the background check tab, nothing shows up. The accounts were /admin/participants/1608#/ and /admin/participants/1607#/

> @dboyer This is due to the new sidenav work that was recently merged into preview. I am going to update the background check link with the new sidenav and you should be able to access the background check as normal.

[...]

PR #4462, opened 2024-01-30
Pt2 - Updated UI for ChA background check flow

viviancan:
Refs #4393 

There isn't anything major in this PR - mostly all of the UI/view updates for the rebranded version of the background check flow.

https://github.com/Iridescent-CM/technovation-app/assets/29210380/0673724e-38df-4d2b-991a-f7fc2296c7f1

### DIFF
--- a/app/services/checkr_api_client/api_client.rb
+++ b/app/services/checkr_api_client/api_client.rb
@@ -94,7 +94,7 @@ module CheckrApiClient
 
     def create_checkr_invitation(candidate_country_code:, candidate_id:)
       req_body = {
-        package: "international_basic_plus",
+        package: determine_package(candidate_country_code),
         candidate_id: candidate_id,
         work_locations: [{country: candidate_country_code}]
       }
@@ -117,6 +117,10 @@ module CheckrApiClient
       log_error_msg = "[CHECKR] Error for account #{candidate.id} - #{error}"
       logger.error(log_error_msg)
       error_notifier.notify(log_error_msg)
+    end
+
+    def determine_package(candidate_country_code)
+      candidate_country_code == "US" ? "tasker_standard" : "international_basic_plus"
     end
   end
 end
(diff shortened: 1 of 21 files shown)


app/views/background_checks/rebrand/_show.html.erb
```diff
@@ -0,0 +1,130 @@
+<div class="mx-auto w-full lg:w-2/3">
+  <%= render layout: 'application/templates/dashboards/energetic_container', locals: { heading: "Background Check Status"} do %>
```
shaun-technovation, line 2: Looks like mismatched quotes and missing (or extra space) around the curly braces.

shaun-technovation (approved):
LGTM! Nice video! One thing I noticed in it is that the sidenav isn't there. I would think that it should be there for all onboarding steps; it would be weird to have some onboarding steps with the sidenav and some without, imho. Anywho, it can be addressed later on too. Overall, nice work, really exciting to see more of the ChA onboarding coming together! 🙌 💯 🥳

app/services/checkr_api_client/api_client.rb
```diff
@@ -94,7 +94,7 @@ def create_checkr_candidate
 
     def create_checkr_invitation(candidate_country_code:, candidate_id:)
       req_body = {
-        package: "international_basic_plus",
+        package: determine_package(candidate_country_code),
```
viviancan, line 97: @shaun-technovation I realized the US bg check invitations were erroring because US-based participants cannot use the international package. Added `determine_package` to handle this.

app/services/checkr_api_client/background_check_invitation_validator.rb
```diff
@@ -25,7 +25,7 @@ def call
 
     Result = Struct.new(:requires_background_check_invitation?, :error?, :error_message, keyword_init: true)
 
-    attr_reader :account, :current_profile_type
```
viviancan, line 28: Removed after update

spec/features/background_check_spec.rb
```diff
@@ -26,22 +26,6 @@
     expect(page).not_to have_link("Submit Background Check")
   end
 
-  scenario "Complete a chapter ambassador background check", :vcr do
```
viviancan, line 29: Removed since this process is no longer applicable for ChA

app/services/checkr_api_client/api_client.rb
shaun-technovation: Good catch!

PR #4463, opened 2024-01-30
PT1 - Added background check invitation flow for all Chapter Ambassadors

viviancan:
Refs #4393 

This PR makes background checks required for all ChAs regardless of location. I also added a new background check invitation validation service that determines if the profile (Mentor or ChA) requires a background check invitation. Eventually, all profile types will move toward the invitation flow. The service also checks last name length to avoid the "2 alpha character" requirement from Checkr (thanks for the recommendation, Shuan 💯 )! 

Pt 2 will include all of the view/UI updates!


app/controllers/chapter_ambassador/background_checks_controller.rb
```diff
@@ -1,6 +1,9 @@
 module ChapterAmbassador
   class BackgroundChecksController < ChapterAmbassadorController
     include BackgroundCheckController
+    include BackgroundCheckInvitationController
+
+    layout "chapter_ambassador_rebrand"
```
shaun-technovation, line 6: 👍

app/helpers/application_helper.rb
```diff
@@ -18,7 +18,7 @@ def determine_manifest
     namespace = controller_path.split("/").first
     path = controller_path.split("/").second
 
-    if namespace == "chapter_ambassador" && (path == "dashboards" || path == "profiles")
+    if namespace == "chapter_ambassador" && (path == "dashboards" || path == "profiles" || path == "background_checks")
```
shaun-technovation, line 21: 👍 🙈 😂 Hopefully we can remove this hack soon! 😅

app/controllers/mentor/dashboards_controller.rb
```diff
@@ -3,7 +3,6 @@ class DashboardsController < MentorController
     include LocationStorageController
 
     def show
-      @job = current_mentor.jobs.queued.first
```
shaun-technovation, line 6: I'm guessing this is just leftover from the mentor work that was recently done? 🤔

app/models/chapter_ambassador_profile.rb
```diff
@@ -71,12 +72,11 @@ def needs_intro_prompt?
...
 
   def background_check_complete?
-    country_code != "US" or !!background_check && background_check.clear?
+    !!background_check && background_check.clear?
   end
 
   def requires_background_check?
-    country_code == "US" and
-      !background_check_complete?
+    !background_check_complete?
```
shaun-technovation, line 79: Ah nice! Pretty simple to make BG checks available to all countries! 👍

app/services/checkr_api_client/background_check_invitation_validator.rb
```diff
@@ -0,0 +1,34 @@
...
+          error_message: "Checkr requires last name to contain at least 2 alpha characters. Please edit your last name."
+        )
+      end
+
+      case current_profile_type
+      when "ChapterAmbassadorProfile"
+        requires_background_check_invitation = account.chapter_ambassador_profile.requires_background_check?
+      when "MentorProfile"
+        requires_background_check_invitation = account.mentor_profile.requires_background_check_invitation?
+      end
```
shaun-technovation, line 21: Ah, I see how `current_profile_type` is being used here; another way would be to not even pass in the profile type, and just do `if account.is_chapter_ambassador?` and `account.is_mentor?`, but also, not a big deal for this one either. 😄

app/models/mentor_profile.rb
```diff
@@ -300,6 +300,15 @@ def requires_background_check?
       !(background_check.present? && background_check.clear?)
   end
 
+  def background_check_invitation_expired_or_error?
```
shaun-technovation, line 303: Looks like this method can be `private`, since it's not being used outside of this class.

app/services/checkr_api_client/background_check_invitation_validator.rb
```diff
@@ -0,0 +1,34 @@
...
+
+      Result.new(requires_background_check_invitation?: requires_background_check_invitation)
+    end
+
+
```
shaun-technovation, line 26: Looks like an extra line break here.

config/routes.rb
```diff
@@ -173,6 +173,7 @@
     resource :profile_image_upload_confirmation, only: :show
 
     resources :background_checks, only: [:new, :create, :show]
+    post "/background_check_invitation", to: "background_checks#create_invitation"
```
shaun-technovation, line 176: Most of our routes are `resources` I wonder if this can be something like `resources :background_check_invitations, only: :create` 🤔, not a big deal either.  Also, I'm at the end of the PR and don't see any specs. 😅 🙈 😂

shaun-technovation (approved):
Left some feedback, but no blockers. Nice work on adding this to the new ChA flow! And I see it will get merged into the preview branch, right on! 👍 🙌 🎉 🚢

PR #4478, opened 2024-02-02
Security: Update axios node module

shaun-technovation:
This will update axios to address this security vulnerability:

- https://github.com/Iridescent-CM/technovation-app/security/dependabot/173

This will also update axios's dependencies which will also address this security vulnerability:

- https://github.com/Iridescent-CM/technovation-app/security/dependabot/176

I know one of the ways we use axios is on the new registration flow, getting the mentor skills and abilities and mentor types, I tested that out locally and it worked fine for me.


viviancan (approved):
I registered as a mentor and was able to login 💯

PR #4491, opened 2024-02-07
WIP: Super admins can only publish submissions when "editing submissions" is tured off

shaun-technovation:
This update will make it so that super admins can only publish submissions when "editing submissions" is turned off.


viviancan (approved):
LGTM 💯 👍

PR #4494, opened 2024-02-08
Make chapter ambassador sidenav dynamic

shaun-technovation:
This will add a new "todo item" partial (happy to change the name of this), this will display:
- a checkmark (for completed items)
- an empty circle (for not completed)
- a small circle in the circle (if it's the current/active section)

These new todo items will get displayed in the chapter ambassador sidenav.

I also remove the custom CSS for the `tw-active` fix that isn't needed now.


app/views/chapter_ambassador/dashboards/rebrand/_side_nav_content.html.erb
```diff
@@ -1,58 +1,31 @@
...
-      </div>
-    </li>
+  <nav class="p-4" aria-label="Progress">
+    <ol role="list" class="space-y-6 mb-6 overflow-hidden">
+      <%= render "application/templates/todo_item",
+        name: "Background Check",
+        url: "#",
+        is_complete: current_ambassador.background_check_complete?,
+        is_active_item: false
+      %>
```
shaun-technovation, line 8: This is how the new todo items are used, they need to provide a name, URL, a condition on whether it's complete, and a condition for if it's the active item. Happy to rename any of this, I was struggling with the naming.

viviancan: I think the name is fine! I think in the mentor/student experience w/ the Vue apps they are called "completion steps."

viviancan (approved):
Nice improvements here 💯 👍

app/views/chapter_ambassador/dashboards/rebrand/_side_nav_content.html.erb
shaun-technovation: Ah, okay, I think I'll update this to `completion_step` then, so it's more consistent w/ what we have. 👍

PR #4496, opened 2024-02-08
Add new judge recusal reason

shaun-technovation:
This will add a new judge recusal option - "Content doesn't belong to the team".

### DIFF
--- a/app/javascript/config/axios.js
+++ b/app/javascript/config/axios.js
@@ -1,30 +1,34 @@
-import axios from 'axios'
-import applyConverters from 'axios-case-converter'
+import axios from "axios";
+import applyConverters from "axios-case-converter";
 
 // Polyfill Promise for IE11 to support axios
-require('es6-promise').polyfill()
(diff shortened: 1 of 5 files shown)


app/javascript/config/axios.js
```diff
@@ -1,30 +1,34 @@
...
-        // of axios-case-converter (see their documentation)
-        config.headers.common = {
-          'X-Requested-With': 'XMLHttpRequest',
-          'X-Csrf-Token' : csrfTokenMetaTag.getAttribute('content')
+    window.axios.interceptors.request.use(
+      function (config) {
+        if (config.url && config.url.match(/^\//)) {
+          // The following headers must be in "Header-Case" because
+          // of axios-case-converter (see their documentation)
+          config.headers = {
```
shaun-technovation, line 20: All of the changes in this file were to update the formatting except this line^ which is updating `config.headers.common` to `config.headers`, which will properly include/send the CSRF token.

PR #4499, opened 2024-02-12
Fix axios CSRF issue

shaun-technovation:
The changes here will properly include the CSRF token in the headers for axios calls. This will fix the location issue, judge recusal popup submission issue, and most likely a few others too.


shaun-technovation:
Nevermind, I'm closing this one, we can just use the fix in #4496 since it's being merged to QA (and will also get merged to preview).

PR #4503, opened 2024-02-14
Display pending chapter ambassadors in participants list in admin

shaun-technovation:
This will display pending chapter ambassadors in the participants list in the admin.


viviancan (approved):
Did not try locally, but seems pretty straightforward! Nice fix 💯

PR #4510, opened 2024-02-16
Update Follow Redirects node module

shaun-technovation:
This is a very minor version bump from `1.15.3` to `1.15.5` for the Follow Redirects node module that will address this security vulnerability:

- https://github.com/Iridescent-CM/technovation-app/security/dependabot/176


viviancan (approved):
Did not try locally, but looks like a small bump 👍

PR #4534, opened 2024-02-27
Inviting chapter ambassadors to chapters

shaun-technovation:
This will:
- Allow admins to invite chapter ambassadors to a chapter
- After the chapter ambassador registers, it will associate them to that chapter


viviancan (approved):
Wow! This is great 🤩 I was able to invite a ChA and they were associated to the correct chapter. Nice tests too 🚀  Really excited to see this come together!

Issue #4520, opened 2024-02-22
Issues with adding adding new judges to an event


dboyer:
ChAs have the ability to send an invite to a judge to register on the platform and be associated with an event. I was testing that feature and while it works, there are lags and delays that a regular user would likely not be able to get past. My testing process was:

- Log in as a ChA and go to the events tab
- Create a new event or go to an existing event and click the icon to add judges
- Click the "Add judges" button and see the pop-up modal where you can pick a judge from the list or enter an email address
- Enter an email address to send an invite for someone to register on the platform
- Click done
- The invited email address does not appear on the list of judges
- Click "add judges" again to open the modal. The invited judge's email is now in the drop-down list
- Select that email and click done
- The email appears on the list of judges with a checked box to send an invite
- Click to send the invite
- Open the modal again to add a new judge and see the old judge's email address there.
- There's no way to add a new judge

This screen capture shows the whole process - https://drive.google.com/file/d/1O5PAZgCcN7T8_51mgjzS9Z7dfWPC8vj2/view?usp=drivesdk

viviancan:
Confirming this is on QA and working as expected

dboyer:
Confirmed that I could add a judge to an event on QA. When I searched, the list populated possible judge choices based on my search terms. I was also able to invite a new judge without the delayed screen issue reported earlier.

PR #4535, opened 2024-02-27
Updated created async method for adding judges to rpe bug

viviancan:
Refs #4520 

This PR addresses a bug in the ChA RPE management page. There are a lot of code style updates but the main update for this PR was removing a 2nd async create method and adding the logic to the 1st/original created method. 

As a ChA, the auto filter/search does not work when trying to search or add a judge. Similarly, when trying to add a judge by inviting them, there is a lag and a very weird flow. The error in the console is
`TypeError: this.debouncedFetchRemoteItems is not a function`. This led me to review the initial created method, which includes the following code
``` 
this.debouncedFetchRemoteItems = debounce(() => {
      this.fetchRemoteItems({ expandSearch: 1 });
    }, 300);
 ```
I also discovered a 2nd async created method. I believe that because of this 2nd created method, the 1st created method was not being called properly, leading to the `is not a function` error message. I added the logic from the 2nd async method to the original created method, and everything is working as expected, including the "search" option.

### DIFF
--- a/app/javascript/chapter_ambassador/events/AttendeeSearch.vue
+++ b/app/javascript/chapter_ambassador/events/AttendeeSearch.vue
@@ -15,14 +15,12 @@
         New teams cannot be added to events at this time.
       </p>
 
-      <div
-        class="modal-container"
-        v-show="searching"
-      >
+      <div class="modal-container" v-show="searching">
         <div class="modal">
           <div v-if="eventAtCapacity()" class="margin--b-xlarge">
             <div class="flash flash--error margin--none">
-              This event is currently at capacity. No additional teams can be added.
+              This event is currently at capacity. No additional teams can be
+              added.
             </div>
           </div>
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: app/javascript/chapter_ambassador/events/AttendeeSearch.vue [hunk 1]
```diff
@@ -94,185 +83,202 @@
...
         this.fetchRemoteItems();
-      });
+      }
+    );
+
+    this.debouncedFetchRemoteItems = debounce(() => {
+      this.fetchRemoteItems({ expandSearch: 1 });
+    }, 300);
+
+    await this.getRegionalPitchEventSettings();
```
viviancan, line 148: Added this

app/javascript/chapter_ambassador/events/AttendeeSearch.vue
```diff
@@ -94,185 +83,202 @@
...
+        Boolean(this.event.capacity) &&
+        this.event.selectedTeams.length >= this.event.capacity
+      );
+    },
+
+    onQueryChange() {
+      this.debouncedFetchRemoteItems();
     },
 
-    async created() {
```
viviancan, line 271: removed this

Issue #4470, opened 2024-01-31
Add "Background Checks" page to admin portal


viviancan:

dboyer:
Confirming on preview that I can:

- See the new background checks admin page
- Sort by the available filters
- Add columns for profile and country
- Export data

@viviancan One needed change. When I exported the data, the account name column was not included. Could we add that?

viviancan:
[This PR](https://github.com/Iridescent-CM/technovation-app/pull/4536) adds the account name column to the csv export

viviancan:
Confirming the updated csv export is on preview

dboyer:
I also see the account name column in the export now.

PR #4536, opened 2024-02-27
Added account to csv export for background checks grid

viviancan:
Refs #4470 

Added account to csv export for background checks grid


shaun-technovation (approved):
Looks good/straight forward to me! 👍

Issue #4457, opened 2024-01-29
Placeholder - Add background check status to ChA profile in admin portal


viviancan:

dboyer:
I see the background check status on the ChA participant page on QA.

viviancan:
Confirming I see the bg check stats for ChA on prod

viviancan:
@shaun-technovation Moving this back into code review since I discovered a bug related to the work. [This PR ](https://github.com/Iridescent-CM/technovation-app/pull/4540) addresses the issue

viviancan:
This is on prod

PR #4540, opened 2024-02-29
Fixed background check invitation status bug on admin profile

viviancan:
Discovered a bug in admin portal related to the logic for the bg check invitation status details in the profile debugging section. The template errors out for profiles in background check countries for profiles that have not yet requested an invite. The bug is currently on production, so I'm hoping for a quick additional deployment tomorrow. This change updates the logic and adds a catch-all for edge cases. 

You can recreate this on QA by signing up as a new mentor in a background check country. Then, view the mentor details in the admin portal, and the platform will display the default error page. 

spinoff from #4457


shaun-technovation (approved):
LGTM! Good catch! 👍

PR #4547, opened 2024-03-01
Display chapter ambassadors on the admin chapter show page

shaun-technovation:
I created a new section on the chapter show page that will display a list of chapter ambassadors that are associated with the chapter.

I also slightly tweaked the chapter page so it's more consistent withthe admin participants show page.


viviancan (approved):
Did not try locally, but looks good to me 👍

Issue #4483, opened 2024-02-07
Preview: Add "sync now" button for super admin in the background check admin page


viviancan:

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/4483?planning-poker) with Zenhub @shaun-technovation @viviancan

viviancan:
Also, the background check sync will only sync background checks from this season

viviancan:
Confirming this is on PREVIEW

dboyer:
On preview, I clicked the "sync now" button and then received an email for one of my accounts saying that my check was cleared. I believe this is working as expected.

PR #4552, opened 2024-03-04
Add sync now functionality to admin background check view

viviancan:
Refs #4483 

This PR adds (again) the ability for admin to sync background checks (background check status & invitation status) on our platform. This will check all current season background checks that are not a "clear" status.

I added #4551 to remove the old background check sweep functionality.


app/jobs/sync_background_checks_job.rb
```diff
@@ -0,0 +1,25 @@
+class SyncBackgroundChecksJob < ActiveJob::Base
+  queue_as :default
+
+  before_enqueue do |job|
```
shaun-technovation, line 4: This `before_enqueue` (and `after_perform`) code is definitely making me feel pretty daft! 😩 I can't seem to make out why we need it this? 🤔

app/jobs/sync_background_checks_job.rb
```diff
@@ -0,0 +1,25 @@
...
+    admin_profile_id = job.arguments[0][:admin_profile_id]
+    profile_type = "AdminProfile"
+
+    profile = profile_type.constantize.find(admin_profile_id)
```
shaun-technovation, line 8: This part looks like it can be in-lined, `AdminProfile.find` vs `profile_type.constantize.find`. 🤔

app/jobs/sync_background_checks_job.rb
```diff
@@ -0,0 +1,25 @@
...
+    db_job = Job.find_by(job_id: job.job_id)
+    db_job.update_column(:status, "complete")
+  end
+
+  def perform(admin_profile_id)
+    accounts = Account.current.joins(:background_check).where.not(background_checks: {status: :clear})
+
+    accounts.find_each do |account|
+      CheckrApiClient::SyncBackgroundCheck.new(account: account).call
+    end
```
shaun-technovation, line 23: 👍

app/services/checkr_api_client/sync_background_check.rb
```diff
@@ -0,0 +1,29 @@
+module CheckrApiClient
+  class SyncBackgroundCheck
```
shaun-technovation, line 2: Hmm, this file is living under the Checkr API client folder structure, but there isn't anything in this file that uses the API directly.

app/services/checkr_api_client/sync_background_check.rb
```diff
@@ -0,0 +1,29 @@
...
+      @account = account
+    end
+
+    def call
+      background_check = @account.background_check
+
+      if !background_check.invitation_completed?
+        sync_invitation
+      else
+        sync_background_check
```
shaun-technovation, line 13: Ah, interesting, we're checking if the invitation is not complete to `sync_invitation`, but we don't check if a background check is not complete to `sync_background_check`, is that because in the job, we just grab the background checks that haven't been cleared? If so, I wonder if we could include just incomplete invitations in that query? Or does it already cover those? Just wondering b/c the line `!background_check.invitation_completed?` suggests completed invitations could come through this flow. 🤔

app/jobs/sync_background_checks_job.rb
viviancan: Ahh ok, I thought that we typically kept a record of admin background jobs (like the exports!) But I can remove it 😄

app/services/checkr_api_client/sync_background_check.rb
viviancan: 👍 I will move it out of this directory

app/jobs/sync_background_checks_job.rb
shaun-technovation: Ooooh!! 🤦 on my part! I see!!! It tracks the admin who created the job! 🤦 😊 Oooh, so this would popup the little window that indicates the job is running (like the exports)? That seems like a nice feature! 👍

app/services/checkr_api_client/sync_background_check.rb
viviancan: This query
`Account.current.joins(:background_check).where.not(background_checks: {status: :clear})`
includes all accounts with background checks that are not clear. Some of the bg checks have invitations and some do not. So the if/else logic checks if there is an incomplete invitation and then will sync the invitation status. If there is not an incomplete invitation then the background check status is checked/updated. 

> Just wondering b/c the line !background_check.invitation_completed? suggests completed invitations could come through this flow. 🤔

Completed invitations can come through this flow. It is possible that an invitation is complete but the background check is still pending.

PR #4558, opened 2024-03-05
Chapter invites on chapter page

shaun-technovation:
This will a add a form to the chapter show page that will allow an admin to invite a chapter ambassador to that chapter. I also added a basic list of pending invites to this page as well.


viviancan (approved):
Everything worked as expected 💯  Nice work on this 👍

Issue #4538, opened 2024-02-28
Remove DOB from checkr invitation flow


viviancan:

viviancan:
Refs #4537

We will most likely need to remove this from the `create_checkr_candidate` method in the checkr API client 
`dob: candidate.date_of_birth`

dboyer:
Tested on preview and was able to request a background check invite. When I clicked the link, the DOB was not filled in so confirming we are no longer sending that field. I was able to complete the form. The check cleared as expected.

Issue #4537, opened 2024-02-28
Ask ChA to confirm they are over 18


dboyer:
Currently, we ask chapter ambassadors to provide their birth date. We would like to switch to a check box that they manually check to confirm they are over the age we set to be a chapter ambassador. The box should have the words "I confirm that I am 18 years or older."

Note: we need to see if this impacts Checkr as we may be sending the birth date to them to start the background check.

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/4537?planning-poker) with Zenhub @shaun-technovation @viviancan

shaun-technovation:
Re: the work here should accommodate for judges and mentors to use this functionality... Maybe there's not much to that part... I'm wondering if simply adding a new `Account` level field would work, and naming it something generic like `over_required_age`, this could be used by all profiles and would safe guard against different profiles having different age requirements or having the age requirement change in the future. (The frontend would still show the actual age requirement, and the git history would have historical context for what it was set at for a given season if we ever needed it.) And then adding the field/validation which could be reused (or copied) for each profile. 🤔 

Thoughts @viviancan 🤔

dboyer:
It would be great to unify the age question as much as possible. We are good to switch to "confirm you are over 18" for judges next season but have not made that decision for mentors yet.

viviancan:
Agreed^^ @shaun-technovation I think it sounds like a great approach especially to account for potential changes with the other profile types too!

shaun-technovation:
This one is on Preview.

One thing to look out for is that since chapter ambassadors don't enter their DOB anymore, there are places in the app that are expecting a DOB, but there isn't one. So, if we find something broken w/ chapter ambassadors this might be the cause.

The "I confirm that I am 18 years or older" response isn't displayed anywhere, I guess it's assumed all chapter ambassadors on the platform meeting the minimum age requirement; we could display this field as well. Also, I added this new field second on the registration flow, right after first/last name, my thinking was that if we accidentally invite a chapter ambassador who doesn't meet the age requirement, they wouldn't fill out a bunch of fields and then realize they couldn't register (if the field was lower/later in registration). Anywho, we can change the order of this field too.

We could migrate existing chapter ambassadors to the new "I confirm that I am 18 years or older" format which would allow us to remove their DOB (which sets us up better to only collect the data we need).

[...]

PR #4559, opened 2024-03-05
Remove dob from checkr request

viviancan:
Refs #4538 

Removes dob from the checkr request when creating a candidate. This was needed due to the change in DOB collection for ChA in #4537


app/services/checkr_api_client/api_client.rb
```diff
@@ -85,7 +85,6 @@ def create_checkr_candidate
         email: candidate.email,
         first_name: candidate.first_name,
         last_name: candidate.last_name,
-        dob: candidate.date_of_birth,
```
shaun-technovation, line 88: 👍

shaun-technovation (approved):
👍

PR #4577, opened 2024-03-14
Update chapter details page

shaun-technovation:
A very basic update here to add chapter name, location and summmary to the chapter's show page. The layout will be basically the same as the particpant's show page.

On the participant's show page the location is shown via `#address_details`, in order to get the same functionality for chapters, I had to add `include ActiveGeocoded` to the chapter's model.


app/views/admin/chapters/show.html.erb
```diff
@@ -1,20 +1,36 @@
...
-        <dd><%= @chapter.organization_name %></dd>
+      <div class="grid__col-4">
+        <dl>
+          <dt>Organization</dt>
+          <dd><%= @chapter.organization_name %></dd>
 
-        <dt>Location</dt>
-        <dd>------</dd>
+          <dt>Chapter</dt>
+          <dd><%= @chapter.name.presence || "-" %></dd>
```
viviancan, line 10: I prefer the Chapter name (program name) as a sort of "title" for the page like in the mockups!  It was kind of hard to spot in the list of other chapter details. 
CleanShot 2024-03-14 at 18 27 20@2x

app/views/admin/chapters/show.html.erb
```diff
@@ -1,20 +1,36 @@
...
+          </dd>
+        </dl>
+      </div>
+
+      <div class="grid__col-12 grid__col--bleed-y">
+        <dl>
+          <dt>Summary</dt>
+
+          <dd>
+            <%= simple_format h(@chapter.summary) %>
```
viviancan, line 30: Do we want some sort of empty state text or "-" like the others?

viviancan (approved):
LGTM! Left a few comments, nothing major 💯 Great work!

app/views/admin/chapters/show.html.erb
shaun-technovation: Good suggestion, I'll add a "-" placeholder for the summary too! 👍

app/views/admin/chapters/show.html.erb
shaun-technovation: I just treat the mockups as guides/suggestions. 😉 😄  Anywho, not a big deal, so I'll make this change. 👍

Issue #4554, opened 2024-03-05
Add additional question around education


dboyer:
In the additional information section of submissions, add a new question:

“Does your submission address a problem in education?”

The answers should be yes and no, and it should have the same text box as the other additional questions.

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/4554?planning-poker) with Zenhub @shaun-technovation @viviancan

dboyer:
Tested on QA and confirmed:

- For a new submission, I see the education question and am able to complete it
- For a submission that was already complete, the education question appears and I am able to complete it
- I can view the education question on the admin page
- As an admin, I can add "solves education" as a new column
- When I export the submission data, the solves education question is there

@viviancan When I tried to edit the education question as an admin, the change did not save. I confirmed I could edit and save changes to the other additional questions.

<img width="1053" alt="Screen Shot 2024-03-14 at 2 43 31 PM" src="https://github.com/Iridescent-CM/technovation-app/assets/516764/8afe74d7-58d0-43fb-9eb9-a1d702e04182">

viviancan:
@dboyer confirming this is on QA

dboyer:
Confirmed on QA that I can now edit the education question as an administrator.

PT also reviewed this feature and it's working as they expected.

PR #4578, opened 2024-03-14
Added missing education params for admin edit team submissions

viviancan:
Added missing education params for admin edit team submissions

Refs #4554


shaun-technovation (approved):
👍

PR #4582, opened 2024-03-15
Update preview deployment

shaun-technovation:
This should make it so that Preview will get deployed via CircleCI, like how QA and Production are currently. Making it so that CircleCI has to pass first before Preview gets deployed (right now that's not how it is).

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -200,6 +200,13 @@ jobs:
             - "32:c4:d8:7f:60:e0:56:7d:e9:ee:f5:04:e1:9c:2d:4e"
 
       - deploy:
+          name: Deploy Preview to Heroku
+          command: |
+            if [ "${CIRCLE_BRANCH}" == "preview" ]; then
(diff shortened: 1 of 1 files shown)


circle.yml
```diff
@@ -199,6 +199,13 @@ jobs:
           fingerprints:
             - "32:c4:d8:7f:60:e0:56:7d:e9:ee:f5:04:e1:9c:2d:4e"
 
+      - deploy:
+          name: Deploy Preview to Heroku
+          command: |
+            if [ "${CIRCLE_BRANCH}" == "preview" ]; then
+              git push https://heroku:$HEROKU_API_KEY@git.heroku.com/technovation-preview.git preview:master
```
shaun-technovation, line 206: This part (Re: `$HEROKU_API_KEY`) that I copied from the QA steps is making me question whether this will work since I'm not sure what the API or where it is, etc. I guess we'll see if it'll work.

viviancan:
LGTM! Feel free to make any updates as needed in preview to get it working 💯 good idea to get preview working like qa!!

viviancan (approved):

PR #4584, opened 2024-03-18
Security: Update follow-redirects node module

shaun-technovation:
This will bump the `follow-redirects` version from `1.15.5` to `1.15.6` to address this security vulnerability:

- https://github.com/Iridescent-CM/technovation-app/security/dependabot/186


viviancan (approved):
👍

PR #4588, opened 2024-03-18
Chapter Ambassador age confirmation

shaun-technovation:
This PR will add the "I confirm that I am 18 years or older" checkbox to the registration process for chapter ambassadors. 

It will also:
- Remove the date of birth requirement for chapter ambassadors
- Let the date of birth be null in the database
- Make updates where date of birth is being expected (like for division assignment, there might be other areas too that I haven't found yet)
- Update specs and factories to accommodate all of this


shaun-technovation:
Just a heads up on this one, I just pushed a commit to rename `meets_age_requirement` to `meets_minimum_age_requirement` (happy to name it something else too), I did a quick test and it was working, I'm hoping it didn't break something unexpected.

viviancan (approved):
I was able to register as a ChA and confirm I was over 18.

Right now when viewing a ChA details on the participant show view (`app/views/admin/participants/show.html.erb`), I get the following error:

```
ActionView::Template::Error (undefined method `month' for nil:NilClass

    current_month_after_birth_month = now.month > date_of_birth.month
                                                               ^^^^^^):
```

It looks like line 47 is breaking here `<dd><%= @account.age %></dd>`

viviancan:
Nice work on this 💯 Looks like it will also work for other profile types 👍

shaun-technovation:
> Right now when viewing a ChA details on the participant show view (`app/views/admin/participants/show.html.erb`), I get the following error:
> 
> ```
> ActionView::Template::Error (undefined method `month' for nil:NilClass
> 
>     current_month_after_birth_month = now.month > date_of_birth.month
>                                                                ^^^^^^):
> ```

Good catch on this, thank you! I updated this!

PR #4594, opened 2024-03-21
Add new judge registration question

shaun-technovation:
This will add everything (models, migrations, controller, view stuff, etc.) for the "As a judge you may call me a...." question.

I also added this info (just a very basic bulleted list) to the admin side so admins can view what was selected on a judge's participant page.


viviancan (approved):
Tried locally and worked as expected! 👍 🚀 Nice addition!

Issue #4591, opened 2024-03-20
Create page for MOU section


dboyer:
This covers creating the page where ChA will access info about the MOU and the link to Docusign. The link to Docusign will be created after completing the research in #4403.

This page should include links to the following documents. Ideally, the user will need to check a box by each document before they can click the button to go to Docusign. Wireframe is below.

- Non-Disclosure Agreement
- Brand Guidelines
- Social Media Admin Agreement
- Social Media Policy
- Technovation Code of Conduct
- Reviewed Safety Page

<img width="943" alt="Screen Shot 2024-03-20 at 11 15 28 AM" src="https://github.com/Iridescent-CM/technovation-app/assets/516764/ea886c3f-2001-48a0-bdd3-bd89e40e855a">

viviancan:
Confirming this is on preview

dboyer:
I see the MOU page, the text, and the checklist of documents!

@viviancan I'm not sure if it matters since we haven't implemented the click functionality yet but the code of conduct is pre-checked for any ChA when I log in.
<img width="1184" alt="Screen Shot 2024-03-28 at 9 35 12 AM" src="https://github.com/Iridescent-CM/technovation-app/assets/516764/06b7ace3-76f5-4c05-9ab0-2efdb55adfc0">

PR #4599, opened 2024-03-21
Adds MOU page for ChA

viviancan:
Refs #4591 

This PR adds the placeholder MOU page for ChA 
I reused/altered the sidenav approach for the checklist. I'm not sure what our technical approach will be for this section. We have discussed having the checkboxes. I also thought we could capture when the document has been clicked and automatically update the list. I made a screen capture of both approaches. This PR has the 2nd approach where the checklist will automatically/dynamically be checked once each link is clicked. Either way the foundation for the page is here and can be updated as needed once we finalize the approach.

Checkbox approach
CleanShot 2024-03-21 at 15 40 34

Auto checked
CleanShot 2024-03-21 at 15 48 45@2x


app/controllers/chapter_ambassador/mous_controller.rb
```diff
@@ -0,0 +1,5 @@
+module ChapterAmbassador
+  class MousController < ChapterAmbassadorController
```
viviancan, line 2: I went with MOU because I thought `memorandum of understanding` would be too long in all the different places it is needed 😅

app/views/completion_steps/rebrand/_mou.html.erb
```diff
@@ -0,0 +1,56 @@
...
+<p class="mb-8">Please confirm that you have read each of these policies and then click the link
+  to sign the Memorandum of Understanding.
+</p>
+
+<div class="ml-8 mb-8">
+  <ol role="list" class="space-y-4">
+    <%= render "application/templates/completion_step_checklist_item",
+               name: "Non-Disclosure Agreement",
+               url: "#",
+               is_complete: false
```
viviancan, line 15: I used `is_complete` but also thought about `is_viewed` `is_clicked`. Open to suggestions.

app/views/completion_steps/rebrand/_mou.html.erb
```diff
@@ -0,0 +1,56 @@
...
+    <%= render "application/templates/completion_step_checklist_item",
+               name: "Social Media Policy",
+               url: "#",
+               is_complete: false
+    %>
+
+    <%= render "application/templates/completion_step_checklist_item",
+               name: "Technovation Code of Conduct",
+               url: "#",
+               is_complete: true
```
viviancan, line 39: Set this to true as an example

app/views/application/templates/_completion_step_checklist_item.erb
```diff
@@ -0,0 +1,19 @@
+<li>
```
viviancan, line 1: I called this a checklist item but also open to naming suggestions 🤔

dboyer:
I prefer the auto-check method in the second approach.

app/controllers/chapter_ambassador/mous_controller.rb
shaun-technovation: +1 for naming it explicitly. I very briefly looked at our controllers and the only abbreviation I noticed is `API` other that that it seems most things are explicit (like we internally reference chapter ambassadors as ChAs, but they're explicitly named chapter_ambassadors in the code). 

Actually this controller name is longer than `memorandum_of_understanding_controller`:
`chapter_ambassador_profile_additions_controller.rb`

Also, randomly, `chapter_ambassador_profile` seems to be about the same length as `memorandum_of_understanding`.

Anywho, definitely not a blocker. 😄

app/views/application/templates/_completion_step_checklist_item.erb
shaun-technovation: I can't think of anything at the moment. But I see that we already have a `_completion_step` partial, which gets displayed in the sidebar and has the completion status functionality... and this partial also has that functionality... And to make things more confusing the sidebar header mentions "checklist", and this has checklist in the name too. I wonder if adding "sub" to this (sub_checklist?), or removing completion_step, or maybe rename completion step to checklist item, and this could be sub-checklist item? 🤔 🤷

app/views/completion_steps/rebrand/_mou.html.erb
shaun-technovation: `is_complete` is generic and sounds good to me. 👍  `is_viewed` or `is_clicked` seems too specific to this MOU use-case and would make the checklist partial less versatile/reusable in my opinion.

shaun-technovation (approved):
Looks good and straight forward to me! 👍 💯 🥳 🚢

app/controllers/chapter_ambassador/mous_controller.rb
viviancan: Ty for the feedback! It def makes sense to be more explicit 💯 I am going to wait on updating this because we recently received new news that ChA may not be signing MOUs in the ChA onboarding experience! I am waiting for confirmation. If we go with MOU i will update it to `memorandum of understanding` and if ChA are signing something different I will make the appropriate update!

PR #4601, opened 2024-03-26
Judge age confirmation

shaun-technovation:
This will add the " "I confirm that I am 18 years or older" for judges and make the date of birth field optional.

A lot of the work here was copied from the chapter ambassador work, which is on the Preview branch and this work here is on the QA branch, so I had to copy a lot of that work.

### DIFF
--- a/db/migrate/20240326155545_age_related_updates.rb
+++ b/db/migrate/20240326155545_age_related_updates.rb
@@ -1,0 +1,12 @@
+class AgeRelatedUpdates < ActiveRecord::Migration[6.1]
+  def up
+    if Account.column_names.exclude?("meets_minimum_age_requirement")
+      add_column :accounts, :meets_minimum_age_requirement, :boolean
+      change_column_null :accounts, :date_of_birth, true
+    end
+  end
+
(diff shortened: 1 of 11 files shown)


db/migrate/20240326155545_age_related_updates.rb
```diff
@@ -0,0 +1,12 @@
+class AgeRelatedUpdates < ActiveRecord::Migration[6.1]
+  def up
+    if Account.column_names.exclude?("meets_minimum_age_requirement")
+      add_column :accounts, :meets_minimum_age_requirement, :boolean
+      change_column_null :accounts, :date_of_birth, true
+    end
```
shaun-technovation, line 6: These changes have already been made on Preview, so I wrapped the changes here in a conditional so this won't get run again.... But actually, as I'm writing this, I wonder if the migration on Preview needs this too since when that work gets merged into QA, QA will already have these changes... 🤔

viviancan (approved):
Did not try locally, but looks pretty straightforward like the other age confirmation! 💯 Great work on this!

PR #4605, opened 2024-03-29
Update ansi-regex node modules

shaun-technovation:
This will update ansi-regex which should address these security vulnerabilities:

- https://github.com/Iridescent-CM/technovation-app/security/dependabot/73
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/72
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/71


viviancan (approved):
👍

PR #4607, opened 2024-03-29
Update decode-uri-component

shaun-technovation:
This will update decode-uri-component to address this security vulnerability:

- https://github.com/Iridescent-CM/technovation-app/security/dependabot/105


viviancan (approved):
👍

Issue #3683, opened 2022-10-21
Add age validation check when returning students log in and offer option to convert to mentor if over participation age


dboyer:
On production, there are some some students (26403 and 4510) who were participants several years ago and logged back in for the new season. Their current age would place them in the mentor category but the platform kept them as students who were able to match with teams.

Perhaps when students log in for the new season, we run a date check and if they are over 18, they receive a message indicating their account will be converted to a mentor account.

shaun-technovation:
> ... they receive a message indicating their account will be converted to a mentor account.

Does this mean we will still manually convert them (via the admin) to be a mentor? Just clarifying, this issue is only to display a flash/notification message?

dboyer:
Good question - I think this issue could cover just running the check and displaying a notification to the user that their age no longer allows them to compete. Perhaps we would want a text box that would ask if they want to have their account converted to a mentor account? Although if they say no, it's not like there is really another option.

dboyer:
Decision after discussion with PT/VET:

1. Implement age validation check when returning student logs into new season
2. If student will be over 18 on Aug 1, show an interrupt screen with two options - go to alumnae resources or convert account to a mentor account
3. Student cannot access their current student account and must choose an option. Or if they just close the screen, they return to the same screen if they log in again.

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/3683?planning-poker) with Zenhub @shaun-technovation @viviancan

dboyer:
Language update to the interrupt screen:

Thank you for participating in Technovation Girls. Since you will be older than 18 on August 1, you are no longer eligible to participate as a student. Choose from the options below to access our alumnae resources or learn more about being a mentor for a team of students.

Button: View Alumnae Resources (link to https://www.technovation.org/alumnae-community-resources/)
Button: Become a Mentor (transition student account to a mentor account)

shaun-technovation:
This one is on QA ready for a looksie. 👍 

And randomly, the verbiage:

> Click the button below to learn more about being a mentor.

The button will convert the student to a mentor, maybe that text needs to be changed? Anywho, just a minor thing. 😄

shaun-technovation:
For QA'ing, I found that if I updated an existing student's birthday in the admin to make them older and then impersonated them, I could test the student -> mentor conversion page.

[...]

PR #4617, opened 2024-04-04
Updates to age related functionality

shaun-technovation:
This will:

- make it so that the `#age` method only returns an age if there is a date of birth set
- update the `#requires_background_check?` method to check for a date of birth before checking if their age is >= 18, and also add new a check to see if they `meets_minimum_age_requirement?` to see if a background check is required for them

### DIFF
--- a/app/models/mentor_profile.rb
+++ b/app/models/mentor_profile.rb
@@ -169,8 +169,8 @@ class MentorProfile < ActiveRecord::Base
     prefix: true,
     allow_nil: true
 
-  def method_missing(method_name, *args)
-    account.public_send(method_name, *args)
+  def method_missing(method_name, *args) # standard:disable all
+    account.public_send(method_name, *args) # standard:disable all
   end
 
   def training_required?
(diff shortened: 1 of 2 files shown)


app/models/mentor_profile.rb
```diff
@@ -295,7 +295,7 @@ def background_check_complete?
   end
 
   def requires_background_check?
-    (account.valid? && account.age >= 18) &&
+    (account.valid? && (account.date_of_birth.present? && account.age >= 18 || account.meets_minimum_age_requirement?)) &&
```
shaun-technovation, line 298: I'm contemplating removing this "age" check b/c I realized it also needs to happen at the Vue layer too (for the sidenav).

Right now we're checking if a mentor is over 18 and in a background check country, but all new mentors (and judges) would be over 18 since that's a requirement to register. I *think the only way for this to happen is if **we** convert a student to a mentor? 🤔 Which if the requirement is that they are over 18 I don't know why we would do that? @dboyer? (We could prevent that from happening from the admin side too.)

Anywho, this PR I think would fix things breaking, but the mentor sidenav could display the wrong info on whether a background check is required (since that's in Vue, and the fix/patch here is in Rails), this would be for new judge/mentor accounts, since some of them won't have a DOB.

Or maybe we can leave it, and then fix the Vue sidenav issue after this (and any other issues that popup from the missing DOB).... 🤔

dboyer:
I will check on whether admins should be able to convert students to mentors if the mentor is under 18! We did add this age validation check recently #3683 but that's more to keep users in the student category if they are over 18.

Issue #4532, opened 2024-02-26
Indicate primary contact for a chapter


dboyer:
Each chapter should have one chapter ambassador who is identified as the primary contact and whose info is displayed to the students when selecting a chapter. This issue covers adding a new "primary contact" field to the chapter model.

The ChA for a chapter will choose who is the primary contact as part of the chapter ambassador onboarding. The visual interface for selecting the primary contact is covered under #4611.

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/4532?planning-poker) with Zenhub @shaun-technovation @viviancan

dboyer:
Confirming I now see the primary contact option.

shaun-technovation:
@viviancan Just curious, did the migrations for this run as expected when this got deployed to Preview? (I think this might be the first thing merged into Preview that had a migration since #4581.)

PR #4625, opened 2024-04-10
V1 - Migration to add primary chapter ambassador to chapters table

viviancan:
Refs #4532

Migration to add primary chapter ambassador to chapters table


viviancan:
> Adding a field to the `chapter` table is the way to go I think. In my opinion, a primary contact only makes sense in the context of a chapter; a primary contact is **for** a chapter is the way I think of it, and being able to do `chapter.primary_contact` (or `primary_chapter_ambassador`) which would return the ChA makes sense to me.

woohoo!  thx @shaun-technovation I will continue with this path then 💯

viviancan:
Closing this - it was handled in #4639

Issue #4626, opened 2024-04-11
New ChA account prompted for chapter location but cannot enter it


dboyer:
As part of testing #4608, I ran into the following issue:

- Went to a chapter admin page and used the email invite section to invite a new ChA to that chapter (/admin/chapters/12)
- Clicked the link in the invite email and created a new account for that ChA, making sure to use the email address I sent the invite to (/admin/participants/1620)
- The ChA dashboard loaded at the Chapter Profile tab with the Location Details section open. There's a prompt to complete location but no way to do so
- I cannot access any other tabs including the dashboard for ChA onboarding as it won't let me leave this screen

I reproduced the same issue with the participant /admin/participants/1621

<img width="1292" alt="Screen Shot 2024-04-11 at 10 31 32 AM" src="https://github.com/Iridescent-CM/technovation-app/assets/516764/f81350df-8ad9-4b7f-baa8-1e4a64c15f6c">

viviancan:
Confirming the fix is on QA (@dboyer)

dboyer:
Confirming I no longer run into this issue on preview.

PR #4631, opened 2024-04-11
Fixes issue where setting/updating location not available for ChAs

viviancan:
Refs #4626 

- Reverted original "location details" logic/routes so ChAs can set and update their location
- Added `chapter_location` route/controller so chapter location can eventually be set/updated

These change resolve the bug described in #4626.


shaun-technovation (approved):
Didn't try locally, but looks fine to me! 👍

PR #4633, opened 2024-04-12
Update ini node module

shaun-technovation:
This will update all versions of the ini node module to `1.3.8` which should address this security vulnerability:

- https://github.com/Iridescent-CM/technovation-app/security/dependabot/1


viviancan (approved):
👍

PR #4635, opened 2024-04-12
Update tar node module

shaun-technovation:
This will update the tar node module (and its dependencies) to address this security vulnerability:

- https://github.com/Iridescent-CM/technovation-app/security/dependabot/190


viviancan (approved):
👍

PR #4638, opened 2024-04-15
WIP: Update migration for meets_minimum_age_requirement

shaun-technovation:
This should ensure that this migration will only run if the `meets_minimum_age_requirement` column doesn't exist on QA or Preview. I'm pretty sure this is needed because both QA and Preview have this migration (one was added on QA for judges and one for chapter ambassadors on Preview), and this should ensure the column will only be added one time.


db/migrate/20240318172542_meets_age_requirement_to_accounts.rb
```diff
@@ -1,5 +1,8 @@
 class MeetsAgeRequirementToAccounts < ActiveRecord::Migration[6.1]
-  def change
-    add_column :accounts, :meets_minimum_age_requirement, :boolean
+  Account.reset_column_information
+  if Account.column_names.exclude?("meets_minimum_age_requirement")
```
shaun-technovation, line 3: I used this Stack Overflow post for the conditional migrations (I have never done conditional migrations before):
https://stackoverflow.com/questions/4980895/how-to-write-conditional-migrations-in-rails

viviancan (approved):
Did not try locally, but did review the post! Nice find 💯

Issue #4572, opened 2024-03-14
Display judge type as a filter and additional column on the participants page


dboyer:
Front-end created in #4571 

1. On the participants page, add a filter for "judge type" that works similar to the mentor type filter
2. On the participants page, add a "judge type" option in the more columns section

dboyer:
May also need to be on judges admin page

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/4572?planning-poker) with Zenhub @shaun-technovation @viviancan

PR #4646, opened 2024-04-17
Added judge type as a filter and column to participants grid

viviancan:
Refs #4572 

Added judge type as a filter and column to participants grid. I followed the mentor types approach!


shaun-technovation (approved):
LGTM! 👍

PR #4651, opened 2024-04-18
DocuSign API service to send MOU

shaun-technovation:
This will add a new DocuSign API service that can has functionality to send someone an MOU. The service has a new method `send_memorandum_of_understanding`, pass this method a name, email address, organization name, and optional job title; and it will send them (the email address) an MOU to sign via DocuSign.


app/services/docusign/api_client.rb
```diff
@@ -0,0 +1,142 @@
...
+                    italic: true,
+                    underline: true,
+                    value: organization_name,
+                    width: 12,
+                    height: 23,
+                    xPosition: 98,
+                    yPosition: 180
+                  },
+                  {
+                    anchorString: "Name of Organization (if applicable):",
```
shaun-technovation, line 71: For the tabs that have an `anchorString`, DocuSign will add a tab (aka field) where ever the `anchorString` appears in the document, that's why I've included the colons at the end so the anchors are more specific. 

More info on anchors:

> Anchor tagging enables you to place tabs at every location where a specified text string, also known as an anchor, is found in a document. When tabs are added with anchor tagging, DocuSign searches the document for instances of an anchorString property that you provide. For each found instance, it places a tab of the specified type for the designated recipient. Tab positions in relation to the string instances can be set by providing x and y offsets.

https://developers.docusign.com/docs/esign-rest-api/esign101/concepts/tabs/auto-place/

app/services/docusign/authentication.rb
```diff
@@ -71,7 +71,7 @@ def get_access_token_from_refresh_token
     def process_response(response)
       response_body = JSON.parse(response.body, symbolize_names: true)
 
-      if response_body[:access_token].present?
+      if response.success?
```
shaun-technovation, line 74: A small update to the authentication system, Faraday has this nice `success?` helper that I wanted to use here.

viviancan (approved):
Woohoo! This is great! Can't wait to see it in action 💯

Issue #4642, opened 2024-04-16
Add UI so ChAs can select to be or not to be visible on the map


viviancan:

dboyer:
Mockup of the UI for opting-out of appearing on the map

<img width="758" alt="Screen Shot 2024-04-16 at 12 08 06 PM" src="https://github.com/Iridescent-CM/technovation-app/assets/516764/e3bc1f0f-721e-4aad-8a5a-80e83c4b034a">

viviancan:
Note: this also includes migration for chapter table (something like visibile_on_map)

viviancan:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/4642?planning-poker) with Zenhub @shaun-technovation

viviancan:
Confirming this is on PREVIEW

I tried with [this chapter](https://technovation-preview-28bd069458f9.herokuapp.com/admin/chapters/12#/). By default they were set to "display on map." I logged in a a ChA and was able to set it to "No, do not display." I verified the text on the ChA Pubic info page, the chapter details page, and also the filter on the chapter datagrid.

dboyer:
Tested on preview and confirmed:

- the default status is "yes"
- As a ChA, I could change the status to no and see it was updated
- As an admin, I can see the status update on the chapter admin page
- As an admin, I can add the "visible on map" filter on the chapter admin page
- As an admin, I can export the list of chapters and see the column for map visibility with true/false values listed correctly

PR #4653, opened 2024-04-18
Add ability for ChAs to update their chapter's public map visibility status

viviancan:
Refs #4642 

This PR adds the following
- Ability for ChAs to change their chapter public visibility setting
- Basic specs for chapter public info features

CleanShot 2024-04-17 at 16 01 25@2x

### DIFF
--- a/app/data_grids/chapters_grid.rb
+++ b/app/data_grids/chapters_grid.rb
@@ -30,6 +30,10 @@ class ChaptersGrid
       "#{value}%")
   end
 
+  filter(:visible_on_map, :xboolean)
+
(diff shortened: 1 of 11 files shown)


app/data_grids/chapters_grid.rb
```diff
@@ -30,6 +30,10 @@ class ChaptersGrid
       "#{value}%")
   end
 
+  filter(:visible_on_map, :xboolean)
```
viviancan, line 33: I didn't know about this option from datagrid!
`xboolean - the select of "yes", "no" and any`

PR #4654, opened 2024-04-18
Legal contact for chapters

shaun-technovation:
This will add all the necessary pieces so admins can add legal contacts to chapters.


app/views/admin/chapters/_chapter_ambassadors.html.erb
```diff
@@ -0,0 +1,56 @@
+<div class="grid">
```
shaun-technovation, line 1: This file is extracting/moving the "chapter ambassadors" section on the chapters page to this new file.

viviancan (approved):
Lgtm! Nice work on this 💯🤘

PR #4658, opened 2024-04-19
Update word-wrap node module

shaun-technovation:
This will update word-wrap from version `1.2.3` to `1.2.5` to address this security vulnerability:

- https://github.com/Iridescent-CM/technovation-app/security/dependabot/160


viviancan (approved):
👍

PR #4659, opened 2024-04-19
WIP: Update Filestack

shaun-technovation:
This will update Filestack from version `3.22.0` to version `3.25.0`, which should address this security vulnerability:

- https://github.com/Iridescent-CM/technovation-app/security/dependabot/150


viviancan (approved):
👍

Issue #4662, opened 2024-04-22
Support: Judge profiles who were previously ChAs cannot be added to an RPE


viviancan:

viviancan:
We had a [support request](https://iridescentlearning.slack.com/archives/C06CHBQ4E/p1713800651787139) come in where a ChA could not add a judge (by email or name) to an RPE. After investigating further, it appears that this judge was previously a ChA. I looked for ways a ChA could be converted to a judge and the only way I could reproduce this was through the following steps:
1. As a ChA click the "mentor mode" button to create a mentor profile
2. Convert the mentor account to a "judge only" account. In order to do this, they must not be on any teams. This also deletes their mentor profile
3. As a ChA2 (the one with the RPE event) try to add ChA1 to the event. They will not display in the list of judges even though they have a judge profile

I believe this is happening due to this scope
```
 scope :live_event_eligible, ->(event) {
    includes(:judge_profile)
      .references(:judge_profiles)
      .left_outer_joins(:mentor_profile, :chapter_ambassador_profile)
      .where(
        "judge_profiles.id IS NOT NULL AND " +
        # temporary fix, see: https://github.com/Iridescent-CM/technovation-app/issues/2111
        # "mentor_profiles.id IS NULL AND " +
        "chapter_ambassador_profiles.id IS NULL"
      )
  }
```

[This PR](https://github.com/Iridescent-CM/technovation-app/pull/4663) updates the query to include an `OR` condition to include judges who have a ChA profile with a pending status (meaning they are not an active chapter). Something to keep in mind is that the ChA status functionality is(or may have already been) removed from the ChA functionality.

It also appears that there was a similar issue related to joint judge-mentor accounts noted [here](https://github.com/Iridescent-CM/technovation-app/issues/2099)  and [here](https://github.com/Iridescent-CM/technovation-app/issues/2111)

viviancan:
Confirming this in on QA

dboyer:
Tested on QA using the following steps:

- Created a new ChA account and clicked to mentor mode to give it a mentor profile
- Kept that ChA account as "pending" as a ChA
- While logged into the admin, clicked "convert to a judge only account" on that participant's profile
- Was able to assign that participant to an event

I also tried to add a chapter ambassador with an active ChA account to an event and was not able to do so.

PR #4663, opened 2024-04-22
Updated the live_event_eligible account scope to include chapter ambassador profiles that have a pending status

viviancan:
Refs #4662 

I updated the live_event_eligible account scope to include chapter ambassador profiles that have a status of pending. I will add a more detailed write-up on the ticket.


shaun-technovation (approved):
The change looks good/straight forward to me. 👍 🚢  The only thing on my mind if there is going to be other side effects/impacts of making this change. 🤔

Issue #4643, opened 2024-04-16
As a ChA you are able to save a program summary longer than the word count limit


viviancan:

viviancan:
Refs #4616

dboyer:
Tested on preview and confirmed that I could not add a program summary that was over the word count.

PR #4668, opened 2024-04-24
Adds length validation for chapter summary

viviancan:
Refs #4643 

Adds validation to chapter summary so that a summary can be a max of 280 characters


shaun-technovation (approved):
Nice fix! LGTM! 👍 🥳 👏 🚀

PR #4695, opened 2024-05-01
Add host to Saleforce API calls

shaun-technovation:
By default if a host is not specified, Restforce will use the production host (`login.salesforce.com`), for the sandbox we need to use `test.salesforce.com`. The changes in this PR will allow us to specify which host to use so we can set it for the sandbox environment.


viviancan (approved):
💯 LGTM

PR #4703, opened 2024-05-01
Install Scout APM gem

shaun-technovation:
This will install Scout to hopefully help us troubleshoot the memory issues on Production.

From the [setup guide](https://scoutapm.com/docs/ruby/setup), no configuration should be necessary for Production or QA, since those environments already have the Scout add-on.

> Heroku Customers: We initially provision your application using environment variables instead of the `config/scout_apm.yml`. Configuration values set via env variables will take priority over those set via yaml.

I checked both QA and Production and I see ENV settings for `SCOUT_KEY`, `SCOUT_LOG_LEVEL`, and `SCOUT_MONITOR`, so I think adding the gem is all that we need.

There's also a section of the setup guide on how to use Scout locally, if we want to add/configure that later:
https://scoutapm.com/docs/ruby/setup#local

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -58,6 +58,7 @@ gem "redis"
 gem "sidekiq", "~> 7.0"
 
 gem "airbrake", "~> 11.0.3"
+gem "scout_apm"
 
 gem "dalli", "~> 3.0"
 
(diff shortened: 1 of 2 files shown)


viviancan (approved):
Did not try locally, but it looks like a good addition for the memory issue! Great thinking 💯

Issue #4678, opened 2024-04-29
Add additional filters and columns to judge admin page


dboyer:
Two requests to assist with score monitoring:

1.  In #4572, judge type was added as a filter and additional column on the participants page. VET would like the same filter and column to be available on the judges admin page at /admin/judges
2. Also add the "virtual or live" filter and additional column from the participants admin page to the judge admin page

viviancan:
This is on QA

dboyer:
Tested on QA and confirmed that I could:

- Filter by judge type
- Add judge type as a column
- Export judge type information
- Filter by virtual or live
- Add virtual or live as a column
- Export virtual or live information

PR #4704, opened 2024-05-01
Added virtual or live column & judge types column/filter to judge datagrid

viviancan:
Refs #4678 
Added virtual or live column & judge types column judge datagrid

### DIFF
--- a/app/data_grids/judges_grid.rb
+++ b/app/data_grids/judges_grid.rb
@@ -148,7 +148,21 @@ class JudgesGrid
         "judge_profiles.onboarded = ?",
         value == "onboarded"
       )
-    end
+  end
+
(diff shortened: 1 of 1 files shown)


app/data_grids/judges_grid.rb
```diff
@@ -27,6 +27,22 @@ def self.current_judging_round
   column :last_name, mandatory: true
   column :email, mandatory: true
 
+  column :judge_types do
+    if judge_profile.present?
```
shaun-technovation, line 31: Hmm, since this is the judge datagrid, wouldn't everyone have a `judge_profile`? 🤔

shaun-technovation (approved):
Looks good/straight forward to me! 👍

PR #4705, opened 2024-05-02
Only update registration invite if an invite code exists

shaun-technovation:
This simple update will make it so that the `UpdateRegistrationInviteJob` will only get called if an invite code exists. I think this should help with all the failed jobs that don't have an invite code.

Randomly, I had already made this change on Preview, but of course isn't on QA/Production, doh!

https://github.com/Iridescent-CM/technovation-app/blob/16846f1df092d60197600f75a0ef0466c9844ea6/app/controllers/new_registration_controller.rb#L161-L166

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -58,6 +58,7 @@ gem "redis"
 gem "sidekiq", "~> 7.0"
 
 gem "airbrake", "~> 11.0.3"
+gem "scout_apm"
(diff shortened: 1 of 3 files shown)


viviancan (approved):
Nice update!! I think it will def help with the failed jobs 💯

Issue #4698, opened 2024-05-01
Continue adding additional fields for chapter program info form


viviancan:
- [x] Program Length - Single Selection
- [x] Participant Count - Single Selection
- [x] Low Income - Single Selection
- [x] Organization Type - multi
- [x] Meeting time - multi
- [x] Meeting facilitator - multi

viviancan:
Program length is in code review

viviancan:
Participant count in code review

viviancan:
low income in code review

viviancan:
organization type in code review

viviancan:
meeting time and meeting facilitator is in [code review](https://github.com/Iridescent-CM/technovation-app/pull/4723)

viviancan:
@dboyer Confirming this is on PREVIEW

dboyer:
Confirmed that all these fields are on preview. Looks great!!! I could:

- Add and edit fields
- Save without completing every field
- Return and edit existing answers later
- Login as a different ChA for the chapter and see the same info
- Edit that info as a different ChA

Issue #4644, opened 2024-04-16
Add fields on program info page


dboyer:
PT has confirmed the fields for the program info page. The list of fields is available [on this doc](https://docs.google.com/document/d/1vQ_tewXkVJOKMif_aoDtDgVRE-1bo-tj9XIZ_nqrzKs/edit#heading=h.w48fwa0pa51) and in the wireframe below.

![Chapter-Ambassador-Onboarding-2024-•-Technovation-Girls](https://github.com/Iridescent-CM/technovation-app/assets/516764/483401fb-7baf-469f-8a6e-e965b6289fc2)

dboyer:
Note: Questions 8-10 should be a multi-select

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/4644?planning-poker) with Zenhub @shaun-technovation @viviancan

dboyer:
Confirmed that all these fields are on preview. Looks great!!! I could:

- Add and edit fields
- Save without completing every field
- Return and edit existing answers later
- Login as a different ChA for the chapter and see the same info
- Edit that info as a different ChA

PR #4706, opened 2024-05-03
Add text and date fields for Chapter program info

viviancan:
Refs #4644 

Adds basic functionality for ChAs to create/edit their chapter program information needed for PT support. This only includes text and date fields. #4698 will add the other fields

CleanShot 2024-05-03 at 11 06 37@2x

CleanShot 2024-05-03 at 11 06 48@2x

### DIFF
--- a/app/inputs/date_time_input.rb
+++ b/app/inputs/date_time_input.rb
@@ -1,0 +1,5 @@
+class DateTimeInput < SimpleForm::Inputs::DateTimeInput
+  def input(wrapper_options)
+    template.content_tag(:div, super, class: "custom-simpleform-date")
+  end
+end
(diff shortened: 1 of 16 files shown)


app/inputs/date_time_input.rb
```diff
@@ -0,0 +1,5 @@
+class DateTimeInput < SimpleForm::Inputs::DateTimeInput
+  def input(wrapper_options)
+    template.content_tag(:div, super, class: "custom-simpleform-date")
```
viviancan, line 3: This was needed in order to add a custom style to the datepicker. From the docs.

shaun-technovation (approved):
👍

PR #4710, opened 2024-05-06
Update rake task that converts ChA links to chapter links

shaun-technovation:
A very simple change that will make it so that we only find and loop over unique chapter ambassadors.

Before this change, the number of times it would loop over a chapter ambassador would be the equal to the number of links the chapter ambassador had, e.g. if a chapter ambassador had five links it would loop over that chapter ambassador five times.


viviancan (approved):
Did not try locally, but looks straightforward! Nice update 💯

PR #4711, opened 2024-05-07
Add program length (single selection) to program info 

viviancan:
Refs #4698 & #4644 

This PR will add the "program length" single selection to the chapter program info form


shaun-technovation (approved):
Looks like `order` is being set, but I couldn't find where it's being used, unless it's being used implicitly somewhere. 🤔 Either way, seems fine to me! 👍

viviancan:
> Looks like `order` is being set, but I couldn't find where it's being used, unless it's being used implicitly somewhere. 🤔 Either way, seems fine to me! 👍

Ty! I will remove 👍

PR #4715, opened 2024-05-10
DocuSign history, webhook and improvements

shaun-technovation:
This will setup the DocuSign webhook (locking down the webhook endpoint will be future work). 

- When sending the initial request to DocuSign for a new MOU, we will capture the envelope id in a new `documents` table (which will used to keep track of who signed the MOU and when)
- There is a new job `ProcessDocusignWebhookPayloadJob` that will process the webhook payload
  - This job will find the `document` with the envelope id matching the webhook payload, and update the `signed_at` and `season_signed` fields
- In the admin area on the chapter's details page, we'll display the signed at date

Other changes:
- Making legal contacts a separate entity (instead of having the values right on the chapter model)
  - Adding the corresponding routes, controller, model, views, etc.
- Making the legal contact's full name and email address required
- Updating the existing DocuSign service to accept a legal contact
- Adding a base controller and functionality for future webhooks


app/controllers/webhooks/base_controller.rb
```diff
@@ -0,0 +1,11 @@
+class Webhooks::BaseController < ApplicationController
+  skip_forgery_protection
+
+  before_action :save_webhook_payload, only: :create
+
+  private
+
+  def save_webhook_payload
+    @webhook_payload ||= WebhookPayload.create!(body: request.body.read)
+  end
```
shaun-technovation, line 10: Some notes regarding future webhooks, they should inherit from this base controller, this base controller will:
- disable CSRF protection (since these requests aren't coming from our platform)
- save the webhook payload, which can be accessed via `@webhook_payload` (and then a job should process the payload)

app/jobs/process_docusign_webhook_payload_job.rb
```diff
@@ -0,0 +1,31 @@
...
+    docusign_envelope_completed_at = parsed_payload.dig(:data, :envelopeSummary, :completedDateTime)
+    document = Document.find_by(docusign_envelope_id: docusign_envelope_id)
+
+    if document.present? && docusign_envelope_status == "completed"
+      document.update(
+        signed_at: docusign_envelope_completed_at,
+        season_signed: Season.current.year
+      )
+
+      webhook_payload.delete
```
shaun-technovation, line 23: Future webooks should have a job to process the payload, for DocuSign we're just seeing if the document has been "completed" and marking the corresponding `document` record as signed (via `signed_at` and `season_signed`). Of course each webhook would be doing something different.

One thing to call out here is the the saved webhook payload record in the database is being deleted after it has been processed. We could also have a status column and mark them as completed/processed, but I couldn't see a good reason to keep the records around after they have been processed. Anywho, just flagging this as something to keep in mind and maybe something we'll have to revisit.

app/jobs/process_docusign_webhook_payload_job.rb
```diff
@@ -0,0 +1,31 @@
...
+    else
+      error = "[DOCUSIGN WEBHOOK] Couldn't process payload with envelope id: #{docusign_envelope_id}"
+
+      logger.error(error)
+      error_notifier.notify(error)
```
shaun-technovation, line 28: Also, noting that if a DocuSign document has a different status other than "completed" then it won't get processed, we'll only log a message for it. I'm not sure if/when this would happen, but this would be followup work to handle other statuses if we run into them.

app/models/legal_contact.rb
```diff
@@ -0,0 +1,9 @@
+class LegalContact < ActiveRecord::Base
+  belongs_to :chapter, optional: true
+
+  has_one :legal_document, -> { where(active: true) }, class_name: "Document", as: :signer
+  has_many :documents, as: :signer
```
shaun-technovation, line 5: These two lines (line 4 & 5) will change based on the final name for the document. There will be a singular reference/association for the active/current document, and a plural reference/association to get all the documents. This is similar to how the parental consent and/or media consents work.

app/views/admin/chapters/_legal_contact.html.erb
```diff
@@ -5,43 +5,37 @@
     </h3>
 
     <div class="panel">
-      <% if chapter.legal_contact_full_name.present? ||
-          chapter.legal_contact_email_address.present? ||
-          chapter.legal_contact_phone_number.present? ||
-          chapter.legal_contact_job_title.present?
-      %>
+      <% if chapter.legal_contact.present? %>
```
shaun-technovation, line 8: Making legal contact a separate entity really simplifies this logic^.

app/views/admin/chapters/_legal_contact.html.erb
```diff
@@ -5,43 +5,37 @@
...
         <div style="display: flex; justify-content: flex-start; column-gap: 2rem;" >
           <div>
             <%= link_to "Edit",
-              edit_admin_chapter_path(@chapter),
+              edit_admin_chapter_legal_contact_path(@chapter),
               class: "button"
             %>
           </div>
 
-          <% if chapter.legal_contact_email_address.present? %>
```
shaun-technovation, line 35: Having legal contact be separate also makes this unnecessary (b/c legal contacts can't exist w/o an email address).

viviancan (approved):
Amazing work on all of this. I am still processing/understanding all of the changes but happy to have this on preview to get additional eyes on it!! 💯 ⭐  Great job, Shaun!!

PR #4723, opened 2024-05-13
Add multiple checkboxes form questions for meeting times and meeting facilitator

viviancan:
Refs #4698 

This is the final PR for the main work for the chapter program information form. This PR adds the multiple checkboxes form questions for meeting times and meeting facilitator

### DIFF
--- a/app/views/chapter_ambassador/chapter_program_information/_form.en.html.erb
+++ b/app/views/chapter_ambassador/chapter_program_information/_form.en.html.erb
@@ -75,6 +75,14 @@
   </div>
 
   <div class="mb-8">
+    <%= f.association :meeting_times,
+                      as: :check_boxes,
+                      label_method: :time,
+                      label: t("chapters.program_information.meeting_time_question") %>
+  </div>
+
+
+  <div class="mb-8">
     <%= f.association :program_length,
                       collection: ProgramLength.all,
                       as: :select,
@@ -82,6 +90,13 @@
                       required: false,
                       label: t("chapters.program_information.program_length_question"),
                       prompt: "Select a Program Length" %>
+  </div>
+
+  <div class="mb-8">
+    <%= f.association :meeting_facilitators,
+                      as: :check_boxes,
+                      label_method: :name,
+                      label: t("chapters.program_information.meeting_facilitator_question") %>
   </div>
 
   <div class="mb-8">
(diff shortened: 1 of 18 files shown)


db/migrate/20240513150037_create_chapter_program_information_meeting_times.rb
```diff
@@ -0,0 +1,10 @@
+class CreateChapterProgramInformationMeetingTimes < ActiveRecord::Migration[6.1]
+  def change
+    create_table :chapter_program_information_meeting_times do |t|
+      t.references :chapter_program_information, foreign_key: {to_table: :chapter_program_information}, index: {name: "index_cpi_mt_on_chapter_program_information_id"}
```
viviancan, line 4: Originally when running this migration I received this error:
```
ArgumentError: Index name ‘index_chapter_program_information_meeting_times_on_chapter_program_information_id’ on table ‘chapter_program_information_meeting_times’ is too long; the limit is 63 characters
```
So I added a custom index name, but then encountered this error
```
Caused by:
PG::DuplicateTable: ERROR:  relation “chapter_program_information_id” already exists
```
So I decided to go with a custom abbreviated index name

db/migrate/20240513182351_create_meeting_facilitators.rb
```diff
@@ -0,0 +1,10 @@
+class CreateMeetingFacilitators < ActiveRecord::Migration[6.1]
+  def change
+    create_table :meeting_facilitators do |t|
+      t.string :name
+      t.integer :order
```
shaun-technovation, line 5: I think I mentioned this in another PR, but it doesn't look like `order` is being used anywhere? 🤔

app/views/chapter_ambassador/chapter_program_information/_form.en.html.erb
```diff
@@ -74,6 +74,14 @@
                 } %>
   </div>
 
+  <div class="mb-8">
+    <%= f.association :meeting_times,
+                      as: :check_boxes,
+                      label_method: :time,
+                      label: t("chapters.program_information.meeting_time_question") %>
```
shaun-technovation, line 81: The other thing I'm noticing and this is definitely not a blocker is the inconsistency w/ naming, here we reference `chapters.program_information`, but the model reference is `chapter.chapter_program_information`.

shaun-technovation (approved):
Left a couple of comments, but no blockers, LGTM! 👍

db/migrate/20240513182351_create_meeting_facilitators.rb
viviancan: I removed order!

app/views/chapter_ambassador/chapter_program_information/_form.en.html.erb
viviancan: Updated this 👍

PR #4733, opened 2024-05-17
Updates for completed, recused scores

shaun-technovation:
Somehow its possible for a score to be both completed and recused. This will make two changes related to this:
- If a score is both recused and completed, that score will only get displayed under the recused section
- When a score is recused and it happens to be completed, the score will get marked as incomplete and unapproved
  - This will help prevent scores from getting into this state, I also think it would still be possible for a judge to recuse themselves and then we reassign them the same submission


viviancan (approved):
Did not try locally, but looks pretty straightforward! Nice quick fix on this!! 💯

PR #4735, opened 2024-05-17
Restrict access to LearnWorlds

shaun-technovation:
This should restrict access to LearnWorlds so that only people who already have a LearnWorlds account will have access; no new LearnWorld accounts should get created after this update.


viviancan (approved):
Did not try locally! Nice quick win to restrict LearnWorlds access!! 💯

PR #4789, opened 2024-06-13
HMAC security for DocuSign webhook payloads 

shaun-technovation:
This will add a new service that will verify webhook payloads with the HMAC standard. I think this service should be able to be reused for future webhooks.

For DocuSign, the payload will get verified before being saved to the database and being processed by the background job.


viviancan (approved):
Nice work on this 💯

Issue #4699, opened 2024-05-01
SurveyMonkey training integration


dboyer:
Basics of SurveyMonkey integration:

1. ChA can start quiz via SurveyMonkey
2. They must complete quiz in order to have the training onboarding step "checked"
3. Results of quiz do not determine if they passed - just need to complete
4. ChA needs to see correct answers to quiz and which they got wrong

Note: Also on the training page will be either links to videos/slides or potentially embedded videos/slides (similar to judges and mentors).

dboyer:
In dev priorities on 6/3, PT confirmed:

- Training would include five slideshows the ChA would be encouraged to do sequentially
- Preference to have them open/accessible in the platform rather than via a separate website or on a new page but not embedded in a way that would make the page long and scrolling
- Chapter team will create a quiz in SurveyMonkey, as long as the ChA takes the quiz we mark training as “complete” and they can proceed with onboarding
- Ideally the answers to the quiz are accessible after it is done

dboyer:
Confirmed via a conversation with PT on June 5:

- Some data such as quiz ID and score can be stored in SurveyMonkey and do not need to be stored on the platform
- Either embedding or opening the checkpoint in a module are acceptable

viviancan:
Overview
I have been working on integrating SurveyMonkey into our platform and have explored several approaches. I am feeling slightly unsure of which path to take moving forward. 

**Embed Script from SurveyMonkey**
- Uses the provided embed script to integrate the SurveyMonkey form directly on the platform page.
- Pros: Simple to implement; direct integration, script given by survey monkey
- Cons: Unable to send custom variables (e.g., user ID, email) to SurveyMonkey; confirmed with SurveyMonkey support.

**Web Link Collector**
- Directs users to take the quiz via a separate tab using a web link. Upon completion, users are redirected back to the platform and marked as completed quiz (similar to mentor/judge flow)
- Pros: Officially supported by SurveyMonkey; allows the inclusion of custom variables.
- Cons: Redirects users away from the platform which may impact user experience.

**Web Link in an Iframe (Unofficial Approach)**
- Embeds the web link within an iframe on the platform page.
- Pros: Allows custom variables to be sent; keeps users on the platform.
- Cons: Not officially supported by SurveyMonkey. I found this approach on some StackOverflow posts. Multiple users cannot take the survey on the same browser without clearing cookies. Confirmed with SurveyMonkey support that cookies cannot be disabled.

[...]

PR #4790, opened 2024-06-13
Adds survey monkey training checkpoint to Chapter Ambassador onboarding tasks

viviancan:
Refs #4699
This PR adds Survey Monkey to the chapter ambassador onboarding tasks. 

I am opening a separate ticket for the training modules.


app/helpers/survey_monkey_helper.rb
```diff
@@ -0,0 +1,13 @@
+module SurveyMonkeyHelper
+  def chapter_ambassador_training_checkpoint_link(account)
+    base_url = "https://www.surveymonkey.com/r/DCM3VZX"
```
viviancan, line 3: I think this should be an env variable since it will change for production

app/controllers/chapter_ambassador/training_completions_controller.rb
```diff
@@ -3,5 +3,12 @@ class TrainingCompletionsController < ChapterAmbassadorController
     skip_before_action :require_chapter_and_chapter_ambassador_onboarded
 
     layout "chapter_ambassador_rebrand"
+
+    def show
+      current_ambassador.complete_training!
```
viviancan, line 8: Following the same approach as the judge/mentor training

app/views/admin/participants/_chapter_ambassador_onboarding.html.erb
```diff
@@ -44,9 +44,19 @@
         </div>
       </dd>
 
-      <dt>Take Training Quiz</dt>
+      <dt>Take Training Checkpoint</dt>
```
viviancan, line 47: We are officially going with the terminology "checkpoint" instead of quiz. Would this be better as "Complete Training" since completing training implies the checkpoint has also been completed.

app/helpers/survey_monkey_helper.rb
viviancan: Will update!

app/controllers/chapter_ambassador/training_completions_controller.rb
viviancan: 👍 Good point!

app/helpers/application_helper.rb
```diff
@@ -28,6 +28,7 @@ def determine_manifest
       profiles
       chapter_program_information
       public_information
+      trainings
       training_completions
```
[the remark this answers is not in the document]
viviancan: 👍 Will update!

app/helpers/survey_monkey_helper.rb
```diff
@@ -0,0 +1,13 @@
+module SurveyMonkeyHelper
+  def chapter_ambassador_training_checkpoint_link(account)
```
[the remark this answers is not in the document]
viviancan: I agree that there aren't any other references to checkpoint in the back-end. I think calling it "training" link isn't accurate because the actual training (with the training modules will be on the platform) while this is the link to the checkpoint (quiz) that is on survey monkey. I do like adding in the "external" keyword so it is very clear that it is off platform!

PR #4794, opened 2024-06-14
Fix n+1 query on participants datagrid

shaun-technovation:
Selecting the "team matching" option on the participants datagrid is causing an n+1 query, this update will address that.

I found this using the [Bullet](https://github.com/flyerhzm/bullet) gem.


viviancan (approved):
Nice fix!!

PR #4795, opened 2024-06-17
Chapter Ambassador onboarded flag

shaun-technovation:
This will make the chapter ambassador onboarding flag a column in the database, instead of calculating it on the fly every time it is called.

This will also add training to the onboarding requirements.


viviancan (approved):
Works as expected! Awesome job on this! 

After I completed all of the onboarding steps I was marked as onboarded!

Issue #4792, opened 2024-06-14
Add training module links/view to ChA training tab


viviancan:

viviancan:
This is on PREVIEW

dboyer:
Confirming that I see links to the five training modules on preview and am able to click and view each module.

PR #4797, opened 2024-06-20
Adds training module modals to the ChA training tab in onboarding

viviancan:
Refs #4792 

This PR adds the training module links & modals to the training & checkpoint tab for ChA onboarding. 

It would be nice if the training modules were more like a "checklist" and the quiz would only be available once all trainings were completed. Not a must-have but a nice to have! 

Screenshot 2024-06-20 at 1 58 02 PM

Screenshot 2024-06-20 at 1 58 08 PM

### DIFF
--- a/app/views/chapter_ambassador/trainings/_training_module.html.erb
+++ b/app/views/chapter_ambassador/trainings/_training_module.html.erb
@@ -1,0 +1,24 @@
+<li>
+  <%= link_to "Training Module #{module_number}: #{module_title}", '#',
+              data: {
+                opens_modal: "training-modal-#{module_number}",
+              }
+  %>
+</li>
+
+<%= content_tag :div,
+                class: "modal",
+                data: {
+                  heading: "Training Module #{module_number}: #{module_title}",
+                  module_number: module_number,
+                },
+                id: "training-modal-#{module_number}" do %>
+
+  <div class="modal-content">
+    <iframe class="w-full" height="315" src="<%= ENV.fetch("CHAPTER_AMBASSADOR_TRAINING_SLIDES_MODULE_#{module_number}_URL", "")%>"
+            title="YouTube video player" frameborder="0"
+            allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share"
+            referrerpolicy="strict-origin-when-cross-origin" allowfullscreen>
+    </iframe>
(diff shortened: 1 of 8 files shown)


app/views/chapter_ambassador/trainings/_training_module.html.erb
```diff
@@ -0,0 +1,24 @@
...
+<%= content_tag :div,
+                class: "modal",
+                data: {
+                  heading: "Training Module #{module_number}: #{module_title}",
+                  module_number: module_number,
+                },
+                id: "training-modal-#{module_number}" do %>
+
+  <div class="modal-content">
+    <iframe class="w-full" height="315" src="<%= external_chapter_ambassador_training_module_link(module_number) %>"
```
viviancan, line 18: I used the youtube iframe here, but this will need to be updated once final training content is completed. I think the trainings will be google slides

viviancan:
> Looks like this branch needs to be rebased against the preview branch and then I can review it. 👍

@shaun-technovation This is updated!

viviancan:
> Some food for thought...
> 
> Naming these Heroku ENVs as:
> 
> `CHAPTER_AMBASSADOR_TRAINING_SLIDES_MODULE_1_URL` `CHAPTER_AMBASSADOR_TRAINING_SLIDES_MODULE_2_URL` etc.
> 
> Vs. something like:
> 
> `CHAPTER_AMBASSADOR_TRAINING_PROGRAM_OVERVIEW_URL` `CHAPTER_AMBASSADOR_TRAINING_CURRICULUM_URL` etc.
> 
> Naming them with numbers, for me it suggests that the URLs can be reordered (I want to change URL three to be number one now), but this wouldn't work as intended.
> 
> Naming them for what the are, for me it suggests I can update the URL for a certain training module, which would work as intended. Clearer and less confusion, IMHO.
> 
> Also, including `SLIDES` seems unnecessary, it's just a link to some material. Although I could see it being useful if one training module has both slides and a video and we need to distinguish between them.
> 
> Anywho, FWIW. 😄

Great point! 👍  I will remove "slides" from the env variable name. I can see what you mean about how the numbers suggest that they can be reordered, but the numbers also indicate that the modules go in a specific order (which right now, these modules are sequential and go in this order). I will wait to update the names until I have a final answer from PT. It is also possible that the title names will change before this goes live!

PR #4806, opened 2024-06-21
Security: Update rexml gem

shaun-technovation:
This will update rexml which should address this security vulnerability:

- https://github.com/Iridescent-CM/technovation-app/security/dependabot/194


viviancan (approved):
LGTM 💯

PR #4815, opened 2024-06-25
Remove unused ChA introductions code

shaun-technovation:
I removed a lot of the unused chapter ambassador introduction pieces (the route, controller, and some view front-end pieces). The database column can't be removed b/c it will be used when we create new chapters for existing chapter ambassadors.


viviancan (approved):
LGTM! I logged in as an onboarded ChA and was able to access the admin activity and various subtabs!

PR #4819, opened 2024-06-25
Chore: Add bullet gem

shaun-technovation:
This will add the bullet gem as a dev dependency and a basic configuration for it.

There's a lot of config options, the config in this PR one was right from the README. Maybe the same as you have setup locally? If not, we can adjust accordingly; this should give us the basics at least.


viviancan (approved):
Yes I had a similar configuration locally! LGTM!

Issue #4818, opened 2024-06-25
Add bullet gem to dev environment


shaun-technovation:
Add the bullet gem to our dev environment. The bullet gem will help us catch n+1 query violations.

Issue #4810, opened 2024-06-24
Review/update chapters grid for n+1 query


viviancan:

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/4810?planning-poker) with Zenhub @shaun-technovation @viviancan

dboyer:
I used a few filters, added additional columns, and exported data on the chapters admin page. Everything worked as expected!

Issue #4774, opened 2024-06-06
Add filter and additional column for organization type


dboyer:
In the program info section, orgs answer the multi-select question:

Is your organization a:

- School 
- NGO or Nonprofit Organization
- University
- Corporation
- Government agency
- Other institution

This info needs to be a filter and additional column on the chapter admin page (/admin/chapters). Two things:

- [x] Add an additional column of "organization type" 
- [x] Add a filter for "organization type" that works similarly to the mentor and judge type filters where the admin can select multiple choices.
<img width="580" alt="Screen Shot 2024-06-06 at 1 58 06 PM" src="https://github.com/Iridescent-CM/technovation-app/assets/516764/375dba47-2860-4dd2-871b-d4877f5b78d1">

viviancan:
This is on preview

dboyer:
Confirming that on preview I can:

- Add organization type as an additional column and if the org has multiple types, both types appear in the column
- I can use the filter to filter by org type
- Org type appears in the exported data

Issue #4773, opened 2024-06-06
Add option to export program info


dboyer:
Admins can view program info (#4724) but there is no way to export it. PT requested an option to export all program info for all chapters as a CSV. We would want to include:

- chapter name
- org name
- org location
- all the fields in the program info section

Perhaps we add that export option as a link on the chapters admin page (/admin/chapters).

viviancan:
This is on preview

dboyer:
Confirming that I can add the program info questions as additional columns and then export the data.

PR #4820, opened 2024-06-25
Add all chapter program information to chapters grid 

viviancan:
Refs #4774 & #4773 

This PR adds the following:
- all chapter program information responses as columns to the chapters grid 
- It also adds the organization type as a multi-select filter
- I ordered the columns to match the order of the questions/form 
- A small addition to the conditional logic that allows for saved searches. This was an issue where searches could not be saved for the Chapter Ambassadors grid and chapters grid

I am going to update/check this grid for n+1 queries as part of #4810 & once the bullet gem is added to this project structure (#4818)

### DIFF
--- a/app/controllers/concerns/saved_search_controller.rb
+++ b/app/controllers/concerns/saved_search_controller.rb
@@ -23,6 +23,10 @@ module SavedSearchController
       "#{current_scope}_score_exports_path"
     when "scored_submissions_grid"
       "#{current_scope}_scores_path"
+    when "chapters_grid"
+      "#{current_scope}_chapters_path"
+    when "chapter_ambassadors_grid"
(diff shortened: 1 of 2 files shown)


app/controllers/concerns/saved_search_controller.rb
```diff
@@ -23,6 +23,10 @@ def show
       "#{current_scope}_score_exports_path"
     when "scored_submissions_grid"
       "#{current_scope}_scores_path"
+    when "chapters_grid"
+      "#{current_scope}_chapters_path"
+    when "chapter_ambassadors_grid"
+      "#{current_scope}_chapter_ambassadors_path"
```
viviancan, line 29: This was needed in order to use the saved searches on the ChAs & Chapters grid

PR #4826, opened 2024-06-26
Rename regional links to chapter links

shaun-technovation:
This will rename regional links to chapter links.


viviancan (approved):
Tried a few things locally with the links and everything worked as expected! Nice clean up here!

PR #4832, opened 2024-06-27
Add onboarded filter and column to chapter's datagrid

shaun-technovation:
This will:
- Add an onboarded filter and column for the chapter's datagrid
- Group the filter and columns on this datagrid
- Address n+1 issues for this datagrid


viviancan (approved):
Did not try locally but LGTM! Nice find with the eager/pre loading!

PR #4833, opened 2024-06-28
Update webpack-dev-server

shaun-technovation:
This is the first step into updating things related to webpack. This will update webpack-dev-server. I ran `yarn upgrade webpack-dev-server` to do the upgrade, then I ran the command `bin/webpack-dev-server` and everything seems to be running as expected for me locally.


viviancan (approved):
I followed the same steps and everything seemed to work as expected!

PR #4835, opened 2024-07-01
Update rake task that converts ChAs to chapters

shaun-technovation:
This will update the rake task that is used to convert chapter ambassadors to chapters so that it accepts an `account_id` instead of a `chapter_ambassaador_id`.


viviancan (approved):
Did not try locally but looks pretty straightforward 💯 🚀

PR #4841, opened 2024-07-02
Improvements for participants datagrid

shaun-technovation:
This will:
- Address another n+1 query on the participants datagrid
- Add an index on the memberships table on the `member_type` column

We have two queries (in `Account.rb`) that search the `memberships` table, one query for mentors (where `member_type` is for a mentor) and one for students (where `member_type` is for a student). Both of these queries were using a table scan, I added an index on `member_type` to help with this.


viviancan (approved):
👍

PR #4847, opened 2024-07-05
Separate matched with a team filters

shaun-technovation:
This will add two new "matched with a team" filters, one for mentors matched with a team, and one for students matched with a team.

To test this locally, I pulled down production data, using the existing "matched with a team" filter and selecting mentors that are fully onboarded came back with 4,011 results. I tried the same thing using the new "Mentors matched with a team" and it returned the same number of results, so I think we're good.


viviancan (approved):
💯

viviancan:
Nice updates! Excited to see this on Prod!

PR #4848, opened 2024-07-05
HOTFIX for "Separate matched with a team filters"

shaun-technovation:
This has the same changes as in #4847; this one is just showing the changes that will be going to Production (instead of QA).


viviancan (approved):
👍 It's great to try out the hotfix solution here!

PR #4849, opened 2024-07-09
Update Salesforce integration

shaun-technovation:
This will:
- Update our Salesforce integration so that we update contacts in Salesforce by account id (the account id from our platform, referenced as Platform Id in Salesfore)
- Use an `upsert` to update or insert contacts in Salesforce
- Remove the Salesforce Id that we were saving in our database
- Remove the calls to Mailchimp (the integration and functionality is still there, just the calls were removed)
- Send city, state, country, DOB, parent name and email to Salesforce

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -1083,6 +1083,8 @@ class Account < ActiveRecord::Base
       :last_name,
       :email,
       :date_of_birth,
+      :city,
+      :state_province,
       :country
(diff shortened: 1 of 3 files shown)


viviancan (approved):
Wooohooo! Great work on this! It looks good to me. I think it will be great once everything is hooked up and we get feedback.

PR #4857, opened 2024-07-11
Remove submission type question

shaun-technovation:
Just a basic update to remove the submission type question from the submission process and renaming that section to Coding Language.


viviancan (approved):
I was able to select the language but not AI or mobile app (since it was removed). I think everything is working as expected!

Issue #4800, opened 2024-06-20
Add fields for entering Scratch project for the submission


dboyer:
Depends on #4799 

If a beginner division team selects Scratch for their coding language, they should see this text with two upload options:

1. A field saying "What is the URL for your Scratch project page?

Helper text: Enter the Scratch project page URL. To create a project page, click the orange "Share" button in your Scratch project. Then when the project page appears, click the Copy Link button. This is the URL you will enter here. 

2. An upload option saying: If you used a Scratch environment other than the [scratch.mit.edu](http://scratch.mit.edu/) website, you will save your project to your computer, and then upload that file in the Technical Additions section of this form. The file will have a .sb3 file extension name.

As with Thunkable, if the student uploads a file, that file should automatically display in the Technical Additions section.

- [x] Scrach option in dropdown on coding language section for beginner 
- [x] Migration/field to save `scratch_project_url`
- [x] Scratch URL validation
- [x] Update the technical additions to show the scratch link automatcally
- [x] Update views for admin and student for scratch platform 
- [x] Update admin edit form
- [x] General cleanup (remove submission type, reset other fields)

viviancan:
This is on QA. There seems to be an issue with the "technical additions" overview section where the Scratch link is not displaying automatically. I am investigating!

viviancan:
I will add the fix for this ^^ in the cleanup ticket #4871

dboyer:
Tested on QA and confirmed:

- I see the Scratch link when I log in as a beginner division team
- I entered a junk URL for the Scratch link and it gave me an error message
- I entered a Scratch URL and it saved correctly
<img width="689" alt="Screen Shot 2024-07-25 at 9 32 11 PM" src="https://github.com/user-attachments/assets/fbdaf1f1-8db4-4688-a48b-dadd14b3403c">

@viviancan The check marks for both the coding language and technical additions are not getting a green check mark but I assume that is something being fixed in the other issue.

dboyer:
Confirming the check marks are now working as expected based on the clean up in #4871

Issue #4799, opened 2024-06-20
Make coding language choices dependent on division


dboyer:
If possible, the curriculum team would like the answers to the coding language question to depend on the division of the team. The current choices are:

- App Inventor
- Thunkable
- Other

If the team is in the beginner division, they should also have the option to select Scratch.

viviancan:
This is on QA. Only beginner teams should see the Scratch option

dboyer:
Tested on QA and confirmed that I see Scratch when I'm a beginner but not when logging in as other divisions.

PR #4872, opened 2024-07-23
Adds Scratch as a development platform option + other updates related to this addition 

viviancan:
Refs #4799 & #4800 

This PR adds `Scratch` as a development platform option to the student and admin experience. It also adds the following:

- Scratch only available as an option for beginner division teams
- Migration/field to save scratch_project_url
- Scratch URL validator 
- Update the technical additions to show the scratch link automatcally
- Update views for admin and student for scratch platform
- Update admin edit submission form
- General cleanup (remove submission type, reset other fields)


app/models/team_submission.rb
```diff
@@ -52,6 +53,7 @@ class TeamSubmission < ActiveRecord::Base
   before_validation :reset_development_platform_fields_for_app_inventor
   before_validation :reset_development_platform_fields_for_thunkable
   before_validation :reset_development_platform_fields_for_other_platforms
+  before_validation :reset_development_platform_fields_for_scratch
```
viviancan, line 56: I followed the current pattern, but I am wondering if there is a better way to do this. From my understanding, this logic will "clear out" the other platform fields

app/models/team_submission.rb
```diff
@@ -77,8 +79,8 @@ class TeamSubmission < ActiveRecord::Base
       Rails.logger.warn("Published submission id=#{id} is missing required fields.")
     end
 
-    if source_code_external_url.blank?
-      columns[:source_code_external_url] = copy_possible_thunkable_url
+    if developed_on?("Thunkable") || developed_on?("Scratch")
```
viviancan, line 82: This logic will update the `source_code_external_url` field with either the scratch or thunkable project link. There are a few paths for development platforms
- Thunkable - project link only
- Other, App Inventor - details + technical uploads
- Scratch - Can have a project link or technical uploads (but not both)

app/models/team_submission.rb
```diff
@@ -568,15 +572,18 @@ def senior_division?
   end
 
   def development_platform_text
```
viviancan, line 574: Updated the platform text to remove submission type references

app/models/team_submission.rb
```diff
@@ -568,15 +572,18 @@ def senior_division?
...
-      "#{submission_type} - #{development_platform} - #{development_platform_other}"
+    if development_platform == "Other"
+      "#{development_platform} - #{development_platform_other}"
     else
-      submission_type
+      development_platform
     end
   end
 
+  def self.development_platform_options(team_submission)
```
viviancan, line 582: created this method so that scratch is removed from the options displayed for beginner team submissions

Issue #4170, opened 2023-08-22
As a student, you cannot change between mobile app - app inventor and mobile app- thunkable for submission type 


viviancan:
When you select either app inventor or thunkable as your submission type and then go back to update to the other submission type (thunkable to app inventor or app inventor to thunkable) your project is not updated. The only workaround right now is to select "other" or switch to an AI project, save, and then you can select the thunkable/app inventor. Verified locally and on QA

dboyer:
I'm seeing the same issue on QA

viviancan:
The switch should work now on QA, however some of the other platform switches are still kind of weird. I think this should be addressed and is logged in #4881

dboyer:
Tested on QA and I'm seeing the same problem still:

- Set my coding type as App Inventor and upload a zip file. Both coding language and technical additions are green check marks
- Went back to coding type and changed it to Thunkable and added my project details page
- Clicked Save but the changes wouldn't save

@viviancan Screen capture of the [issue here](https://drive.google.com/file/d/1_VQ5mEMjqN6V4AgRRip_ptPaS4HHkvBb/view?usp=drivesdk).

viviancan:
Pulling this back to in progress

viviancan:
@dboyer This is on QA. I did something weird with the original PR and the actual code bug fix was not on QA 🤦‍♀️ It should work as expected for app inventor/thunkable

dboyer:
Tested on QA and confirmed:

- Had the coding language set to Thunkable and was able to switch to App Inventor and then upload a file
- Had the coding language set to App Inventor and was able to switch to Thunkable. The uploaded file disappeared and I had to enter a Thunkable link
- The green check marks in coding language and technical additions worked as expected

Issue #2826, opened 2020-12-23
Bug: when switching dev platform to thunkable, a previous uploaded source code is saved [3,5,0.6]


rsgonzal:
If I've previously uploaded a file under a different development platform and then decide to go with a thunkable project, my old source code is still saved on my submission

![Screen Shot 2020-12-23 at 12 21 02 PM](https://user-images.githubusercontent.com/22801026/103034524-ee555200-4519-11eb-9c8c-a24b382984d7.png)
![Screen Shot 2020-12-23 at 12 25 01 PM](https://user-images.githubusercontent.com/22801026/103034536-f2816f80-4519-11eb-9572-5e2e4f441762.png)

Probably an unlikely scenario, but worth checking out. How does this look on judging?

hellafitz:
next step: @rsgonzal to check out how this looks on Judging

rsgonzal:
A judge can only see thunkable project, so this doesn't look like a big issue for now
![Screen Shot 2021-03-17 at 12 22 07 AM](https://user-images.githubusercontent.com/22801026/111430026-066ee480-86b7-11eb-8f9a-7898eb8712dc.png)

viviancan:
So weird to see the old styling 🤯

dboyer:
Tested on QA:

- Selected "other" as the coding language and uploaded a zip file - saw two green check marks for coding language and technical additions
- Switched the coding language to Thunkable and added a link to the Thunkable details page. The link saved successfully in the technical additions section.

PR #4882, opened 2024-07-25
Updated reset logic to address bugs when switching development platform

viviancan:
Refs #2826 & #4170 

This update addresses 2 bugs

The 1st is very old (#2826). If you previously uploaded technical additions and then switched to a thunkable project, the upload was still part of the submission and would display in the technical additions detailed view

CleanShot 2024-07-25 at 13 17 55@2x

the 2nd bug (#4170) - You could not change between app inventor and thunkable 

This PR addresses these 2 bugs, however I think there is a better way to address resetting the platform fields when switching between development platform types. I added #4881 to capture this.

### DIFF
--- a/app/models/team_submission.rb
+++ b/app/models/team_submission.rb
@@ -736,13 +736,19 @@ class TeamSubmission < ActiveRecord::Base
   def reset_development_platform_fields_for_app_inventor
     if development_platform == "App Inventor"
       self.development_platform_other = nil
+      self.thunkable_account_email = nil
+      self.thunkable_project_url = nil
+      self.scratch_project_url = nil
     end
   end
 
   def reset_development_platform_fields_for_thunkable
     if development_platform == "Thunkable"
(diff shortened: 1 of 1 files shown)


app/models/team_submission.rb
```diff
@@ -736,13 +736,19 @@ def reset_development_platform_fields_for_ai_projects
...
       self.development_platform_other = nil
+      self.thunkable_account_email = nil
+      self.thunkable_project_url = nil
+      self.scratch_project_url = nil
     end
   end
 
   def reset_development_platform_fields_for_thunkable
     if development_platform == "Thunkable"
+      self.remove_source_code! if source_code.present?
```
viviancan, line 747: I found this approach in the carrierwave docs [here](https://github.com/carrierwaveuploader/carrierwave?tab=readme-ov-file#removing-uploaded-files) 

Setting `source_code` to nil would not work.

PR #4888, opened 2024-07-31
Update student and mentor program info in Salesforce

shaun-technovation:
The main feature here is that during a season, when certain student or mentor details are updated, those values will get updated in Salesforce: 

- Mentor Types
- Mentor on a team vs. not on team
- Project Link
- Team Name
- Pitch Video
- Submitted project


viviancan (approved):
Great work on this! 💯

PR #4895, opened 2024-08-02
Setup back-end for 2025 season

shaun-technovation:
This will add the following for the 2025 season:
- judging questions
- additional questions

And it will update some date-related CircleCI ENVs for the 2025 season.


circle.yml
```diff
@@ -31,54 +31,54 @@ jobs:
...
-          DATES_QUARTERFINALS_BEGINS_YEAR: 2022
+          DATES_QUARTERFINALS_BEGINS_YEAR: 2025
           DATES_QUARTERFINALS_ENDS_DAY: 20
           DATES_QUARTERFINALS_ENDS_MONTH: 5
-          DATES_QUARTERFINALS_ENDS_YEAR: 2022
+          DATES_QUARTERFINALS_ENDS_YEAR: 2025
           DATES_REGIONAL_PITCH_EVENTS_BEGINS_DAY: 1
           DATES_REGIONAL_PITCH_EVENTS_BEGINS_MONTH: 4
-          DATES_REGIONAL_PITCH_EVENTS_BEGINS_YEAR: 2022
+          DATES_REGIONAL_PITCH_EVENTS_BEGINS_YEAR: 2024
```
viviancan, line 57: I think RPEs only run when QF are open

circle.yml
```diff
@@ -31,54 +31,54 @@ jobs:
...
           DATES_REGIONAL_PITCH_EVENTS_ENDS_DAY: 1
           DATES_REGIONAL_PITCH_EVENTS_ENDS_MONTH: 5
-          DATES_REGIONAL_PITCH_EVENTS_ENDS_YEAR: 2022
+          DATES_REGIONAL_PITCH_EVENTS_ENDS_YEAR: 2024
           DATES_REGISTRATION_OPENS_DAY: 1
           DATES_REGISTRATION_OPENS_MONTH: 10
-          DATES_REGISTRATION_OPENS_YEAR: 2022
+          DATES_REGISTRATION_OPENS_YEAR: 2025
```
viviancan, line 63: Should this be 2024?

viviancan (approved):
I just left 2 comments about the year value, but besides that LGTM! Thx for taking care of this 💯

circle.yml
viviancan: ohh ok! I think it's ok then, especially since we are in that weird time window 🙂

viviancan: Sounds good to me!

Issue #4878, opened 2024-07-24
Add phone number as an optional field on judge profile


dboyer:
When judges register, there should be an optional field for them to provide their phone number - similar to what we do for ChAs. This phone number field should appear in the personal data on their profile.

Let's include the helper text:

Your phone number will be shared with the Technovation Ambassador for your region and may be used to contact you regarding volunteer judging opportunities. Providing your phone number is optional.

Issue #4877, opened 2024-07-24
Add phone number as an optional field on mentor profile


dboyer:
When mentors register, there should be an optional field for them to provide their phone number - similar to what we do for ChAs. 

This will be stored on mentor profile, not the account profile.

Let's include the helper text:

Your phone number will be shared with the Technovation Ambassador for your region and may be used to contact you regarding volunteer opportunities. Providing your phone number is optional.

<img width="726" alt="Screen Shot 2024-07-24 at 11 24 36 AM" src="https://github.com/user-attachments/assets/a170b3f5-484a-4cce-be8d-539190bc4e33">

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/4877?planning-poker) with Zenhub @shaun-technovation @viviancan

dboyer:
Oh good question. Yes, if someone signed up as a judge but now also has a mentor account, the ChA should be able to see their phone number. I'll add it to the dev priorities discussion, but I feel pretty confident that we will want judge phone numbers to display for ChAs, even when they are looking at the judge's mentor profile or mentor record on the participant admin page.

dboyer:
Testing on QA and I think there is some validation that is making the phone field throw an error even if it is blank.
<img width="809" alt="Screen Shot 2024-08-08 at 11 41 57 AM" src="https://github.com/user-attachments/assets/6d26a661-70fa-463d-b9dc-12f6fbac44e7">

I was registering as a ChA and left the phone number field blank. When I got to the last page to submit my details, the button just didn't work. I went back and saw the error message of "phone number is invalid" on the phone number field but hadn't added any phone number. If I add a number, the error message goes away.

dboyer:
Confirmed on QA that I can now create a mentor or ChA account without adding a phone number. I also confirmed I could create a mentor and ChA account if I added a phone number.

PR #4905, opened 2024-08-06
Phone number registration field for mentors and judges

shaun-technovation:
This will add an optional phone number field to the registration flow for mentors and judges. It was already there for chapter ambassadors, but it was being saved to their chapter ambassador profile, now all phone numbers will get saved at the account level.

There are a bunch of auto-formatting fixes in this PR too.

Addresses: #4877, #4878


viviancan (approved):
Nice work on this 💯 I was able to register as a cha, mentor, & judge and enter my phone number during registration. I also saw the phone number in the admin portal under the detailed participant information. 

Since this field is open to anyone who registers (not just cha invites), should we add validation to the field? I was able to add random characters and also really long strings.

viviancan:
Also, the phone number displays twice for ChAs in the admin detailed view
Screenshot 2024-08-06 at 9 36 58 PM

Issue #4851, opened 2024-07-10
List onboarding tasks needing completion in error message


dboyer:
When a ChA tries to access the chapter admin activity without completing all the onboarding tasks, the error message should list which tasks remain. We can use the titles of the tasks in the left sidebar. For this version, we don't need to update which specific lines on which forms are missing.

<img width="1274" alt="Screen Shot 2024-07-10 at 11 08 20 AM" src="https://github.com/Iridescent-CM/technovation-app/assets/516764/a63836e7-2c03-4d57-b5c2-b4b2f7cae530">

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/4851?planning-poker) with Zenhub @shaun-technovation @viviancan

viviancan:
<img width="2000" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeTgvQmc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--870352cd8f4013c684dbc29348875333fad81d51/Screenshot%202024-08-06%20at%208.40.39%E2%80%AFPM.png" alt="Screenshot 2024-08-06 at 8.40.39 PM.png" />

viviancan:
This is on QA

dboyer:
Tested on QA and confirmed that I see the error message and that it updates based on which tasks that I've completed. The message goes away after I've completed all tasks.

PR #4906, opened 2024-08-07
Added welcome message that includes incomplete cha and chapter onboarding tasks to cha dashboard and chapter tab

viviancan:
Refs #4851 

This PR adds the following:
- `OnboardingTasksConcern`
- Logic to determine which chapter ambassador and chapter onboarding tasks are incomplete
- Welcome partial (similar to student experience) that displays incomplete tasks

Screenshot 2024-08-06 at 8 40 39 PM

### DIFF
--- a/app/models/chapter.rb
+++ b/app/models/chapter.rb
@@ -2,7 +2,6 @@ class Chapter < ActiveRecord::Base
   include ActiveGeocoded
   include OnboardingTasksConcern
 
-
   belongs_to :primary_contact, class_name: "ChapterAmbassadorProfile", foreign_key: "primary_contact_id", optional: true
 
   has_one :legal_contact, dependent: :destroy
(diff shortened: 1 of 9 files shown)


app/models/chapter.rb
```diff
@@ -72,4 +73,14 @@ def location_complete?
   def program_info_complete?
     chapter_program_information&.complete?
   end
+
+  def required_onboarding_tasks
```
viviancan, line 77: I'm wondering if something about the `status` in the name would be more appropriate/descriptive

viviancan:
> We already have an an onboarding steps flow (see students and mentors, and onboarding_steps_required). Not sure why we need a new way of doing this, but If it works, I'm good with it. 👍 :shipit:

Thanks for sharing, I did not know that existed. I will take a look at that.

Issue #4917, opened 2024-08-08
Docusign wrap up


shaun-technovation:
- [x] Disable send buttons after sending to help prevent duplicate documents from being sent
- [x] Prevent more than one affiliation agreement from being sent
- [x] Make the title field required
- [x] Update the organization name and location fields
- [x] "Unlink" location and title fields
- [x] Make location required
- [x] Add organization name (to the footer) - **A new issue was created for this #4948**

dboyer:
One note on this. I found that when moving between send and void, the text on the button doesn't change until after a page refresh. For example as shown [in this recording](https://drive.google.com/file/d/1iLRKM6ZnaVPZNaLcHz9SE98s8evT4yyW/view?usp=drivesdk):

1. I clicked send agreement
2. The page showed that sending the agreement was in the queue and the send button remained the same
3. I refreshed the page and now the button said void
4. I clicked the button to void the agreement and got the header message that voiding would happen
5. The void button remained and did not change to a send button until I refreshed the page

Issue #4876, opened 2024-07-24
Set up official 2025 agreements on DocuSign


dboyer:
PT will set up the two templates in DocuSign.

- ChA legal agreement
- Chapter Affiliation Agreement

This task covers linking to those final agreements on DocuSign. @shaun-technovation Please feel free to add anything else here that you would like. I know we also have #4701 to cover the go live on production.

One question: Is it possible to automatically apply a signature from someone at Technovation? We don't want to manually countersign each doc. If an auto-applied signature isn't possible, we'll just go with having the ChA sign it and not someone from Technovation.

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/4876?planning-poker) with Zenhub @shaun-technovation @viviancan

dboyer:
@shaun-technovation We would need to set up a new signature spot in the doc for us to pre-fill after the doc is signed. I think we can launch without that though because as you said, an auto-signature might not be ideal.

Tested the new docs on QA and I think the fields in the header of the Chapter Affiliation Agreement are off. It should be:

"an understanding made between Technovation, a US 501c3 nonprofit, and ORGANIZATION with offices in LOCATION (herein referred to as Partner)"

- It looks like spot where the organization name is supposed to be pre-filled is currently pre-filled with the legal contact's name. The spot where the location should go is being pre-filled with the organization's name. The location spot can just be an empty spot where the legal contact fills in the legal address for the org. 

- Can we make the title field required on the Chapter Affiliation Agreement?

No concerns with the ChA agreement!

dboyer:
Tested on QA and confirmed:

- Org field is now pre-filled with the org name
- I see a field to add a location
- The title field is required

@shaun-technovation One issue - it looks like the field for the location in the header is linked to the title field in the footer. Whatever I put in one field is added to the other field too. I can't have different text in each field.

dboyer:
One more note: looks like the bottom fields aren't fillable on the production version
<img width="776" alt="Screen Shot 2024-08-09 at 1 57 28 PM" src="https://github.com/user-attachments/assets/186aa6ee-fecb-4e86-8a75-ce53bf6fbb3a">

dboyer:
Confirming that the two fields are not linked on the production documents.

PR #4928, opened 2024-08-09
Update affiliation agreement Docusign fields

shaun-technovation:
This will update the location and title fields for the affiliation agreement.

Location:

- Making it required
- Adjusting the width
- Adding a label and id

Title:

- Adding a label and id

The ids are needed so they're treated as separate fields.

Related to: #4917, #4876


viviancan (approved):
👍

Issue #4934, opened 2024-08-12
Create rake task that will set a ChA as the legal contact for an org


dboyer:
There are several existing ChAs that will serve as the legal contact for their org. If possible, could we create a rake task that would take a participant ID and an org name (or ID) and set that participant as the org's legal contact? It would not need to automatically send the legal agreement. Program team can take care of sending it.

dboyer:
Moving to done as all chapters are migrated.

PR #4943, opened 2024-08-13
Rake task to make ChAs legal contacts

shaun-technovation:
This rake task will make a chapter ambassador a legal contacts for their chapter.

This rake task needs a list of account ids, for each account id, if they are a chapter ambassador, they belong to a chapter, and a legal contact isn't already setup, it these conditions are met it will make the chapter ambassador the legal contact for the chapter.

The rake task is run with this command:

`bundle exec rake set_chapter_ambassadors_as_legal_contacts[account_ids]`

Addresses: #4934


viviancan (approved):
Did not try locally, but looks straightforward! Nice work on this 💯

Issue #4913, opened 2024-08-07
Add Salesforce program participant record for profile conversions/additions


shaun-technovation:
When profiles get converted or added during the season, we need to create a new program participant record in Salesforce for the new profile type.

- Student to mentors
- Judges to mentors
- Mentors to judges
- Mentors to ChAs

And any other conversions that happen during a season.

For context, each profile type has their own program participant record in Salesforce. When someone gets registered to the season, all known profile types at that time will get their program participant created in Salesforce. But any profile types added during the season will need to have their program participant record created in Salesforce.

Issue #4904, opened 2024-08-06
Salesforce wrap up work


shaun-technovation:
From the Salesforce email sent on 7/24:

- Map email address to alternate email in Salesforce (`npe01__AlternateEmail__c`)
- Only send date of birth for students

From the Salesforce email sent on 8/6:

- Send "Club Ambassador" as "Mentor Role"
- Remove "Club Ambassador" from "Mentor Types"

I couldn't find this documented, but last I heard we wanted this:
- Convert periods to LNU in the conversion rake task and service (that updates their info)

dboyer:
@shaun-technovation I see you saw the update that we're not moving to LNU for the period. In our Salesforce [meeting yesterday](https://docs.google.com/document/d/1x0pUaf6BrhiGEXKY1WuO9fa-VVCDE_2MNozFVsawx50/edit#heading=h.4dewty1ilrg1), Sasha also said that we can keep club ambassador as a mentor type. Apologies if you had already changed it to a mentor role. I didn't think you'd get to it before I updated the issue this morning.

PR #4944, opened 2024-08-13
Only add Salesforce program participant record if upsert was successful

shaun-technovation:
I mentioned on standup that the Salesforce errors are coming in pairs, the first error message is when the upsert fails:
`DUPLICATE_VALUE: duplicate value found: Platform_Participant_ID__c`

Which leads to this error (b/c we don't have a contact id which gets returned from the upsert):

`REQUIRED_FIELD_MISSING: Required fields are missing: [Contact__c]`

This PR addresses the second error message. It's basically a band-aid/hack fix where if the upsert to Salesforce fails, we won't try to add a program participant record because we won't have a contact id for them (which is needed to create the program participant record). A real fix is figuring out why the upsert can fail with the error message:

`DUPLICATE_VALUE: duplicate value found: Platform_Participant_ID__c`

Upsert should update or insert by the `Platform_Participant_Id__c`, there shouldn't be any duplicate ids in Salesforce because it's marked as distinct. I don't have a fix for this at the moment, so the "fix" here is to stop trying to create the program participant record if there isn't a contact id, which should stop these errors from coming in:

`REQUIRED_FIELD_MISSING: Required fields are missing: [Contact__c]`

Related to: #4904 #4913

### DIFF
--- a/app/services/salesforce/api_client.rb
+++ b/app/services/salesforce/api_client.rb
@@ -45,16 +45,18 @@ module Salesforce
         salesforce_contact_id = upsert_contact
       end
 
-      handle_request "Setting up account (#{account.id}) for current season" do
-        client.insert!(
-          "Program_Participant__c",
-          {
-            Contact__c: salesforce_contact_id,
-            Platform_Participant_Id__c: account.id,
-            Year__c: Season.current.year,
-            Type__c: profile_type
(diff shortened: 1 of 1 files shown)


viviancan (approved):

viviancan:
Lgtm! Let me know if there's anything I can do to support.

Issue #4909, opened 2024-08-07
Post-launch feedback


dboyer:
Running list of issues reported after August 12 launch

- [x] If you login as a ChA that does not belong to a chapter, clicking Chapter Ambassador Training will cause an error
- [x] Default the ChA admin page (/admin/chapter_ambassadors) to show ChAs for the current season
- [x] Docusign webhook could not update document on our end

dboyer:
Confirming that I see the 2025 filter applied as a default to the ChA admin page.

One note is that it's impossible to remove the season and just view all ChAs. This is also true when searching for participants on other admin pages so may be something we want to address globally at another time.

dboyer:
Confirmed on production that I could remove the year filter and search across seasons.

PR #4945, opened 2024-08-14
Make current season the default season for chapter ambassador datagrid

shaun-technovation:
This will make the current season the default season for the chapter ambassador's datagrid.

Addresses: #4909


viviancan (approved):
👍

PR #4946, opened 2024-08-14
Prevent duplicate legal agreements from being sent

shaun-technovation:
This will add a check at the controller level to prevent a duplicate legal agreement from being sent.

I also added a little hack to hide the "Review and Sign" button, where we pass a `legal_agreement_sent` param to the view and if that is present, we will hide the button, this should help with people clicking the button again (because it won't be displayed).

Addresses: #4917, #4947


viviancan (approved):
Nice update on this! I think it will really help prevent multi-click!

PR #4955, opened 2024-08-15
Update season value fuctionality for ChA datagrid

shaun-technovation:
A pretty simple update here for the season field on the chapter ambassador's datagrid:

- If `params[:chapter_ambassadors_grid]` aren't present, i.e. coming from another page
   - Use the current season

- If `params[:chapter_ambassadors_grid]` are present, i.e. submitting a request to the chapter ambassador datagrid
   - Use the season value that was selected

Related to: #4909


viviancan (approved):
This is such a nice addition! Will be very beneficial for all of our datagrids 🚢

PR #4970, opened 2024-08-22
Salesforce updates and rake task

shaun-technovation:
This will add a rake task that will ensure contact and program participant info has been created and is up-to-date in Salesforce (making it so that it matches the info that is on our platform).

I had to update some of the existing functionality to do this, mainly adding upsert functionality for the program participant record:
- if a program participant record exits it will get updated (this is existing functionality)
- if a program participant record doesn't exits, one will get created for them (this is new functionality)

The rake task can be run like this:
`bundle exec rake update_salesforce_contacts[account ids]`

I tested this locally and worked out a few kinks, I plan to test it more when it gets to QA.


viviancan (approved):
Nice updates! LGTM 💯

Issue #4954, opened 2024-08-15
Support searching across seasons


dboyer:
Currently on most participant admin pages, the season search filter defaults to the present season. While you can select past seasons individually, there is no way to search across all seasons. This makes it challenging to find out if a person ever participated in Technovation at any point.

dboyer:
Work was done under #4909 to add this to the ChA page. Deb to test and list out specific pages that should have the same search functionality across seasons added.

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/4954?planning-poker) with Zenhub @shaun-technovation @viviancan

dboyer:
Tested on QA and confirmed that I could remove the season filter to search across all seasons on:

- the participants page
- the teams page
- the submissions page

On the participants page, I think there may be a way to use the seasons option filter which has options for "match any season" or "match all seasons" to search across seasons but I wasn't consistently getting the same results. I think being able to just remove the year from the search filter works better.

dboyer:
Confirmed I can remove the year filter on production and search across seasons.

PR #4971, opened 2024-08-22
Search across seasons functionality for participants, teams, and submissions datagrid

shaun-technovation:
This will update the participants, teams, and submissions datagrid so that no season can be selected so that we can search across all seasons.

Addresses: #4954


viviancan (approved):
Lgtm 💯 nice addition!

Issue #4719, opened 2024-05-10
Rename ChA and chapter legal documents


shaun-technovation:
Throughout the code base, legal documents are referred to as MOU, "memorandum of understanding", legal documents...  Make all of this consistent:

- For chapter ambassadors their legal document is referred to as a "legal agreement".
- For chapters their legal document is referred to as an "affiliation agreement".

Related to #4691

## DEPLOY NOTES:

Remove this Heroku setting on Production:
`DOCUSIGN_CHAPTER_AMBASSADOR_LEGAL_AGREEMENT_TEMPLATE_ID`

dboyer:
On QA

Chapter Affiliation Agreement is listed:

- On the chapter admin page on the button to send the doc to a legal contact or void it
- On the chapter admin that lists the agreement's status and the years for which it is valid
- On the chapter tab of the ChA onboarding dashboard

Chapter Volunteer Agreement is listed:

- On the onboarding tab of the ChA dashboard
- On the ChA admin page

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/4719?planning-poker) with Zenhub @shaun-technovation @viviancan

dboyer:
Confirmed on QA that I see the text as "chapter volunteer agreement" in the following places:

- In the left sidebar on the ChA onboarding page
- In the header of the ChA legal agreement tab
- In the list of onboarding tasks on the ChA participant admin page

dboyer:
Confirming I see the updated language to "chapter volunteer agreement" on production.

PR #4973, opened 2024-08-23
Rename ChA legal agreement -> chapter volunteer agreement

shaun-technovation:
All references to the chapter ambassador "legal agreement" will be referred to "chapter volunteer agreement".

Addresses: #4719


viviancan (approved):
👍

Issue #4957, opened 2024-08-15
BUG: ChA viewing their own profile in Chapter Admin Activity


shaun-technovation:
If a chapter ambassador views their own profile under Chapter Admin Activity it causes an error.

```
ActionView::Template::Error: Missing partial chapter_ambassador/participants/_chapter_ambassador_onboarding, chapter_ambassador/_chapter_ambassador_onboarding, application/_chapter_ambassador_onboarding with {:locale=>[:en], :formats=>[:html], :variants=>[], :handlers=>[:raw, :erb, :html, :builder, :ruby]}. Searched in:   * "/app/app/views"   * "/app/vendor/bundle/ruby/3.1.0/gems/datagrid-1.6.3/app/views"   * "/app/vendor/bundle/ruby/3.1.0/gems/actiontext-6.1.7.7/app/views"   * "/app/vendor/bundle/ruby/3.1.0/gems/actionmailbox-6.1.7.7/app/views"
```

dboyer:
Confirming on QA that when I access the admin activity as a ChA I could view the ChA's profile via the participants page and there was no error. The onboarding tasks and ability to update their status is gone.

I think ChAs should show in their own participant list since a chapter can have multiple ChAs. There may be a reason that a ChA would need to view the profiles of the other ChAs in their chapter.

Confirmed that as an admin, I still see the onboarding steps for a ChA but no longer see the ChA status box (pending, active, declined, spam). I'll double check with PT but I think that is fine. If we don't want a ChA to onboard, we can just unassign them from a chapter.

dboyer:
Logged in as a ChA on production and confirmed that I could view my own profile on the chapter admin activity participants tab.

PR #4978, opened 2024-08-23
Fix bug when a ChA views their own profile

shaun-technovation:
This will hide chapter and chapter ambassador onboarding steps for chapter ambassadors (this was causing an error); this info will only be available to admins.

I also removed the status field (active, pending, etc) for chapter ambassadors since this isn't being used now.

Fixes: #4957


viviancan (approved):
Did not try locally, but lgtm 💯 Nice fix on this!

Issue #4896, opened 2024-08-04
Add additional question on gadget selection


dboyer:
PT has requested that we add this additional question for the 2025 season. It is an optional question that is not viewed by the judges and can go anywhere in the additional questions section. It should be multi-select.

Which of the following additional gadgets are used in your project? Select all that apply.

- Micro:bit
- Arduino
- Raspberry Pi
- Other (Please specify)
- None

viviancan:
This is on QA

dboyer:
Tested on QA and confirmed:

- I was prompted to select yes/no if I had used a gadget
- If I selected no, I didn't receive any other option and could just save
- If I selected yes, I could select one or more options
- If I selected other, I received a field to enter the name of the other gadget I used

@viviancan It looks like not all the selected options are appearing on the submission overview page. I selected "Arduino" and "Other - Marbles" on this submission (/student/published_team_submissions/oak-trees-by-birnam-wood) but only Yes - Marbles is showing on the submission overview, both on the page to review before submitting and the page after submitting, and on the admin page for the submission. 

<img width="560" alt="Screen Shot 2024-09-02 at 11 22 16 PM" src="https://github.com/user-attachments/assets/cc5768e6-a564-41e5-9f1b-426042f3f9a8">

<img width="513" alt="Screen Shot 2024-09-02 at 11 25 14 PM" src="https://github.com/user-attachments/assets/3e3400c2-b6e9-4673-8c1f-603a72a65db2">


<img width="572" alt="Screen Shot 2024-09-02 at 11 22 42 PM" src="https://github.com/user-attachments/assets/87940456-132d-48fe-84f2-f875fe20a2df">

<img width="567" alt="Screen Shot 2024-09-02 at 11 23 20 PM" src="https://github.com/user-attachments/assets/10079cff-7ac3-4c63-839d-fbafdf641731">

viviancan:
Moving this to in progress while I address this ^^

dboyer:
Confirmed on QA that all selections now appear on the summary page before submitting, the page confirming submission, and the admin page. 

[...]

PR #4979, opened 2024-08-26
Backend work for uses gadgets additional question

viviancan:
Refs #4896 

This is the 1st part for 4896 which adds the backend work for the `uses gadgets` additional question
- Migrations
- Model updates
- Test updates/additions

### DIFF
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -24,6 +24,20 @@ COMMENT ON EXTENSION hstore IS 'data type for storing sets of (key, value) pairs
 
 
 --
+-- Name: pg_stat_statements; Type: EXTENSION; Schema: -; Owner: -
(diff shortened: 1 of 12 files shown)


db/structure.sql
```diff
@@ -2178,7 +2256,9 @@ CREATE TABLE public.team_submissions (
     solves_health_problem_description character varying,
     solves_education boolean,
     solves_education_description character varying,
-    scratch_project_url character varying
+    scratch_project_url character varying,
+    uses_gadgets boolean,
+    uses_gadgets_description character varying
```
viviancan, line 2261: I followed the existing pattern for "descriptions" and added the `uses_gadgets_description` to the submissions table

PR #4980, opened 2024-08-26
PT2 - Added uses gadget additional question frontend and other updates

viviancan:
Refs #4896 

This is part 2. This PR adds the frontend work for the `uses gadgets` additional question. 
The following was added
- JS to show/hide other gadget description text box
- gadget types to the submission grid
- Updated the submission form to include the uses gadget question in addition to the yes/no dropdown
- Updated the additional question section overview to display uses gadget details 
- Validation for at least 1 gadget type being selected if user selects "yes"
- Updated the `before_commit` callback to `before_save`. I will leave my notes in a comment

https://github.com/user-attachments/assets/303bd444-fe76-4a7b-af61-25a77b750873

### DIFF
--- a/app/models/team_submission.rb
+++ b/app/models/team_submission.rb
@@ -63,11 +63,15 @@ class TeamSubmission < ActiveRecord::Base
     self.pitch_video_link = standardize_url(pitch_video_link)
   }
 
-  before_commit -> {
+  before_save -> {
     self.ai_description = "" if ai.blank?
     self.climate_change_description = "" if climate_change.blank?
     self.game_description = "" if game.blank?
-    self.gadget_types = [] unless uses_gadgets?
+    self.solves_education_description = "" if solves_education.blank?
+    if uses_gadgets.blank?
+      self.gadget_types = []
+      self.uses_gadgets_description = ""
+    end
   }
 
   after_commit :update_student_info_in_crm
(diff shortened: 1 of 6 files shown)


app/models/team_submission.rb
```diff
@@ -63,11 +63,15 @@ class TeamSubmission < ActiveRecord::Base
     self.pitch_video_link = standardize_url(pitch_video_link)
   }
 
-  before_commit -> {
+  before_save -> {
```
viviancan, line 66: Switched before_commit to before_save to ensure descriptions are cleared.

I discovered that descriptions were not being cleared as expected. For example, if you selected yes on one of the additional questions and added a description then later selected no or the blank option, the description would not be removed from the record. 

Upon researching before_commit, I found/think that it is not intended for public use and is marked with :nodoc: in the core docs.
-  [link 1](https://github.com/rails/rails/issues/31936)
-  [link 2](https://github.com/rails/rails/issues/25896)
- [link 3 and most recent](https://github.com/rails/rails/issues/44458)

Switching to before_save resolved the issue, ensuring descriptions are removed when selecting 'no' or the 'blank' option.

viviancan:
> LGTM! 👍 :shipit:
> 
> The only feedback is w/ the part 1, part 2 PRs, where I have to toggle between the two and like where things are added in part 1 only to be removed in part 2 (happened in the last one too I think). Anywho, for the future 1 PR would be easier to review, IMHO. 😅

Sorry about that. I thought it would be more difficult due to the number of files changed. In the future, I will keep all related work to 1 PR 👍

Issue #4968, opened 2024-08-21
Add seasons valid column


dboyer:
As part of #4942 , we need to add a "seasons valid" column to the database so that an agreement can be set to less than the standard three years.

## RAKE TASK

A new rake task was added that will set the new season expires field for all signed chapter affiliation agreements.

- [x] QA: 
`heroku run bundle exec rake add_season_expires_to_chapter_affiliation_agreements --app technovation-qa`

- [x] Prod: 
`heroku run bundle exec rake add_season_expires_to_chapter_affiliation_agreements --app technovation`

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/4968?planning-poker) with Zenhub @shaun-technovation @viviancan

dboyer:
Confirmed with the chapter team that the custom agreements will likely be less than 3 years. The one with the school district is for one year.

dboyer:
I agree that having the agreement expire in the third and final season makes sense. In that scenario, it would still be valid for the third season, correct? Like an agreement that expires in 2027 would still allow the chapter to be onboarded for the 2027 season.

I'm checking on how many seasons the ChA volunteer agreement should be valid for.

dboyer:
Tested on QA and confirmed that I now see four fields related to chapter agreement expiration:

- Signed date
- Season it was signed
- Season it expires
- Seasons for which it is valid
<img width="432" alt="Screen Shot 2024-09-09 at 11 23 08 AM" src="https://github.com/user-attachments/assets/bfe84b05-f1be-43cf-add5-41baae798a0f">

I completed a new agreement and the seasons updated as expected. When I viewed an already signed agreement, the values were filled in for all the fields.

dboyer:
Confirming that I see the new fields on production. 

Can we run the rake task to update existing signed agreements @shaun-technovation?

PR #4982, opened 2024-08-27
Add season expires to documents

shaun-technovation:
This will add a new `season_expires` field to our documents table. When a chapter affiliation agreement is signed we will set this field to the season the document expires in. We already have a seasons valid method, I just updated it to be based on the season signed and season expired fields.

I also added a rake task to set this new `season_expires` field for all signed chapter affiliation agreements.

I will test the webhook portion of this on QA.

Addresses: #4968


viviancan (approved):
💯 LGTM

PR #5008, opened 2024-09-09
Off-platform Chapter Affiliation Agreements

shaun-technovation:
This will add basic functionality to allow admins to create an off-platform chapter affiliation agreement.

An affiliation agreement is now considered complete (for onboarding purposes) if it's either signed or off-platform.


viviancan (approved):
LGTM! The only thing I noticed is that the void button is visible. Not sure if we want the button visible for off-platform agreements since it seems to be for docusign agreements! I wonder if we will also need to be able to void off-platform agreements.

dboyer:
We shouldn't need to void off-platform agreements as we probably won't even hear about them until they're already signed.

PR #5012, opened 2024-09-11
Add Salesforce duplicate rule request header

shaun-technovation:
This will make it so that if Salesforce detects a duplicate record it will save the duplicate record; this is safe to do because we upsert records in Salesforce by their account id.

More info about `Sforce-Duplicate-Rule-Header`:

- https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/headers_duplicaterules.htm
- https://salesforce.stackexchange.com/questions/380814/acknowledging-duplicate-rule-alerts-when-creating-records-via-rest


viviancan (approved):
LGTM! 💯

PR #5021, opened 2024-09-13
Chapter selection groundwork

shaun-technovation:
This will setup the groundwork for chapter selection for students. The main pieces are:

- a new service to find chapters
  - if we revamp our locations, only this service would need to be
    updated
- the chapter selection interrupt screen
  - this will only displayed to students now

Also, there isn't a way to save the chapter selection yet, that will happen in #5030.


viviancan (approved):
Really nice groundwork here! Excited to see this move forward. The only feedback I have is on the spacing of the text. I think the left alignment looked a little off in the energetic container. 
CleanShot 2024-09-25 at 09 18 38

PR #5035, opened 2024-09-23
Update first/last name validation

shaun-technovation:
This will make it so that a first name or last name can't begin with one or more dashes or periods; dashes and periods are fine anywhere else though.


viviancan (approved):
LGTM, nice update on this 💯

Issue #5036, opened 2024-09-24
Phone number updates


dboyer:
Two updates requested by chapter ambassadors:

- [x] Add phone number as an additional column on the participants, mentors, and judges admin pages - this should be visible to both ChAs and admins and be exportable
- [x] Add formatting to the phone number field (for ChAs, judges, mentors) that only allows entering numbers and not letters. Ideally it would be helpful to use an existing gem that applies phone number formatting but even just restricting to numbers would be an improvement.

viviancan:
This is on QA. It can be tested as a logged in mentor, judge, and ChA. It can also be tested in the admin portal when editing a participant.

dboyer:
Confirming on QA that I can add the phone number as an additional column on the participants page and it is exportable.

@viviancan Could the phone number also be added to /admin/judges and /admin/mentors? 

Confirming that I can only add numbers, dashes, and plus signs to the phone number field. This is true when I'm logging in as the individual or editing the info as an admin.

viviancan:
@dboyer Yes, will add it to the other datagrids!

dboyer:
Confirmed that phone number is now on the mentor and judge admin pages and can be exported.

viviancan:
Confirming this is on prod

PR #5051, opened 2024-09-30
Added phone number validation to only allow +, -, numbers, and spaces

viviancan:
Refs #5036 

This PR adds basic phone number validation to only allow +, -, numbers, and spaces.

I checked prod DB and found the following variations of phone numbers
- +11234516791
- +1 1234516791
- 1234516791
- 123 4516791
- 123-451-6791

Some had country code, a country code with leading + sign, no country code, dashes, no dashes, and spaces. I thought allowing + sign, dashes, and numbers fit the variety best. I also researched a phone number [validator/formator](https://github.com/daddyz/phonelib) gem that could be helpful for this functionality. Especially given the variety of formats we already have. I did mock it out, and it seems like it would integrate nicely. 

I also added the phone number field to the participants datagrid so it can be exported.

### DIFF
--- a/app/views/admin/participants/edit.html.erb
+++ b/app/views/admin/participants/edit.html.erb
@@ -39,10 +39,10 @@
 
     <% if !f.object.student_profile.present? %>
       <p>
-        <%=  f.label :phone_number  %>
+        <%= f.label :phone_number %>
         <%= f.text_field :phone_number %>
       </p>
-    <%  end %>
+    <% end %>
 
     <p>
       <%= f.label :gender,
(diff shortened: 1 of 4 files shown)


app/views/admin/participants/edit.html.erb
```diff
@@ -64,16 +64,6 @@
             <br>
           <% end %>
         </p>
-
-        <% if f.object.mentor_profile.errors.present? %>
```
viviancan, line 68: I removed these chunks of code since they displayed any errors multiple times. This also caused issues with the phone number errors. I verified that the errors would still display appropriately - just not multiple times
See screenshot 
Screenshot 2024-09-30 at 9 59 45 AM

PR #5055, opened 2024-10-01
Save chapter assignments

shaun-technovation:
The main change here is that this will allow someone to save their chapter selection, including the "I'm not sure" option. After their selection is saved, the force chapter selection screen will no longer be displayed for them.

A lot of the other changes are to support the saving functionality.


viviancan (approved):
Everything worked as expected! One thing that we may want to standardize is the striped green header. On the location selection page it is there, but then when the chapter selection page displays it's not. I think it would be good to either have it there for both or neither. Just an idea! Besides that, awesome job on this! So great to see it start coming together 👍 🚀

PR #5058, opened 2024-10-02
Update onboarding status for off-platform documents

shaun-technovation:
An off-platform document is "created" which isn't triggering the `after_update` onboarding status update. 

A simple change here that will change the `after_update` hook to `after_save` which will cover both creating and updating. And a few other changes to support this change.


viviancan (approved):
Looks pretty straightforward! Nice update!

Issue #4201, opened 2023-09-14
Pre-populate invited email address during registration


shaun-technovation:
When an admin invites someone to register, the invited email address should be pre-populated on the registration form..

dboyer:
The invited email address should also be read only on the registration form so that the user cannot edit it to use a different email. We discovered that an invited ChA had forwarded the invite to someone else who changed the email address and the first email address could then no longer be used for a new invite. [Details here](https://iridescentlearning.slack.com/archives/C02J1GU8VRA/p1725562451448039).

viviancan:
@shaun-technovation I can't move this card into code review! It won't save - but there is a PR open now for it! Ty!

viviancan:
This is on QA

dboyer:
Tested on QA and confirmed:

- I was not able to edit the email address when I was sent a ChA invite
- I was able to edit the email address on my profile tab after I completed registration
- I invited a mentor and the address was pre-filled but editable in the registration flow
- I invited a judge and the address was pre-filled but editable in the registration flow
- I invited a student and the address was pre-filled but editable in the registration flow

PR #5067, opened 2024-10-04
Pre populate email address in the registration form when participant is invited 

viviancan:
Refs #4201 

This PR adds functionality to pre-populate the registration form's email address when a participant is invited. This will apply to any invited user, however, the email field will only be disabled for an invited chapter ambassador (the other profile types can still edit the field).

Main changes

- Added `UserInvitationEmailsController which follows the `ChapterOrganizationNamesController` pattern. 
- Added `getUserInvitationEmail()` method
- Disabled the email form field for chapter ambassador profile type
- Added test

### DIFF
--- a/app/controllers/api/registration/user_invitation_emails_controller.rb
+++ b/app/controllers/api/registration/user_invitation_emails_controller.rb
@@ -1,0 +1,11 @@
+module Api::Registration
+  class UserInvitationEmailsController < ActionController::API
+    def show
+      invite = UserInvitation.find_by(admin_permission_token: params[:invite_code])
+
+      render json: {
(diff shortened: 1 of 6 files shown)


viviancan:
> LGTM! Nice addition! The only thing I noticed is that our formatting is inconsistent, when I edit one of these Vue/JS files the formatting will be different - not saying my way is the right way (I can't remember what I have setup, Prettier I think); but we should make our styling consistent, otherwise we'll just keep overwriting each other's styles. I created #5071 to address this. 👍

Ahh ok, yes I agree. I remember setting up Prettier too (a while ago). I will double-check my local settings.

PR #5069, opened 2024-10-07
Off-platform Chapter Volunteer Agreements

shaun-technovation:
This will add functionality that will admins to create off-platform Chapter Volunteer Agreements for chapter ambassadors. This functionality will only be available if the chapter ambassador doesn't already have a Chapter Volunteer Agreement.


viviancan (approved):
Screenshot 2024-10-08 at 12 12 04 PM
Should we update the text on the ChA dashboard if the agreement is off platform?

Great work on this, works as expected!

Issue #5073, opened 2024-10-08
US Mentor background check cleanup 


viviancan:
- [x] Remove `in_background_check_invitation_country?` method
- [x] Update any logic related to ^^

dboyer:
Tested on prod and confirmed that mentors in the US are prompted to complete a background check and it appears that checks have cleared since the deployment.

Issue #5056, opened 2024-10-01
Adjust US mentor background check process to use Checkr invite process


dboyer:
The US mentor process currently uses a custom form hosted on the TG site to send information to Checkr. Remove that and integrate the same Checkr invitation system used for ChAs.

Note that this does not include removing all the old code for the old pages.

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/5056?planning-poker) with Zenhub @shaun-technovation @viviancan

dboyer:
Confirming that the package name is still Tasker Standard for US candidates.

viviancan:
This is on QA

dboyer:
Tested on QA and was able to:

- Request a Checkr invitation
- Received a link in my email and clicked through to Checkr
- Was prompted to complete the form
- The background check page on my profile updated the status to "pending" after I completed the form
- I manually engaged the candidate and the check was marked as clear. I was able to access the page to find a team.

dboyer:
Confirming the new Checkr invite system is in place for US mentor background checks.

PR #5076, opened 2024-10-09
Update mentor background check to use invitation flow

viviancan:
Refs #5056 
This PR updates the mentor background check flow to use the checkr invitation system similar to international participants and ChAs.

Main changes:
- Removed "US" conditional that separated the 2 tracks
- Extracted the invitation status code from ChA debugging into a separate partial
- Update the mentor debugging section to use the invitation status partial (now used for ChAs and Mentors - so they match)
- Updated/added tests

Also created #5073 to complete any cleanup.

### DIFF
--- a/app/models/mentor_profile.rb
+++ b/app/models/mentor_profile.rb
@@ -295,6 +295,12 @@ class MentorProfile < ActiveRecord::Base
       background_check.clear?
   end
 
+  def background_check_flagged?
+    background_check.consider? ||
+      background_check.suspended? ||
+      background_check.suspended?
+  end
+
   def requires_background_check?
     (account.valid? && (account.date_of_birth.present? && account.age >= 18 || account.meets_minimum_age_requirement?)) &&
       in_background_check_country? &&
--- a/app/models/mentor_profile.rb
+++ b/app/models/mentor_profile.rb
@@ -295,12 +295,6 @@ class MentorProfile < ActiveRecord::Base
       background_check.clear?
   end
 
-  def background_check_flagged?
-    background_check.consider? ||
-      background_check.suspended? ||
-      background_check.canceled?
-  end
-
   def requires_background_check?
     (account.valid? && (account.date_of_birth.present? && account.age >= 18 || account.meets_minimum_age_requirement?)) &&
       in_background_check_country? &&
--- a/app/models/mentor_profile.rb
+++ b/app/models/mentor_profile.rb
@@ -319,7 +319,6 @@ class MentorProfile < ActiveRecord::Base
 
   def in_background_check_invitation_country?
     country_codes = ENV.fetch("BACKGROUND_CHECK_COUNTRY_CODES", "").split(",")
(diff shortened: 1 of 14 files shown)


app/models/mentor_profile.rb
```diff
@@ -313,7 +319,6 @@ def in_background_check_country?
 
   def in_background_check_invitation_country?
```
viviancan, line 320: Logging this in #5073 - US Mentor background cleanup. This method should be removed and any logic related to it should be updated.

app/views/admin/participants/_background_check_invitation_status.html.erb
```diff
@@ -1,47 +1,23 @@
...
-          text: profile.background_check.invitation_status.humanize
-        ) %>
-  <% elsif profile.background_check.invitation_expired? %>
-    <%= web_icon(
-          "exclamation-circle icon-red",
-          text: profile.background_check.invitation_status.humanize
-        ) %>
-    <br>
-    <span class="hint">
+<% if profile.background_check.invitation_status.present? && !profile.background_check.invitation_completed? %>
```
viviancan, line 1: Extracted from the cha onboarding section to it can be used on ChA and mentor debugging sections

app/views/admin/participants/_mentor_debugging.html.erb
```diff
@@ -34,14 +34,6 @@
       <% end %>
     </dd>
 
-    <dt>
-      Background Check Invitation Status
```
viviancan, line 38: Removed the separated bg check statuses so this can mirror the ChA debugging view

app/views/background_checks/_show.html.erb
```diff
@@ -40,10 +40,9 @@
 
       <h5>Possible status values:</h5>
 
-      <dl>
+      <dl class="padding--l-large">
         <dt>Invitation Required</dt>
         <dd>
-          You are an international candidate.
```
viviancan, line 46: Removed this since now all mentors in bg check countries will go through invitation flow

viviancan:
> > Random question/thought, when we deploy this, will we have to support mentors who still have an outstanding background check through the old/current system? 🤔
> 
> Responding to my own comment, is this following correct: there are currently two paths into Checkr, but there is currently only one path back to our platform? If so, then I think we'd be good then, right? Since we'd only be affecting how people get into Checkr...🤔

Yes, right now we 2 ways into checkr. This will just affect how the bg check gets started. I tested multiple scenarios and I believe if there is already a pending bg check this will not affect the process!

app/models/mentor_profile.rb
```diff
@@ -295,6 +295,12 @@ def background_check_complete?
       background_check.clear?
   end
 
+  def background_check_flagged?
```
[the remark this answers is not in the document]
viviancan: Ok, ty for the feedback! I struggled with a name to describe the non clear background checks. Maybe just `not_cleared?` open to any other suggestions!

PR #5079, opened 2024-10-10
Update student to mentor conversion functionality

shaun-technovation:
For when a student who will be over 18 years old by the cutoff date signs-in to a new season:

- They won't get registered to the current season as a student
- If they choose to become a mentor, they will get registered to the current season as a mentor (including receive the mentor welcome email), this is new functionality


viviancan (approved):
LGTM 🚀

PR #5085, opened 2024-10-17
Legal documents datagrid and void functionality

shaun-technovation:
This will add a basic datagrid for legal documents (chapter affiliation agreements and chapter volunteer agreements). It will also add functionality to void a sent/unsigned document.


viviancan (approved):
LGTM 💯

PR #5102, opened 2024-10-21
Associate ChA to chapter after registering via an invite

shaun-technovation:
This will use our new chapter-account assignments way of associating chapter ambassadors to chapters for when a chapter ambassador registers with an invite; which should properly assign them to their chapter so it displays correctly in the GUI.


viviancan (approved):
LGTM! Works as expected

PR #5103, opened 2024-10-22
Update mentor chapter selection functionality

shaun-technovation:
Some updates to the chapter selection screen/process:

- Making the text area dynamic so students and mentors can have different text displayed

- Adding team matching options for mentors to pick from


app/views/chapter_account_assignments/_extra_mentor_form_fields.en.html.erb
```diff
@@ -0,0 +1,23 @@
+<%= current_account.id %>
```
viviancan, line 1: I think these lines might have been accidentally left in!

viviancan (approved):
LGTM! Works as expected 💯 Nice updates to this

PR #5112, opened 2024-10-24
Update onboarding status for off-platform legal docs

shaun-technovation:
A very simple fix for a very annoying bug, this will address an issue where the chapter and chapter ambassador's onboarding statuses weren't getting updated for an off-platform legal agreement. The onboarding check was basically using a stale copy of the legal agreements. Using `reload` will cause the legal agreements to be refreshed to ensure the onboarding check is using the updated version.


viviancan (approved):
LGTM! Nice fix!! 💯

PR #5117, opened 2024-10-28
Update chapter ambassador datagrid onboarding functionality

shaun-technovation:
This will update the chapter ambassador datagrid to accommodate for the new chapter-account assignment way of doing things. I also addressed some Bullet warnings, which are the other changes here.


viviancan (approved):
LGTM!

PR #5118, opened 2024-10-29
Add rake task to convert primary ChA location to chapter

shaun-technovation:
This will add a rake task that will convert a primary chapter ambassador's location to their chapter, if the chapter is missing a location.


viviancan (approved):
Did not try locally, but looks pretty straightforward! This will be great for missing locations! 💯

PR #5120, opened 2024-10-29
Add missing chapter timestamps

shaun-technovation:
When I initially created this table I forgot to add the timestamps. 🤦 Anywho, this will add the `created_at` and `updated_at` columns to the chapters table.


viviancan (approved):
💯

PR #5133, opened 2024-11-01
Update datagrid search result total count

shaun-technovation:
In b8889b3c7 I updated `.count` to `.length` b/c the chapter ambassador datagrid wasn't working correctly w/o it, but now `.length` isn't working correctly, and more changes have been made to the chapter ambassador datagrid and that datagrid seems to work now with `.count`.

Anywho, this change it basically to revert `.length` to `.count`.


viviancan (approved):
It looks like this needs to be rebased, but I reviewed the change and tried locally. I was able to see updated total counts

PR #5137, opened 2024-11-04
Update mentor searchability check for background check exemptions

shaun-technovation:
This will make it so that if an admin grants or revokes a background check exemption for a mentor, it will trigger a check to see if the mentor's searchability flag should be enabled or not.


viviancan (approved):
Tried locally and my "searchable" flag changed true with the bg ground check exemption 💯

PR #5140, opened 2024-11-04
Don't display ChA legal agreement status if there isn't one

shaun-technovation:
There are some cases on Preview where a chapter ambassador's chapter affiliation agreement doesn't have a status which is causing an error. I'm not sure how this is happening (it could just be old, corrupt testing data). Anywho, for the time being, this update will make it so their status won't be displayed if there isn't one.


viviancan (approved):
Nice update! I think I encountered this error locally, but hadn't tracked it down yet. LGTM 👍

PR #5143, opened 2024-11-04
Display primary ChA name on chapter selection screen

shaun-technovation:
This will display the primary chapter ambassadors name (if one exists) on the chapter selection screen.


viviancan (approved):
Tried locally, lgtm 👍

PR #5159, opened 2024-11-07
Chapter selection updates

shaun-technovation:
A pretty big refactoring that will create 8 different permutations of the chapter selection screen, for both students and mentors:
- Chapters found only in state/province
- Chapters found in state/province AND country
  - State/province chapters will be displayed by default
  - There will now be a button to display additional chapters in someone's country
- Chapters found only in the country
  - No chapters will be displayed, they will have to click a button to display additional chapters in their country
- No chapters found

And a few other updates:
- Add mentor mentoring options to no chapters available screen
- Update some of the text


viviancan (approved):
LGTM 👍 Tried locally, and I was able to click on the "see all chapters" button, and was able to select a chapter.

PR #5175, opened 2024-11-13
Update admin student-chapter assignment functionality

shaun-technovation:
A small update here that will make it so that when an admin updates a student-chapter assignment it will now remove them from the unaffiliated list, which will trigger the "assigned to chapter" email.


viviancan (approved):
Did not try locally, but LGTM 👍

Issue #5109, opened 2024-10-24
Update Meet Your ChA for mentors


viviancan:
Refs #4814

Update `app/views/dashboards/_chapter_ambassador_intro.en.html.erb`

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/5109?planning-poker) with Zenhub @shaun-technovation @viviancan

dboyer:
Tested on QA and confirmed that:

- I registered as a mentor with no chapter in my country. The Meet your ChA section showed that no chapter was available.
- I logged in as a mentor assigned to a chapter and it displayed the correct information for the chapter and ChA
- I logged in as a mentor and selected a chapter. The correct ChA and chapter info displayed.

PR #5177, opened 2024-11-14
Update meet your cha for mentors to use their selected chapter/primary cha information

viviancan:
Refs #5109


app/javascript/components/DashboardHeader.vue
```diff
@@ -26,20 +26,19 @@
           <small v-if="surveyLink">
             <a :href="surveyLink" target="_blank">{{ surveyLinkText }}</a>
           </small>
+
+          <small>
+            <drop-down label="More Details">
```
viviancan, line 31: Added "more details" since now if mentors are not affiliated with a chapter, they will see the not affiliated text

PR #5178, opened 2024-11-14
Email mentors when they're assigned to a chapter

shaun-technovation:
This will send a notification email to a mentor after they have been assigned to a chapter.


viviancan (approved):
This worked as expected. Just noting that it only worked for ChAs. I know this was brought up before, but wasn't sure if that had been resolved!

PR #5186, opened 2024-11-18
Add Heroku release phase tasks

shaun-technovation:
This will add release phase tasks for Heroku (moving them from `DEPLOY_TASKS` to Heroku's `Procfile`).

We're currently using an outdated, no-longer-supported plugin for this: https://github.com/gunpowderlabs/buildpack-ruby-rake-deploy-tasks

This outdated way^ relies on an ENV named `RELEASE_TASKS`. I copied the `RELEASE_TASKS` from the Heroku ENV setting and moved them to the Heroku `Profile`.

More info about the Heroku release phase being used: https://devcenter.heroku.com/articles/procfile#the-release-process-type


viviancan (approved):
Wasn't sure what to test, but lgtm 👍 I checked out the documentation!

Issue #5171, opened 2024-11-11
Create club model in the database


dboyer:
Create a club model in the database. The club should consist of:

- Club name
- Club location
- Club summary
- Club links
- Primary contact (assigned club ambassador)
- Participants table
- Teams table
- Submissions table

The club will not have any associated legal docs.

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/5171?planning-poker) with Zenhub @shaun-technovation @viviancan

dboyer:
Confirming that it is now possible to create and view clubs on preview

dboyer:
Confirming that it is now possible to create and view clubs on QA.

PR #5188, opened 2024-11-18
Added club migration

viviancan:
Refs #5171 

This PR adds the club migration


viviancan:
> I need to add a few additional fields to this

@shaun-technovation this is ready! Added visible on map and headquarters location

PR #5204, opened 2024-11-26
Assign mentor to chapters when an invitation is accepted

shaun-technovation:
When a mentor accepts an invitation from a student, this will assign the mentor to the unique chapters of the students on the team.


viviancan (approved):
Works as expected!

PR #5216, opened 2024-11-27
Update cross-spawn node module

shaun-technovation:
This will update the cross-spawn node module that should address these security vulnerabilities:

- https://github.com/Iridescent-CM/technovation-app/security/dependabot/230
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/229


viviancan (approved):
LGTM!

PR #5220, opened 2024-12-02
Distinguish between no chapter selected and no chapters available participants

shaun-technovation:
This will separate "None of the above" aka `no_chapter_selected` participants from participants who don't have a chapter in their area, which will categorized as `no_chapters_available` participants now. 

This will also let admins filter on "No chapter selected" and "No chapters available" on the unaffiliated participants datagrid.

There's quite a bit to chapter selection, I think I covered all areas, but I might have missed something too.


viviancan (approved):
LGTM! 

I think it may be nice to add something on the admin participants detailed page to distinguish between the unaffiliated type.

CleanShot 2024-12-04 at 13 52 03@2x

PR #5222, opened 2024-12-03
Update rake task that adds students to chapters

shaun-technovation:
The main change here will make it so that if a student is not registered to the current season, they will not be assigned to a chapter.

In order to add a student to a chapter, the following conditions must be met:
- Must be a valid account/email address
- Must be a student account
- Must be registered to the current season
- Must not be assigned to a chapter already


viviancan (approved):
Did not try locally, but lgtm!

Issue #5209, opened 2024-11-27
Club ambassadors can register on the platform


dboyer:
After receiving an invite (added as part of #5169), club ambassadors should be able to register on the platform. The registration form should include the fields [outlined here](https://docs.google.com/spreadsheets/d/11Wy_yT3wsFrn6y1X9KvbhiFUlBn76AK34RToMMajhpE/edit?gid=0#gid=0)

- First name
- Last name
- Age confirmation (Check box "I confirm that I am 18 years or older")
- Gender identity
- Job title
- Club (pre-filled by admin)
- Phone number (optional)
- How did you hear about Technovation (optional)
- Email address (pre-filled by admin - not editable, same as ChA)

NOTE: Need to confirm that the location screen automatically displays after registration. But NOT the chapter/club selection.

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/5209?planning-poker) with Zenhub @shaun-technovation @viviancan

viviancan:
@dboyer This is on PREVIEW

- Just calling out that once you complete registration the platform "errors" out since the CLA dashboard is in code review. You will have to manually go to /signout to signout! 
- Also I added #5253 which will update the flow to actually create the club assignment for the CLA. This wasn't possible until the refactor work was completed. So for this ticket, the CLA can be invited and register but the assignment won't be there yet.

dboyer:
Confirming on preview:

- I received an email invite to register as a club ambassador
- The club name was prefilled
- The registration fields match the list in this GH issue
- After completing registration, I was taken to the dashboard

NOTE: I did not see a location form. I'll open a separate issue for that.

dboyer:
Confirming that I received an invite to register as a club ambassador and was able to complete the registration process.

dboyer:
VET has been sending invites to club ambassadors and they have been able to register on production. I see new club ambassadors on the data grid.

Issue #5169, opened 2024-11-11
Admins can invite club ambassadors to register


dboyer:
Club ambassadors will only be able to register via invite (/admin/user_invitations). Admins will associate a club ambassador with a club. Fields for the invite are:

- Name (optional)
- Email address (required)
- Club with which they should be associated (required)

Club ambassadors should be able to register at any time.

viviancan:
This is on preview

dboyer:
Confirmed on preview that I could:

- See clubs on the registration invites page
- Invite a club ambassador to register for a particular club
- See the pending invite on the invites datagrid
- Get an email as the club ambassador and when I clicked through the club name was pre-filled
- I was able to complete registration and ended on my dashboard
- On the admin pages, the registration invite updated and the user now appears on the club ambassador datagrid

dboyer:
Also confirmed the ChA invites were working as expected

dboyer:
Confirmed on QA that admins can invite a club amb to register. Also confirmed that inviting a ChA was still working as expected.

PR #5229, opened 2024-12-05
Club Ambassadors can register after receiving an invitation 

viviancan:
Refs #5209 

This PR enables club ambassadors to register on the platform after receiving an invite (from #5169). 

Updates include:
- Updating registration flow to include club ambassador fields. 
- Ensured club and email fields are pre-filled based on the invitation
- Added club name API 
- Updated the registration controller and services to handle club ambassadors
- Included tests and validations to support club ambassadors


app/models/club_ambassador_profile.rb
```diff
@@ -5,7 +5,8 @@ class ClubAmbassadorProfile < ActiveRecord::Base
 
   belongs_to :current_account, -> { current },
     class_name: "Account",
-    foreign_key: "account_id"
+    foreign_key: "account_id",
+    required: false
```
viviancan, line 9: @shaun-technovation I removed the `required: false` option in [this thread](https://github.com/Iridescent-CM/technovation-app/pull/5226#discussion_r1876760162). However now, the tests are failing and I am unable to actually complete my registration as a club ambassador. All of the other profile types have this setup, so now I'm unsure of how to proceed. I am adding it back in but maybe we should investigate this further. The error I kept getting was `Current account must exist`.

PR #5231, opened 2024-12-06
Chapterables groundwork and saving clubs

shaun-technovation:
This will introduce the concept of something being chapterable, i.e. something chapter-like, something that can either be a chapter or a club; this will allow one bucket for chapterable assignments that can be used for either chapters or clubs (or any future chapter-like thing).

The work in this PR just lays the groundwork for chapterable assignments.

Some examples of how chapterables will be used:

- a chapterable assignment would be for a club or a chapter
- a chapterable view partial would be able to handle a chapter or a club
- a student's chapterable would be their chapter or club
- a chapter ambassador's chapterables would be their chapters
- a club ambassador's chapterables would be their clubs

### DIFF
--- a/app/views/chapterable_account_assignments/_additional_clubs.en.html.erb
+++ b/app/views/chapterable_account_assignments/_additional_clubs.en.html.erb
@@ -16,9 +16,11 @@
   </div>
 
   <div id="additional-clubs" class="hidden">
-    <h2 style="font-size: 1.875rem; font-weight: 600; color: #1f2937; margin-top: 0;">
-      Clubs
-    </h2>
+    <% if @primary_clubs.empty? %>
+      <h2 style="font-size: 1.875rem; font-weight: 600; color: #1f2937; margin-top: 0;">
+        Clubs
+      </h2>
+    <% end %>
 
     <% @additional_clubs.each do |club| %>
       <%= render "chapterable", chapterable: club %>


lib/tasks/add_chapterable_type_to_existing_chapter_account_assignments.rake
```diff
@@ -0,0 +1,6 @@
+desc "Add chapterable type to existing chapter assignments"
+task add_chapterable_type_to_existing_chapter_assignments: :environment do |_, args|
+  ChapterableAccountAssignment.update_all(chapterable_type: "Chapter")
+
+  puts "Sucessfully converted existing chapter assignments to new chapterable model"
```
viviancan, line 5: Typo on 'Sucessfully'

viviancan (approved):
Amazing work on this 💯 Wow!

Confirming that locally I
- ran the rake task to convert all current assignments to the "chapter" chapterable type
- Registered as a student and joined a club. I can see Club as the chapterable type in the db
- Registered as a mentor and joined a chapter. I can See Chapter as the chapterable type in the db

For the student in the club, the green header needs to be updated, which we all already discussed! Great work on this Shaun! I know there are a lot of other scenarios to test so it would def be good to get other 👀 on it once on preview!

PR #5254, opened 2024-12-17
Ability for admins to add club ambassador profile to mentors

shaun-technovation:
This will let admins add a club ambassador profile to an existing mentor account.


viviancan (approved):
Works as expected! If the mentor is already assigned to a primary chapter should that assignment be removed or at a minimum not be the primary assignment?

Issue #5248, opened 2024-12-16
Display club affiliation in the Meet Your Ambassador header


dboyer:
For both students and mentors

viviancan:
[This PR](https://github.com/Iridescent-CM/technovation-app/pull/5256) updates the meet your ambassador section for students. 

I will make a separate pr for mentor meet your ambassador section

dboyer:
Confirming that on QA I now see: 

- the correct club info displaying for students in the meet your ambassador header
- the correct chapter info displaying for students in the meet your ambassador header

viviancan:
@dboyer this is on QA for mentors

- Logged in as participant 3704 - unaffiliated - and saw the "not assigned to chapter or club" message
- Logged in as participant 3472 - assigned to Chapter Arizona Star - and saw Arizona star chapter info
- Logged in as participant 3472 - assigned to Chapter Valenciaaa (org name) with no chapter name set -and saw correct info
- Logged in as participant 4102 - mentor assigned to Los Angeles Coding Club and saw correct info

dboyer:
Tested on QA and confirmed:

- Existing mentor assigned to a club is now showing that club in the meet your ambassador section
- I selected a chapter and it showed the correct ChA info
- I selected a club and the club info showed for that mentor

PR #5256, opened 2024-12-17
Update student meet your Ambassador section to account for chapters, clubs, & unaffiliated chapterable assignments

viviancan:
Refs #5248 

This PR updates the tg green striped header and the nested meet your ambassador dropdown to account for chapter, clubs, & unaffiliated chapterable assignments.

I did a refactor because the tg header felt messy. The refactor separates out the different chapterable tracks (chapter assignment, club assignment, and unaffiliated participant). I think this also allows for more customization depending on the different needs of PT (for chapters) vs VET (for clubs). Mentor updates will be in separate PR.


viviancan:
> Some naming isn't quite lining up, but if it's working (didn't try it locally), then i think it's good enough to ship! 🚀

I will update the naming 👍

PR #5259, opened 2024-12-18
Update mentor meet your Ambassador section to account for chapters, clubs, & unaffiliated chapterable assignments 

viviancan:
Refs #5248 

Update mentor meet your Ambassador section to account for chapters, clubs, & unaffiliated chapterable assignments 

This PR addresses the shift from Chapter to Chapterable (chapter, club, unaffiliated mentors). I tried to remove chapter specific references in favor of "ambassador" references and then more specifically chapter and club when needed. I cleaned up the header just a bit and followed the same approach a the header for students. I don't want to spend too much extra time here since this is a vue app and we will be ripping it out eventually! I was hesitant to remove all references to chapter ambassador (for ex the serializer) since I am unsure if this will break other parts of the mentor experience. I don't think it will, but with the holiday time coming up, I feel it is better to be cautious.

### DIFF
--- a/app/assets/stylesheets/_application.scss
+++ b/app/assets/stylesheets/_application.scss
@@ -106,6 +106,7 @@ $chosen-sprite-retina: image-url("chosen-sprite@2x.png");
 }
 
 h1.page-heading .chapter-ambassador-intro,
+.ambassador-intro,
 .chapter-ambassador-intro {
   color: #333;
   font-family: "Source Sans Pro", sans-serif;
--- a/app/javascript/components/DashboardHeader.vue
+++ b/app/javascript/components/DashboardHeader.vue
@@ -1,23 +1,23 @@
 <template>
   <div class="grid dashboard-notices">
     <div class="grid__col-sm-6 grid__col--bleed-y">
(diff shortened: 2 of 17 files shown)


viviancan:
@shaun-technovation This is ready!

viviancan:
> My general comment is that I see a bunch of places where we're using "chapterable ambassador" and also "ambassador" in places. It would be nice to make this consistent and to also have a standard for these new terms.
> 
> My on the-spot suggestion:
> 
> * chapterable refers (only) to chapters or clubs
> * ambassador refers to chapter ambassador or club ambassadors
> 
> Thoughts^? 🤔
> 
> Chapterable ambassador takes an extra second to determine if it's a chapter/club, or chapter ambassador/club ambassador, but if chapterable _only_ referred to chapters/clubs it would make things simpler/more straight forward, imho. Anyhwo, not a blocker, I think if it's working as expected, I think it's good as is so we can deploy. 👍

I think you make a great point! I will merge this in and then make additional updates to just use ambassador when referencing either club or chapter ambassador. Ty for the feedback

PR #5267, opened 2024-12-18
Assign club ambassador to club after registering

shaun-technovation:
After a club ambassador registers via an invite, this will make it so that they get assigned to the club associated with the invite.


viviancan (approved):
LGTM and works as expected 🎉

PR #5273, opened 2024-12-20
Restrict which participants ChAs can view

shaun-technovation:
This will allow ChAs to only view participants who belong to their chapter, and it will remove all instances of `allow_out_of_region`.


viviancan (approved):
LGTM! I checked a few different profiles and submissions as a ChA and didn't encounter any error pages. nice update 💯

PR #5274, opened 2024-12-23
Send separate parent/guardian first and last names to Salesforce

shaun-technovation:
This will split up the parent/guardian name into two parts before sending it to Salesforce, one part (the first part separated by whitespace) will be sent as the first name, and the remaining part will be sent as the last name.

And then in Salesforce the first and last name can be combined to form the full name like it is on the platform.

### DIFF
--- a/app/models/student_profile.rb
+++ b/app/models/student_profile.rb
@@ -141,6 +141,14 @@ class StudentProfile < ActiveRecord::Base
     end
   end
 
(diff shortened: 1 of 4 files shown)


viviancan (approved):
Did not try locally, but looks pretty straightforward! Nice update and good idea for the future if we split in the db 💯

PR #5275, opened 2025-01-03
Club ambassadors can edit location and public info

shaun-technovation:
This will add the Public Info and Club Location sections (under the Club Profile tab).

Issue #5211, opened 2024-11-27
Create club profile tab on club ambassador dashboard


dboyer:
On the club ambassador dashboard (created in #5210), club ambassadors should have a "Club Profile" tab (similar to Chapter Profile) that includes these tabs. Editing this info will be covered in a different GH issue.

1. Public info
- Club name (set by admin and cannot be edited by club ambassador)
- Program summary (text field with 1000 character limit)
- Primary contact (load list of club ambassadors)
- Public map visibility (yes/no)

2. Club location
- Headquarters location (reuse from Chapters)
- Standard location form

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/5211?planning-poker) with Zenhub @shaun-technovation @viviancan

viviancan:
This is on preview. The links will be updated with [Shaun's PR](https://github.com/Iridescent-CM/technovation-app/pull/5275)

dboyer:
Confirming I now see the club profile tab with two links in the sidebar to public info and club location.

dboyer:
Club profile tab is working on QA.

dboyer:
Confirming I see the club profile tab on production.

PR #5281, opened 2025-01-06
Added club ambassador club profile tab

viviancan:
Ref #5211 

This PR adds the foundation for the club ambassador club profile tab. The navigation will need to be updated to account for the club public info and club location work completed in [this PR ](https://github.com/Iridescent-CM/technovation-app/pull/5275).


app/views/club_ambassador/club_profile/_side_nav_content.html.erb
```diff
@@ -0,0 +1,17 @@
+  <nav class="p-4" aria-label="Progress">
```
viviancan, line 1: These links need to be updated

PR #5282, opened 2025-01-06
Morocco

shaun-technovation:
This follows our existing approach for addressing Palestine, except this is for Morocco. I just copied the Palestine functionality and updated it for Morocco and it seems to be working. We'll need to test further to see if this has any unintended side effects.

### DIFF
--- a/app/technovation/handle_geocoder_search.rb
+++ b/app/technovation/handle_geocoder_search.rb
@@ -6,6 +6,8 @@ module HandleGeocoderSearch
 
     if geocoder_query.country_code == "PS"
       handle_palestine_search(db_record, geocoder_query, controller)
+    elsif geocoder_query.country_code == "MA"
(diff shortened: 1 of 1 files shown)


viviancan (approved):
This is a nice way to handle this issue! Especially since we have a working version for the Palestine region.

The only thing I noticed is that state/province region does not seem to populate for other locations in Morocco if you leave it blank. I confirmed the difference on QA and local. Here is a screen recording. I used `Meknes, Fez-Meknès, Morocco` which is a location from production.
Video links below
- [QA Morocco Location](https://drive.google.com/file/d/1XoodO8JYOFVN4ohIU1Fhwpn8L4-MVF-Z/view?usp=drive_link)
- [ Local Morocco Location](https://drive.google.com/file/d/1VTPOypjY6JEQxonBDlvVnvhHNlevv_aB/view?usp=drive_link)

Issue #5285, opened 2025-01-07
When student's DOB is updated on platform, update it in Salesforce


shaun-technovation:
When a student's DOB is updated on our platform, it needs to be updated in Salesforce. Only student's DOB gets sent to Salesforce, it doesn't get sent for any other profile type.

This used to work for Mailchimp because all profile type's DOB used to be sent, but now that only student's DOB is sent it doesn't work anymore.

dboyer:
Tested on QA and confirmed with Sasha:

- Student's birthday on platform matched birthday on Salesforce
- I logged in as the student and changed her birthday
- Updated birthday listed in Salesforce
- Picked a different student and confirmed her birthday on platform and SF matched
- As an admin, I changed her birthday and the new birthday was updated in SF

PR #5286, opened 2025-01-07
Send updated student info to Salesforce

shaun-technovation:
When upserting contact info in Salesforce, students have additional info that needs to be sent (students are the only profile type that has additional contact info). The additional info is:

- parent guardian email address
- parent guardian full name
- parent guardian first name
- parent guardian last name
- DOB

But in order to send this additional info, the `profile_type` needs to be specified when calling `upsert_contact` and it wasn't. This means that parent/guardian info and DOB for students wasn't getting updated after it was initially created.

This change will address updating parent guardian contact info and DOB will get addressed in #5285 .

I also updated the Salesforce rake task so it will work properly too.

### DIFF
--- a/lib/tasks/update_salesforce_contacts.rake
+++ b/lib/tasks/update_salesforce_contacts.rake
@@ -7,9 +7,12 @@ task update_salesforce_contacts: :environment do |_, args|
       account = Account.find(account_id)
 
       if !account.is_admin?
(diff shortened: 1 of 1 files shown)


viviancan (approved):
Did not try locally, but would be good to get on QA for testing! Nice update on this 💯

PR #5295, opened 2025-01-10
Setup club ambassador profiles in Salesforce

shaun-technovation:
This will setup a club ambassador profile in Salesforce when:

- a club ambassador registers
- a club ambassador profile is added to a mentor


viviancan (approved):
Did not try locally, but look straightforward 💯

PR #5299, opened 2025-01-13
Update parent/guardian first/last name functionality

shaun-technovation:
The `parent_guardian_name` gets split up to make a first name and last name that we send to Salesforce. There was an issue with a parent/guardian name that only has one name, it would use that one name as both the first name and the last name. The update here will address that.


viviancan (approved):
LGTM 💯

Issue #5246, opened 2024-12-12
Assign a mentor to a club when they join a team


shaun-technovation:
Related to #5156, when a mentor joins a student's team, assign the mentor the the student's club.

dboyer:
Is this actually done @shaun-technovation? Or should it be in the product backlog?

viviancan:
- [ ] Student sending an invite to a mentor with no chapter and who was a previous season mentor
- [ ] Student invites a mentor not on the platform to register and join the team
- [ ] Student sends invite to mentor found on search page
- [ ] Mentor already on a team requests to be on another team

viviancan:
This is on QA

dboyer:
Tested on QA and confirmed:

- As a student, I invited a mentor from a previous season who had not yet registered for this season or selected a chapter. The mentor (/participants/1338#/) registered and selected a chapter. That chapter was listed on their admin page. They accepted the invite to join the team. When I searched the admin participants page by the club associated with the team, the mentor appeared on the list of club participants.
- As a student, I invited a mentor who was already on a chapter. The mentor (/participants/2779#/) accepted the invite. Their primary affiliation on their participant admin page still showed as the original chapter. When I filtered the participants page for members of the London Tech Club, they showed up on the list, which shows they were connected to the club after joining the team.
- As a mentor (/participants/3871#/) already in a club, I requested to join a team affiliated with a different club. The team approved the request. The club affiliation on the mentor's participant page remained the Detroit club but when I searched by the London Tech Club, the mentor appeared on the participants list after joining the team with students from the London Tech Club.
- As a student, I invited a new mentor to register on the platform and join my team. Registered as that mentor (/participants/5254#/) and selected "none of the above" on the chapterable selection screen. Finished onboarding and accepted the team invite. After accepting the invite, the participant admin page still said "not assigned to club or chapter." When I searched the participants page by the club name (Detroit Tech Club), the mentor was on the participants list. To me, that means their primary affiliation is still "none" but the secondary affiliation is Detroit Tech Club. 

@viviancan We had talked with VET about the following scenario: "If a mentor doesn't have a primary affiliation and joins a team, the chapter for that team will become their primary affiliation. They will no longer be on the unaffiliated list. You will be able to see the chapter name on their participant admin page in the chapter field." I can't remember if that still needs to be implemented or if the current behavior is correct.

viviancan:
[...]

PR #5308, opened 2025-01-15
Assign mentor to club when joining team

viviancan:
Refs #5246 

This PR updates the mentor invite/join team flow so they can also be assigned to a club

### DIFF
--- a/app/models/chapter.rb
+++ b/app/models/chapter.rb
@@ -10,10 +10,10 @@ class Chapter < ActiveRecord::Base
   has_one :legal_contact, dependent: :destroy
   has_one :chapter_program_information, dependent: :destroy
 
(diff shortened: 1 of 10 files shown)


app/models/chapter.rb
```diff
@@ -10,10 +10,10 @@ class Chapter < ActiveRecord::Base
   has_one :legal_contact, dependent: :destroy
   has_one :chapter_program_information, dependent: :destroy
 
-  has_many :chapter_account_assignments, as: :chapterable, class_name: "ChapterableAccountAssignment"
-  has_many :accounts, through: :chapter_account_assignments
+  has_many :chapterable_account_assignments, as: :chapterable, class_name: "ChapterableAccountAssignment"
```
viviancan, line 13: Not sure if this was (or is going to be) updated in the refactor work. I updated this association to match the chapterable account assignment association in the club model

Issue #5309, opened 2025-01-16
Club ambassador testing feedback


dboyer:
All these issues were found while testing on preview:

- [x] Searching by club name on the club ambassador data grid (/admin/club_ambassadors#/) results in the error page. This is true if it is the only filter used or if it is combined with other filters.

- [x] When exporting from the club ambassador admin page (/admin/club_ambassadors#/), the club name column is a combination of the person's first and last name (not the club name) unless the person is not assigned to a club in which case it says "no club"

- [x] Add country as an additional column and filter on the clubs admin page (/admin/clubs#/)

viviancan:
This is on QA

dboyer:
Tested on QA and confirmed:

- I can search by club name on the club ambassador datagrid
- In the exported data, first and last name and club name are appearing in the correct columns
- I can add country as a column on the clubs admin page and the column is included in the export

dboyer:
Confirming on production:

- I can search by club name on the club ambassador grid
- Data exported as expected
- I can add country as a column on the clubs admin page

PR #5314, opened 2025-01-17
Club ambassador testing feedback updates

viviancan:
Refs #5309 

This PR addresses the following

- Searching by club name on the club ambassador data grid (/admin/club_ambassadors#/) results in the error page. This is true if it is the only filter used or if it is combined with other filters.
-  When exporting from the club ambassador admin page (/admin/club_ambassadors#/), the club name column is a combination of the person's first and last name (not the club name) unless the person is not assigned to a club in which case it says "no club"
- Add country as an additional column and filter on the clubs admin page (/admin/clubs#/)

I updated the club names search with the same functionality as the chapter ambassador name search so that partial matches and accented names would still be returned. Also noting that this may need to be updated with the refactoring work.

### DIFF
--- a/app/data_grids/club_ambassadors_grid.rb
+++ b/app/data_grids/club_ambassadors_grid.rb
@@ -94,10 +94,12 @@ class ClubAmbassadorsGrid
     end
 
   filter :club_name do |value, scope|
+    processed_value = I18n.transliterate(value.strip.downcase).gsub(/['\s]+/, "%")
     scope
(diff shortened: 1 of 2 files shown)


viviancan:
> Not sure if I am supposed to review this since I'm not assigned to it! 😊 But anyway, LGTM! 👍 😊

I totally forgot to tag you, sry about that! 
Ty 💯 💯

Issue #5291, opened 2025-01-08
Support Ticket: 1/8/25 - 2/5/25


viviancan:

viviancan:
- [ ] Investigate potential bg check issue - [More details here](https://iridescentlearning.slack.com/archives/C06CHBQ4E/p1736368471434659) (This is on hold due to Checkr being down. Checkr status can [be viewed here](https://checkrstatus.com/)

dboyer:
A mentor reported difficulty adding her location in Kosovo. [Full details here](https://iridescentlearning.slack.com/archives/C06CHBQ4E/p1734443100752169).

We've had a few reports [over the years](https://iridescentlearning.slack.com/archives/C06CHBQ4E/p1669159653905799) that people cannot add locations in Kosovo. What's odd is two things:

- A student was able to get the platform to accept her location as Pristina with that name in both the city and country fields. I was not able to replicate that on QA.
- On QA, I was able to get a suggested location of Viti, Kosovo, but when I tried to select it, nothing happened.

<img width="315" alt="Image" src="https://github.com/user-attachments/assets/af4a5fb9-73bd-41c5-a7c8-63a9cc6d05e9" />

Could someone look into the database to see how that location is saved? Similar to what we did for the Morocco ticket. I'm not sure if we want to add an entry but I'd like to know why some locations in Kosovo are appearing but not saving in all circumstances.

viviancan:
I will investigate Kosovo issue above ^^

viviancan:
https://en.wikipedia.org/wiki/International_recognition_of_Kosovo

viviancan:
Suppor request about the "find a missing participant" pop up-[ thread here](https://iridescentlearning.slack.com/archives/C06CHBQ4E/p1737489307059589)

viviancan:
request from ChA about team missing from teams g[rid - thread here](https://iridescentlearning.slack.com/archives/C06CHBQ4E/p1737490231315839)

dboyer:
On the submissions page for students in the technical additions section, we need to update the text.

- [ ] Remove this text

[...]

PR #5317, opened 2025-01-22
Temporary solution for kosovo location issue

viviancan:
Refs #5291 

This is a temporary solution fo the Kosovo location issue. I will add more details about this location in the 5291 Support ticket


viviancan:
Closing this as we are no longer investigating this

PR #5322, opened 2025-01-22
Address Salesforce duplicate records issue

shaun-technovation:
Salesforce gets updated via scheduled jobs. A job can get scheduled for various reasons (not an inclusive list):

- when someone logs in for the first time during a new season
- when someone gets converted from a student to a mentor
- when a chapter ambassador profile is adding to a mentor
- when a club ambassador profile is adding to a mentor

If two of these actions are done back-to-back it can cause duplicate records in Salesforce.

I added some arbitrary pauses/delays in between each of these jobs that will space them out more which will hopefully help with this issue.


viviancan (approved):
Did no try locally, but lgtm!

PR #5328, opened 2025-01-24
Club ambassadors can assign participants to their club

shaun-technovation:
This will allow club ambassadors to assign participants to their club. In order to do this, I refactored the existing chapter ambassador -> chapter assignment functionality to make it more versatile and generic so club ambassadors can use it as well.

### DIFF
--- a/app/models/club_ambassador_profile.rb
+++ b/app/models/club_ambassador_profile.rb
@@ -8,6 +8,7 @@ class ClubAmbassadorProfile < ActiveRecord::Base
     foreign_key: "account_id",
     required: false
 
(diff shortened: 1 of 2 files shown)


viviancan (approved):
LGTM!! 

The only thing that should be updated is the "Assing to your chapter"/chapter references in the grid. Not sure if this is updated elsewhere?


CleanShot 2025-01-27 at 14 20 43@2x



I also could not export the datagrid - error in terminal
```
NoMethodError (undefined method `exports' not found for #<ClubAmbassadorProfile:0x00000001189cd380>

    raise NoMethodError,
    ^^^^^):

```

PR #5329, opened 2025-01-24
Refactor and add clubs to participants datagrid

shaun-technovation:
Summary of updates:

- create a new ambassador/participants_controller
  - for the show, edit, and update actions
  - ChAs and CAs will share this controller

- create a new data_grids/participants_controller
  - ChAs and CAs will share this controller

- consolidate the views so ChAs and CAs will use the same views
- for the datagrid view it will be shared for admins, ChAs, and CAs


viviancan (approved):
LGTM! Thanks for handling this refactor and setting up the foundation for all of the other datagrids 💯

PR #5341, opened 2025-01-28
Teams datagrid for clubs and refactoring

shaun-technovation:
This will add the teams datagrid for clubs and include a little bit of refactoring work.

Also noting, I removed the chapter ambassador team edit/update routes and controller actions, I didn't see that they were being used. 🤔


viviancan (approved):
LGTM 💯 We may want to double check the export - there were some instances where I couldn't export but then I couldn't accurately recreate it! Nice clean up too 🚀

PR #5350, opened 2025-02-03
Update chapterable functionality

shaun-technovation:
This will:

- add a `#chapterable` method to both chapter ambassadors and
club ambassadors, that will refer the either their chapter or club
  - this is what was causing the error on the club profile tab, good catch!
- replace `NullChapter` with `NullChapterable`
  - I don't think we need both, but if you can think of a reason, would be easy to add it back 👍


viviancan (approved):
Nice updates, did not try locally, but would be good to get on preview for testing!! 💯

PR #5360, opened 2025-02-05
Handle blank parent/guardian names

shaun-technovation:
When trying to split the parent/guardian name into first/last name, there is an error being caused if the parent/guardian name doesn't exist, this will address that by checking if a parent/guardian name is present before trying to split it up.


viviancan (approved):
LGTM 💯

PR #5370, opened 2025-02-06
Allow club ambassadors to edit their location

shaun-technovation:
A simple update here to add the link that will allow club ambassadors to edit their location.


viviancan (approved):
Works as expected 💯

PR #5372, opened 2025-02-06
Fix unaffilated list for club ambassadors not assigned to a club

shaun-technovation:
Simple fix for club ambassadors who aren't assigned to a club accessing the unaffiliated list.

Seems like an edge case since all club ambassadors should be assigned to a club; and if they're not maybe they shouldn't have access to certain things. Anywho, since it's possible now, this a simple fix for the time being.


viviancan (approved):
👍 Looks good

PR #5385, opened 2025-02-13
Remove "club ambassador" mentor type

shaun-technovation:
This should remove the "club ambassador" mentor type from all areas. Just noting that I think this covers everything, it was tricky to search the code base for "club ambassador" or "club_ambassador" since we're using that profile type now... 🧐


viviancan (approved):
LGTM! I also checked out some areas where I remember club ambassador type being and didn't see any references to it. 💯

PR #5399, opened 2025-02-21
Ensure one primary chapter assignment per profile type per season

shaun-technovation:
This will add a validation to ensure that only primary chapterable assignment per profile type per season is created.

Also, a lot of the specs had to be updated because the specs were creating more than one assignment and only one primary assignment is valid now.

### DIFF
--- a/app/models/chapterable_account_assignment.rb
+++ b/app/models/chapterable_account_assignment.rb
@@ -3,6 +3,8 @@ class ChapterableAccountAssignment < ApplicationRecord
   belongs_to :chapterable, polymorphic: true
   belongs_to :profile, polymorphic: true
 
+  validate :ensure_one_primary_assignment_per_profile_type_per_season, if: :new_record?
+
   scope :current, -> {
     where(season: Season.current.year)
   }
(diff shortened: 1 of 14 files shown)


app/models/chapterable_account_assignment.rb
```diff
@@ -14,4 +16,20 @@ class ChapterableAccountAssignment < ApplicationRecord
   scope :clubs, -> {
     where(chapterable_type: "Club")
   }
+
+  private
+
+  def ensure_one_primary_assignment_per_profile_type_per_season
+    if primary && ChapterableAccountAssignment
+        .where(
+          account_id: account_id,
```
viviancan, line 25: From my understanding, we only want 1 primary assignment, but right now I think this checks all other chapterable assignments. Should `primary: true` be included?

I tried assigning a mentor through the admin portal that had a secondary assignment to another chapter and it failed.

app/models/chapterable_account_assignment.rb
```diff
@@ -14,4 +16,20 @@ class ChapterableAccountAssignment < ApplicationRecord
...
+          profile_type: profile_type,
+          profile_id: profile_id,
+          season: season
+        )
+        .exists?
+
+      errors.add(:base, "A participant can only be assigned to one chapter or club")
```
viviancan, line 32: Not a blocker, but I think the error message could be more specific like `A participant can only have one primary assignment`

viviancan (approved):
I just left one main comment (I think the where clause may need to be updated)! Thanks for updating all of the tests!

PR #5402, opened 2025-02-24
Add rake task to remove duplicate chapterable assignments

shaun-technovation:
A pretty simple rake task will take a list of chapterable assignment ids and remove them.

We can use the CSV that you posted in Slack, or we can pull another dataclip to get the ids we want to remove. 👍 

To run it:

```
bundle exec rails remove_duplicate_chapterable_assignments[idx,idy,idz]
```


viviancan (approved):
Tried locally with QA data and works as expected 💯

Issue #5336, opened 2025-01-24
Add recusals section to scores admin page


dboyer:
Builds on the work done in #5335 

On the scores admin page (/admin/scores#/), add a recusals section that lists all submissions with 5 or more recusals. It should list the team name, submission, number of recusals, and a link to the scores page for that submission (i.e. /admin/score_details/794#/).

dboyer:
Draft idea:

This could be a new column on scores admin page (/admin/scores#/) and have a filter for "Removed from judging pool." Does not need to be part of suspicious scores section.

viviancan:
I added this as a partial to the scores page, but will touch base with you both to see how to best move forward. I think adding it to the datagrid would be good so there are still all of the filtering options etc. I think it depends on how score monitors use the datagrids!

dboyer:
Update from VET:

- Column and filter for "removed from judging pool" are useful and admins can then click through to the scores page for that submission to see the recusals 
- If there was a section added to the top, it would only be helpful for submissions removed from the judging pool. A recused submissions section will get as crowded as the current suspicious scores section which we are trying to clear up.

dboyer:
VET tested and this is working as expected. They are fine to have it move to QA.

viviancan:
This is on QA

dboyer:
Confirming that I now see a recusals section on the scores admin page and that once a submission had five recusals it was added to this section.

The removed from judging filter also worked as expected.

dboyer:
Confirmed this section is now on production

Issue #5335, opened 2025-01-24
Submissions with five or more judge recusals should not be given to additional judges


dboyer:
If a submission has received five or more recusals, it should not be given to additional judges. Instead, it should be listed in a "Recusals" section at the top of the scores admin page (/admin/scores#/). 

This issue covers adding the code to have five recusals trigger a submission being removed from the judging pool.

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/gh/Iridescent-CM/technovation-app/5335?planning-poker) with Zenhub @shaun-technovation @viviancan

dboyer:
Tested on preview and confirmed that a submission was removed from the judging pool if it had more than five recusals. If I returned it to the judging pool, it immediately appeared as an option for a judge.

dboyer:
VET tested and this is working as expected. They are fine to have it move to QA.

viviancan:
This is now on QA

dboyer:
Tested on QA and confirmed that a submission with five recusals was moved out of the judging pool. I returned it to the judging pool and it was immediately given to the next judge.

PR #5403, opened 2025-02-25
Remove submissions with more than 4 recusals

viviancan:
Refs #5335 and #5336 

This PR adds functionality to automatically remove a submission from the judging pool on the 5th recusal.

Key Changes:
- Added a new `remove_from_judging_pool` boolean field to team_submissions
- Updated the logic to set `remove_from_judging_pool` to true when `judge_recusal_count exceeds` 4 recusals
- Added scopes
- Updated eligible submissions so that only submissions that are not removed from the judging pool are eligible for judging
- Added a "removed submissions" partial on the scores page (similar to suspicious scores)
- Added removed from judging filter & column to scored submissions datagrid  
- Added tests

### DIFF
--- a/app/controllers/admin/scores_controller.rb
+++ b/app/controllers/admin/scores_controller.rb
@@ -10,7 +10,7 @@ module Admin
       unless request.xhr?
         @round = grid_params[:round]
       end
-      @removed_submissions = TeamSubmission.current.complete.removed_from_judging_pool
+      @submissions_removed_from_judging_pool = TeamSubmission.current.complete.removed_from_judging_pool
     }, only: :index
 
     def show
--- a/app/controllers/admin/scores_controller.rb
+++ b/app/controllers/admin/scores_controller.rb
@@ -10,6 +10,7 @@ module Admin
       unless request.xhr?
         @round = grid_params[:round]
       end
+      @removed_submissions = TeamSubmission.current.complete.removed_from_judging_pool
     }, only: :index
 
     def show
--- a/app/data_grids/scored_submissions_grid.rb
+++ b/app/data_grids/scored_submissions_grid.rb
@@ -84,7 +84,7 @@ class ScoredSubmissionsGrid
 
   column :judge_recusal_count, header: "Recusals", mandatory: true, order: true
 
-  column :recused, header: "Removed from judging pool", mandatory: true do |submission|
+  column :removed_from_judging_pool, header: "Removed from judging pool", mandatory: true do |submission|
     ApplicationController.helpers.humanize_boolean(submission.removed_from_judging_pool)
   end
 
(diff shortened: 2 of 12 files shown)


app/data_grids/scored_submissions_grid.rb
```diff
@@ -7,7 +7,7 @@ class ScoredSubmissionsGrid
 
   scope do
     TeamSubmission.complete.current
-      .includes(:team, {submission_scores: :judge_profile})
+      .includes(team: :division, submission_scores: :judge_profile)
```
viviancan, line 10: Fixed bullet warning

app/models/submission_score.rb
```diff
@@ -565,4 +567,8 @@ def pending_approval?
   def update_team_score_summaries
     team_submission.update_score_summaries
   end
+
+  def update_recused
+    team_submission.reload.update(recused: team_submission.judge_recusal_count > 5)
```
viviancan, line 572: I would like to confirm expected behavior, because in the current flow, if a submission gets a 6th recusal, it will automatically be removed from the judging pool. If an admin reviews it and puts it back into the judging pool, and then it is recused for the 7th time it will be pulled out of the judging pool again. I'm not sure if that is the expected behavior or once it is reviewed it doesn't matter how many recusals the submission has (it should just always be in the judging pool).

app/views/admin/scores/_recused_submissions.html.erb
```diff
@@ -0,0 +1,29 @@
+<div id="submissions-flagged-for-review" class="panel">
+  <h4>Recused Submissions</h4>
```
viviancan, line 2: This is the partial, but waiting on confirmation if score monitoring wants a separate panel here.

app/models/team_submission.rb
```diff
@@ -128,6 +128,9 @@ def developed_on?(platform_name)
   scope :complete, -> { where("published_at IS NOT NULL") }
   scope :incomplete, -> { where("published_at IS NULL") }
 
+  scope :removed_from_judging_pool, -> { where(removed_from_judging_pool: true) }
+  scope :not_removed_from_judging_pool, -> { where(removed_from_judging_pool: false) }
```
viviancan, line 132: Unsure of the scope names. I think `removed_from_judging_pool` is pretty straight forward but I'm wondering if the other should be `eligble_for_judging` or something more general like that?

app/controllers/admin/scores_controller.rb
```diff
@@ -10,6 +10,7 @@ class ScoresController < AdminController
       unless request.xhr?
         @round = grid_params[:round]
       end
+      @removed_submissions = TeamSubmission.current.complete.removed_from_judging_pool
```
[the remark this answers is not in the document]
viviancan: Will update! thx for the suggestion!

Issue #4798, opened 2024-06-20
Remove submission type question


dboyer:
Starting in the 2025 season, there will only be one track - not separate tracks for mobile and AI. I confirmed with Karen that we should remove the submission type question entirely and have the coding language be the only question on this page. We can keep "Submission Type" as the header of the page and the entry in the left sidebar though.

<img width="1008" alt="Screen Shot 2024-06-20 at 3 03 09 PM" src="https://github.com/Iridescent-CM/technovation-app/assets/516764/c080fe9e-2ae8-46cb-938c-2790dce23884">

dboyer:
Will need to consider other places submission type is referenced - judging, project page, etc

Will need to remove any validation on the submission type field.

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/Iridescent-CM/technovation-app/4798?planning-poker) with Zenhub @shaun-technovation @viviancan

dboyer:
@shaun-technovation Karen confirmed we can change the title of this section to "Coding Language."

Help text can say "Please indicate which coding language you used."

dboyer:
Confirming that on QA I no longer see the submission type question.

PR #5405, opened 2025-02-26
Update judging questions for technical section for 2025 season

shaun-technovation:
The technical section questions that are displayed when judging a submission are dynamic based on the submission type, but we removed the submission type in #4798.

The changes here will:

- Update the technical questions
- Associate them to a blank submission type so they will get displayed


viviancan (approved):
LGTM! Technical questions loaded on my end. Nice update

PR #5415, opened 2025-03-05
RPE datagrid updates

shaun-technovation:
- Add judge name filter and column
- Add team name filter and column
- MVP for allowin gchapter ambassadors to access the events datagrid


viviancan (approved):
Nice addition and groundwork for future RPE updates! I was able to see the list as a ChA!

PR #5418, opened 2025-03-06
Fix club ambassadors viewing scores error

shaun-technovation:
Club ambassadors viewing score details was causing an error because the scores show route and controller wasn't setup, this should fix that.


viviancan (approved):
Did not try locally, but lgtm! 💯

PR #5422, opened 2025-03-07
Identify finalists, grand prize winners, and regional winners on platform

shaun-technovation:
This will lay the groundwork to be able to identify finalists, grand prize winners, and regional winners on the platform.


viviancan (approved):
LGTM!! Tried a few things locally and it worked as expected 💯

Issue #5414, opened 2025-03-05
Create rake task to bulk delete accounts


dboyer:
This task will allow bulk deleting accounts

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/gh/Iridescent-CM/technovation-app/5414?planning-poker) with Zenhub @shaun-technovation @viviancan

viviancan:
I created 10 mentor accounts and deleted them with the rake task

`heroku run bundle exec rails bulk_delete_accounts_by_id[5857,5856,5855,5854,5853,5852,5851,5850,5849,5848] --app technovation-qa`

<img width="2000" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMzlLQ1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--680fb30a625c88067cb168060d4f86a4e5d30bd6/Screenshot%202025-03-17%20at%201.03.28%E2%80%AFPM.png" alt="Screenshot 2025-03-17 at 1.03.28 PM.png" />

viviancan:
@dboyer Let me know if you also want to create some test accounts for me to delete!

dboyer:
I think we're good if you ran the test accounts already and the deletion worked as expected!

dboyer:
Will check with Sasha on how to bulk remove from Salesforce

dboyer:
Sasha confirmed there is a bulk mass delete option from Salesforce. We should provide her with a list of the deleted accounts (name and email) prior to deleting them.

viviancan:
[Here is the list ](https://docs.google.com/spreadsheets/d/1k9FRxiaH_vyKqtpxKapjSq6XZmKvtEo4mDN_d3Vje14/edit?usp=sharing)of accounts to be deleted

PR #5423, opened 2025-03-11
Added rake task to bulk delete accounts

viviancan:
Refs #5414 

Added rake task to bulk delete accounts

to run: `bundle exec rails bulk_delete_accounts_by_id[idx,idy,idz]`

### DIFF
--- a/lib/tasks/bulk_delete_accounts.rake
+++ b/lib/tasks/bulk_delete_accounts.rake
@@ -1,0 +1,10 @@
+desc "Bulk delete accounts by id"
+task bulk_delete_accounts_by_id: :environment do |task, args|
+  accounts = Account.where(id: [args.extras])
+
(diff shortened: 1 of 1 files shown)


lib/tasks/bulk_delete_accounts.rake
```diff
@@ -0,0 +1,10 @@
+desc "Bulk delete accounts by id"
+task bulk_delete_accounts_by_id: :environment do |task, args|
+  accounts = Account.where(id: [args.extras])
+
+  accounts.each do |account|
+    account.really_destroy!
```
viviancan, line 6: I had to use the `really_destroy` due to the paranoia gem ([See docs here](https://github.com/rubysherpas/paranoia))

> If you wish to actually destroy an object you may call really_destroy!

PR #5424, opened 2025-03-11
Fix error when trying to export events as a chapter ambassador

shaun-technovation:
This should address an error when trying to export the events datagrid as a chapter ambassador.


viviancan (approved):
LGTM!

PR #5426, opened 2025-03-12
Restrict access viewing scores

shaun-technovation:
Before this change an ambassador could view any score on our platform by manipulating the score id in the URL.

With this change:

- If an ambassador has the "national view" they should only be able to view a score from their region

- If an ambassador doesn't have "national view" they should only be able to view scores from participants/submissions assigned to their chapter or club

### DIFF
--- a/spec/controllers/ambassador/scores_controller_spec.rb
+++ b/spec/controllers/ambassador/scores_controller_spec.rb
@@ -1,0 +1,116 @@
+require "rails_helper"
+
+RSpec.describe Ambassador::ScoresController do
+  describe "GET #show" do
+    context "as a chapter ambassador" do
+      let(:chapter_ambassador) do
+        FactoryBot.create(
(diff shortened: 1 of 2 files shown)


viviancan (approved):
Nice update on this 💯

spec/controllers/ambassador/scores_controller_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+          it "raises an 'ActiveRecord::RecordNotFound' error" do
+            expect {
+              get :show, params: {
+                id: brazil_score.id
+              }
+            }.to raise_error(ActiveRecord::RecordNotFound)
+          end
+        end
+
+        context "when viewing a participant that is not in the chapter ambassador's region" do
```
viviancan, line 102: This context was slightly confusing to me! Should it be when viewing a score?

PR #5432, opened 2025-03-14
Ensure mentors are assigned to unique chapterable assignments when joining a team

shaun-technovation:
This should ensure that when a mentor joins a team (either through an invite or a join request) that they don't get reassigned to the same chapter or club that they already belong to, creating duplicate chapter assignment records.


viviancan (approved):
Did not try locally, but looks pretty straight forward! thanks for fixing this shaun 💯

PR #5436, opened 2025-03-14
Update rack gem

shaun-technovation:
This will update the rack gem that should address these security vulnerabilities:

- https://github.com/Iridescent-CM/technovation-app/security/dependabot/246
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/245
- https://github.com/Iridescent-CM/technovation-app/security/dependabot/241


viviancan (approved):
👍

PR #5438, opened 2025-03-14
Install Brakeman security scanner

shaun-technovation:
This will install the Brakeman security scanner into the development group in our Gemfile.

This can be run locally with the `brakeman` command:
https://github.com/presidentbeef/brakeman?tab=readme-ov-file#basic-options

I ran it locally and put the report of all the "High" security items in a Zenhub issue, it's pinned in the Icebox column.


viviancan (approved):
Thanks for adding this in and documenting issues!

PR #5448, opened 2025-03-18
Fix unathenticated ambassadors trying to access pages

shaun-technovation:
If an ambassador tries to access certain pages without logging in, it will cause an error.

For example, trying to access this page:

- localhost:3000/chapter_ambassador/participants

will cause this error `NoMethodError: undefined method 'chapter_ambassador?' for nil:NilClass`

The changes here should address that.

I also removed the custom error message, "You must be signed in as a XYZ to go there!" because we don't want unnecessary details being leaked in our error messages.

### DIFF
--- a/spec/controllers/ambassador/scores_controller_spec.rb
+++ b/spec/controllers/ambassador/scores_controller_spec.rb
@@ -99,7 +99,7 @@ RSpec.describe Ambassador::ScoresController do
           end
         end
 
-        context "when viewing a participant that is not in the chapter ambassador's region" do
+        context "when viewing a score for a team that is not in the chapter ambassador's region" do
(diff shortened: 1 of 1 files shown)


viviancan (approved):
Nice fix update with this! and a +1 to addressing things as they come in 💯

spec/features/authentication_spec.rb
```diff
@@ -34,7 +34,7 @@ def reload_cookie_names
       visit send(:"#{scope}_dashboard_path")
 
       expect(page).to have_current_path(signin_path)
-      expect(page).to have_css(".flash", text: "You must be signed in as #{scope.indefinitize.humanize.downcase} to go there!")
+      expect(page).to have_css(".flash", text: "You must be signed in to go there!")
```
viviancan, line 37: Good idea!

PR #5462, opened 2025-03-21
Create initial dependabot.yml config

shaun-technovation:
If I'm understanding the documentation correctly, this will allow Dependabot to automatically check for and open PRs for NPM security vulnerabilities:

- https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference

Maybe we could look at setting this to only look at production dependencies too:

- https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#dependency-type-allow

Also, it doesn't seem like there's a way to set this up for Ruby on Rails vulnerabilities.


viviancan (approved):
👍 LGTM!

PR #5476, opened 2025-03-25
Ensure mentors are assigned to unique chapterables when joining a team

shaun-technovation:
I extracted the chapterable assignment functionality out the controllers and moved it into a new service. This new service should only make a chapterable assignment for the mentor if it doesn't exist yet.


viviancan (approved):
Nice cleanup extracting to a service 💯 I tried a few things locally and it seems to be working as expected 👍

PR #5482, opened 2025-03-28
Restrict ambassadors viewing team and participant details

shaun-technovation:
The main change include two helper methods:

- `ambassador_can_view_participant_details?`
- `ambassador_can_view_team_details?`

These helper methods will be used when ambassadors are viewing a team's participants, and when viewing a mentor's teams.

There will still be more things to lock down and more issues to address related to national view (that we talked about on Friday) that aren't included w/ this PR, but this a start to addressing ambassadors clicking things they don't have access to.


viviancan (approved):
LGTM 💯

PR #5495, opened 2025-04-03
Remove suspicious scores section

shaun-technovation:
This will remove the suspicious scores section and the supporting code, I think I got all of it...


viviancan (approved):
LGTM! 💯 I didn't see suspicous scores on the scores page and I think you found all refs to it 👍

PR #5507, opened 2025-04-10
Add "judging location" to judge's "program participant" record in Salesforce

shaun-technovation:
This will add "judging location" when updating a judge's "program participant" record in Salesforce.

If a judge is attending an RPE, the value will be "RPE (In-person)" and if not (if they are removed from an event) it will be "Platform (Virtual)".

From our dev team conversations, we just want to be able to identify RPE judges and this should let us do that; by default everyone is a virtual judge.


viviancan (approved):
Nice solution, given the constraints (re: model)! I did not try locally since I encounter issues when testing CRM locally, but it would be good to get on QA for more eyes 💯

PR #5513, opened 2025-04-15
Basic chapter and club stats

shaun-technovation:
This will add a basic stats section for chapters and clubs, right now it will only display the total number of students and total number of mentors for the current season.

Also, noting that the number aren't matching up, not sure what to do about it at the moment. An admin can make a chapterable assignment for a participant (and that would get counted for the student/mentor totals), but they wouldn't show up under Participants b/c they wouldn't be registered to the current season (the current season isn't added to `account.seasons`). I'm thinking maybe that Participants datagrid might need to get updated.


viviancan (approved):
Nice start to all of this! 💯

PR #5536, opened 2025-04-25
Display message if a judge completes a score too quickly

shaun-technovation:
This will make it so that if a judge reaches the Review Score section (and the have a completed score) in less than ten minutes, an alert will be displayed.

### DIFF
--- a/app/javascript/judge/scores/store/mutations.js
+++ b/app/javascript/judge/scores/store/mutations.js
@@ -1,37 +1,36 @@
-import { airbrake } from 'utilities/utilities'
+import { airbrake } from "utilities/utilities";
 
 export const setComment = (state, commentData) => {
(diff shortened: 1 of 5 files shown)


app/javascript/judge/scores/store/mutations.js
```diff
@@ -67,72 +66,73 @@ export const updateScores = (state, qData) => {
...
-  }
+    name: "",
+    location: "",
+    division: "",
+    photo: "",
+  };
   state.score = {
     id: null,
     incomplete: null,
+    // started_at: null,
```
viviancan, line 105: Is this supposed to be commented out?

viviancan (approved):
LGTM! I tested the main path and the popup did appear when I went through the submission too quickly.

PR #5542, opened 2025-04-28
Add rake task to bulk assign judges to an RPE

shaun-technovation:
This rake task will use a list of judges email addresses and assign them to an RPE.

Here's how it can be used:

```
bundle exec rails assign_judges_to_rpe[judge1@example.com,judge2@example.com,RPE_ID]
```


viviancan (approved):
LGTM 💯 Thx for tackling this so quickly!

PR #5543, opened 2025-04-28
Add rake task to bulk assign teams to an RPE

shaun-technovation:
This rake task will take a list of team IDs and assign the teams to an RPE.

A team must meet this criteria to be assigned to an RPE:

- the team must exist in our system
- the team must be a part of the current season
- the team must have a submission
- the team must not be assigned to the event already

Here's how it can be used:

```
bundle exec rails assign_teams_to_rpe[team_id_1,team_id_2,rpe_id]
```

### DIFF
--- a/lib/tasks/assign_teams_to_rpe.rake
+++ b/lib/tasks/assign_teams_to_rpe.rake
@@ -1,0 +1,23 @@
+desc "Assign teams to RPE"
+task assign_teams_to_rpe: :environment do |task, args|
+  rpe = RegionalPitchEvent.find(args.extras.last)
+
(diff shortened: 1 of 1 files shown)


lib/tasks/assign_teams_to_rpe.rake
```diff
@@ -0,0 +1,23 @@
+desc "Assign teams to RPE"
+task assign_teams_to_rpe: :environment do |task, args|
+  rpe = RegionalPitchEvent.find(args.extras.last)
+
+  puts "Assigning teams to RPE #{rpe.name} (ID #{rpe.id})"
```
viviancan, line 5: Should we check to see if the team is already assigned to another RPE? I'm unsure if a team can be part of more than 1 RPE?

PR #5544, opened 2025-04-28
Download presentation slides from RPE show page

shaun-technovation:
This will add basic functionality that will let admins and chapter ambassadors download presentation slides from the RPE show page; the presentation slides can either be downloaded individually per submission/team or downloaded in bulk per RPE.

This can also be previewed/tested on the Preview environment.


viviancan (approved):
Nice addition here. I tried on preview and it seems to be working as expected. It would be good to test with more eyes and extra data too 💯

Issue #5511, opened 2025-04-14
Add content for consent waiver to rebranded dashboard side nav and energetic container


viviancan:

viviancan:
This is on Preview!

PR #5558, opened 2025-04-30
Added mentor rebrand consent waiver + old consent waiver cleanup

viviancan:
Refs #5511 
This PR updates the following:
- Adds the consent waiver route/view for the mentor rebrand onboarding step
- Adds the consent waiver side nav link
- Removed the old branding consent waiver (we had the rebrand only for judges)
- Updated mentor and judge to use rebranded version of consent waiver
- Updated tests
- Removed consent waiver `show` action. I could not find any reference to this action


app/controllers/concerns/consent_waiver_controller.rb
```diff
@@ -1,10 +1,6 @@
 module ConsentWaiverController
   extend ActiveSupport::Concern
 
-  def show
-    @consent_waiver = ConsentWaiver.find(params.fetch(:id))
```
viviancan, line 5: I could not find anywhere this was used 🤔

app/views/consent_waivers/_judge_text.en.html.erb
viviancan, line 1: This was the rebranded version of the consent waiver that was used for judges. Now, since both profiles are rebranded, I removed the "fork" in the road and just updated the main text with the correct branding

viviancan:
> I saw some inconsistent quotes, but overall LGTM! 👍 :shipit:

TY, I fixed the ones I saw!

PR #5562, opened 2025-05-01
Update functionality to bulk download RPE pitch presentations

shaun-technovation:
A few tweaks to the functionality that will allow admins and ambassadors to download pitch presentations:

- The bulk download button will only be displayed if there are presentations to download
- Only bulk download for submissions that have a pitch presentation
- Added some spacing around the download buttons
- Updated the filename being downloaded, it was replacing the period in the file extension with an underscore, e.g. .pdf would become _pdf


viviancan (approved):
LGTM!

PR #5579, opened 2025-05-08
Fix judge names on events datagrid

shaun-technovation:
Adding the "judge names" column to the events datagrid can cause an error when there are judges who are assigned to an event via an invite but they aren't registered to the platform yet.

This will make it so that if a judge is registered we'll display their name and if they're not registered yet we'll display "Invited Judge" (which is how we refer to them on the events show page).


viviancan (approved):
Did not try locally, but lgtm 🎉

PR #5586, opened 2025-05-13
Restrict club ambassadors from accessing judge mode

shaun-technovation:
This is a quick fix to make the `ENABLE_CHAPTER_AMBASSADOR_SWITCH_TO_JUDGE` config setting apply to both chapter ambassadors and club ambassadors, so that way we can prevent club ambassadors from judging.


viviancan (approved):
Did not try locally, but looks straightforward! 💯 Nice fix

PR #5591, opened 2025-05-15
Fix club ambassadors not being able to edit their profile

shaun-technovation:
If a club ambassador tries to edit their profile it will cause an error, here's a simple update that will address/fix the issue.


viviancan (approved):
Nice catch, looks good to me!

PR #5597, opened 2025-05-19
Background check exemptions and mentor searchability

shaun-technovation:
If an admin grants or revokes a background check exemption for an ambassador that doesn't have a mentor profile an error will be displayed.

The change here will address that by moving the `mentor_profile.enable_searchability_with_save` call to an `after_update` hook.


viviancan (approved):
LGTM 💯

PR #5625, opened 2025-06-06
Chapter selection updates

shaun-technovation:
This will make a few updates to the chapter selection flow:

- display the currently selected chapter/club at the top of the chapter selection page
- remove the currently selected chapter/club from the ones you can pick from

This will also add an "update your chapter/club" button for students on their "my profile" section.


viviancan (approved):
While testing the “Change your chapter” button in this PR, I could not update my assignment. I selected a new chapter or club and clicked “Save” but the assignment did not change.

When I chose “None,” the database flag `no_chapterable_selected` updated correctly, but the previous assignment remained. My profile still shows the old assignment and now I was added to the unaffiliated list.

The post-save redirect also goes to the student dashboard, which feels unexpected. It might be more intuitive to return to the profile tab or show a success message.

Great work on this feature 💯

PR #5638, opened 2025-06-11
Add newly created chapter and clubs to current season

shaun-technovation:
This will make it so that a newly created chapter or club will get assigned to the current season.


viviancan (approved):
LGTM

PR #5642, opened 2025-06-12
Remove inactive chapters and clubs from chapterable selection process

shaun-technovation:
This will remove inactive chapters and clubs (chapters and clubs not participating in the current season) from the chapter selection process.


viviancan:
@shaun-technovation Can you confirm base branch? Not sure if it needs to be rebased!

viviancan (approved):
LGTM! Tried locally and it seems to be working as expected

PR #5644, opened 2025-06-13
Mentors and judges must agree to Code of Conduct each season

shaun-technovation:
This change will make it so mentors and judges have to agree to the Code of Conduct via two new checkboxes on the Consent Waiver form.


viviancan (approved):
LGTM and works as expected 💯

PR #5659, opened 2025-06-24
Activate chapters and clubs for new season

shaun-technovation:
This will add two new services to activate chapters and clubs for a new season. They can be called from the rails console, they're also included in the season reset rake task:

- `SeasonRollover::ClubActivator.new.call`
- `SeasonRollover::ChapterActivator.new.call`

### DIFF
--- a/app/services/season_setup/chapter_activator.rb
+++ b/app/services/season_setup/chapter_activator.rb
@@ -1,0 +1,23 @@
+module SeasonSetup
+  class ChapterActivator
+    def call
+      Chapter
+        .where("'#{previous_season}' = ANY(seasons)")
+        .where.not("'#{current_season}' = ANY(seasons)")
+        .joins(legal_contact: :chapter_affiliation_agreement)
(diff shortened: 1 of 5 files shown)


app/services/season_rollover/chapter_activator.rb
```diff
@@ -0,0 +1,23 @@
+module SeasonRollover
+  class ChapterActivator
+    def call
+      Chapter
+        .where("'#{previous_season}' = ANY(seasons)")
+        .where.not("'#{current_season}' = ANY(seasons)")
+        .joins(legal_contact: :chapter_affiliation_agreement)
+        .includes(legal_contact: :chapter_affiliation_agreement)
+        .where(chapter_affiliation_agreement: {status: ["signed", "off-platform"]})
```
viviancan, line 9: Do we need to account for the "expiration" of an agreement? Do we have logic that removes the signed/offplatform status if the agreement is expired?

viviancan (approved):
LGTM! I just had 1 clarifying question about the agreement expiration but overall looks good!

PR #5662, opened 2025-06-26
Force chapterable selection

shaun-technovation:
This will:

- Add a new database column `force_chapterable_selection`
  - If this is set to true for a participant, they will be forced into the chapterable selection process
- Add a service that will set the `force_chapterable_selection` flag for mentors and students
  - This will be called during the season reset rake task

### DIFF
--- a/app/services/season_setup/force_chapterable_selection_activator.rb
+++ b/app/services/season_setup/force_chapterable_selection_activator.rb
@@ -1,0 +1,11 @@
+module SeasonSetup
+  class ForceChapterableSelectionActivator
+    def call
+      Account
+        .includes(:mentor_profile, :student_profile)
+        .where.not(mentor_profiles: {id: nil})
+        .or(Account.where.not(student_profiles: {id: nil}))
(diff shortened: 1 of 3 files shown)


app/controllers/concerns/force_chapterable_selection.rb
```diff
@@ -11,9 +11,12 @@ def require_chapterable_selection
     return if request.xhr? ||
       !logged_in_and_has_profile? ||
       !valid_location? ||
-      current_account.assigned_to_chapterable? ||
-      current_account.no_chapterable_selected? ||
-      current_account.no_chapterables_available? ||
+      !current_account.force_chapterable_selection? &&
+        (
```
viviancan, line 15: Could this chunk be extracted to it's own method with a descriptive name like the others?

viviancan: Sorry Shaun! I didn't get a chance to reply since I was OOO. Sounds good to revisit! It was just an idea, I will think about a name though 👍

PR #5673, opened 2025-07-02
Reassign mentors and students to previous season's chapterable

shaun-technovation:
For students and mentors they will be assigned to last season's assignment, if:

- They have a primary assignment for last season
- They don't already have an assignment for the current season
- The chapterable is active for the current season

### DIFF
--- a/app/services/chapterable_reassigner.rb
+++ b/app/services/chapterable_reassigner.rb
@@ -1,0 +1,31 @@
+class ChapterableReassigner
+  def initialize(account:)
+    @account = account
+  end
+
+  def call
+    return if !account.is_a_mentor? && !account.is_a_student?
+    return if account.assigned_to_chapterable?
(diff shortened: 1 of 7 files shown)


app/services/chapterable_reassigner.rb
```diff
@@ -0,0 +1,30 @@
+class ChapterableReassigner
+  def initialize(account:)
+    @account = account
+  end
+
+  def call
+    return if !account.is_a_mentor? && !account.is_a_student?
+    return if account.assigned_to_chapterable?
+    return if account.last_seasons_primary_chapterable_assignment.blank?
+
```
viviancan, line 10: Should this assignment also happen only if the chapterable is "active" for the current season?

viviancan (approved):
Besides the one comment everything else lgtm 💯

Issue #5532, opened 2025-04-25
Add program info to club onboarding steps


dboyer:
Currently, clubs have two onboarding steps - public info and club location. We should add another step for program info. Clubs are not required to sign an affiliation agreement so that step is not needed.

All three steps (public info, club location, program info) must be completed before the club is considered as onboarded. This task covers adding program info as a task in the club profile and linking it to the onboarding status for the club. It does not include updating program info content for clubs.

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/gh/Iridescent-CM/technovation-app/5532?planning-poker) with Zenhub @shaun-technovation @viviancan

viviancan:
This is on Preview. Right now the complete club program info is hardcoded to true. This will need to be updated once the content is added for club program info

dboyer:
Tested on QA and confirmed that the program info is a required onboarding step for clubs

PR #5677, opened 2025-07-03
Add program info to club onboarding

viviancan:
Refs #5532 

This PR adds program info to the club onboarding flow. I set the `program_info_complete?` method to false - which will need to be updated when the club program information content is added.


viviancan:
I updated the `program_info_complete?` to true so the tests can pass. They are failing on the onboarded callback because this value was initially set to false.

Issue #5681, opened 2025-07-08
Community connections cleanup


shaun-technovation:
Part of the work in #5535 was to make the `community_connections` table polymorphic to handle both chapter ambassadors and club ambassadors. Chapter ambassadors were converted to use the new polymorphic approach.

After #5535 is completed, we need to:

- Remove the `chapter_ambassador_profile_id` column from the `community_connections` table
- Remove the rake task (that only needs to be run once) `convert_chapter_ambassador_community_conections.rake`

PR #5689, opened 2025-07-10
Club ambassador community connections

shaun-technovation:
The first change here is to make community connections polymorphic so both chapter ambassadors and club ambassadors can use them.

Next, there is a rake task to convert chapter ambassadors to use the new polymorphic approach. (#5681 is for some cleanup work.)

Both chapter ambassadors and club ambassadors will share the same controller and views (but have separate routes).

Lastly, the old chapter ambassador specific controller and views were removed. 😅


app/models/chapter_ambassador_profile.rb
```diff
@@ -43,7 +43,7 @@ class ChapterAmbassadorProfile < ActiveRecord::Base
   has_many :chapterable_assignments, as: :profile, class_name: "ChapterableAccountAssignment"
   has_many :chapter_links, dependent: :destroy
 
-  has_one :community_connection
+  has_one :community_connection, as: :ambassador
```
viviancan, line 46: Should I update the volunteer agreements to use "ambassador" too instead of profile?

viviancan (approved):
Nice updates!! Lgtm and worked as expected  💯 Excited to see all of thees updates!

Issue #5530, opened 2025-04-25
Club ambassadors can sign volunteer agreement on the platform


dboyer:
Club ambassadors should be able to sign the club ambassador agreement directly on the agreement tab of their onboarding dashboard. This will work similar to how mentors and judges sign the consent waiver on their dashboard. 

The agreement will be embedded on that tab and the ambassador will need to type their name to indicate agreement. The text for the agreement is [here](https://docs.google.com/document/d/1dpbtlNfxmyXTY68FHyyP8QlUtssaQ1WMF6Qd9iv49gg/edit?tab=t.0#heading=h.esajyaie4ljd).

<img width="1089" alt="Image" src="https://github.com/user-attachments/assets/d8a993a0-6529-4c01-a30b-eea02fb9592a" />

viviancan:
This is on preview. Some of the links in the agreement are just placeholders for now. I will be updating the env variable name and link in #5660. I tested this with [this participant ](https://technovation-preview-28bd069458f9.herokuapp.com/admin/participants/1841#/)

dboyer:
Tested on preview and confirmed:

- I logged in as a club ambassador affiliated with a club and see the blue box prompting me to complete the volunteer agreement
- I can click the button to sign the agreement and it loads in a new page
- I am able to scroll through the whole agreement and see the signature box at the bottom
- I can sign the agreement and the page reloads with the green check mark, the signature text and the date, and the text still accessible
- I log in as an admin and can see that the agreement is checked for that ambassador

Issue #5660, opened 2025-06-24
Chapter ambassadors can sign volunteer agreement on the platform


dboyer:
Currently the ChA volunteer agreement is signed via DocuSign which has caused confusion as ambassadors cannot find the link in their email or confuse it with the chapter affiliation agreement.

We should implement it as signed on the platform similar to mentors and judges. Club ambassadors will be able to do this as part of #5530.

viviancan:
This is on preview!

dboyer:
Tested on preview and confirmed:

- Logged in as a ChA and I'm prompted to sign the volunteer agreement
- I click the volunteer agreement tab and can see the volunteer agreement loaded in a scrolling window
- I can sign the agreement and the page reloads with the text of the agreement and a time stamp and my signature
- The agreement checkmark turns green
- As an admin, the volunteer agreement is marked as complete for that user on the user admin page
- As an admin, I was able to grant a volunteer agreement exemption and that step updated as completed

dboyer:
Tested on QA and confirmed:

- As a ChA, I could sign the volunteer agreement on the platform and the onboarding task updated as expected
- As an admin, I could grant an exemption and the onboarding task updated as expected

PR #5692, opened 2025-07-11
WIP Club Ambassador Volunteer Agreement 

viviancan:
Refs #5530


app/views/admin/participants/_chapter_ambassador_onboarding.html.erb
```diff
@@ -8,7 +8,7 @@
     <%= render "admin/participants/email_confirmed_status", account: chapter_ambassador.account %>
     <%= render "admin/participants/onboarding/background_check", profile: chapter_ambassador %>
     <%= render "admin/participants/onboarding/training", profile: chapter_ambassador %>
-    <%= render "admin/participants/onboarding/sign_volunteer_agreemnt", profile: chapter_ambassador %>
+    <%#= render "admin/participants/onboarding/sign_volunteer_agreement", profile: chapter_ambassador %>
```
viviancan, line 11: Commented this out until #5660 when ChA volunteer agreement moves on platform

db/migrate/20250709204951_create_volunteer_agreements.rb
```diff
@@ -0,0 +1,10 @@
+class CreateVolunteerAgreements < ActiveRecord::Migration[6.1]
```
viviancan, line 1: I didn't think it was necessary to also include a `signed at` since this record is only created when they "sign" electronically. I thought using `created_at` would work for this situation.

app/views/ambassador/volunteer_agreements/_club_ambassador_text.en.html.erb
```diff
@@ -0,0 +1,469 @@
+<div class="h-screen">
```
viviancan, line 1: Just calling out, I plan to make these URLs ENV variables in #5660.

PR #5695, opened 2025-07-11
Reset unaffilated list

shaun-technovation:
This will add a new service that reset the unaffiliated list.


viviancan (approved):
LGTM 💯  I tried it locally and the values were updated

Issue #5719, opened 2025-07-23
Update platform text for 2026 season (honor code & meet your ambassador header)


dboyer:
The following sections are ready to be updated:

- [x] [Honor Code](https://docs.google.com/document/d/1N84qk5fDLRKQVAOcarpBnhzEBjlJxzkb-Ua1fOBe4gc/edit?tab=t.0#heading=h.bcna60u94wty)
- [x] [Meet your Chapter header](https://docs.google.com/document/d/1N84qk5fDLRKQVAOcarpBnhzEBjlJxzkb-Ua1fOBe4gc/edit?tab=t.0#heading=h.f8k3a8817vhb) for students
- [x] [Meet your Chapter header](https://docs.google.com/document/d/1N84qk5fDLRKQVAOcarpBnhzEBjlJxzkb-Ua1fOBe4gc/edit?tab=t.0#heading=h.l7qcett1b5y2) for mentors

viviancan:
this is on QA

PR #5734, opened 2025-07-29
Update the 2026 honor code and meet your ambassador header

viviancan:
Refs #5719 

Updates 
- The honor code
- Meet your ambassador (now meet your club or chapter) headers


viviancan:
@shaun-technovation Here is an updated PR! Thank you

PR #5756, opened 2025-08-01
Alow participants assigned to only chapter to unselect their chapter during the chapter selection flow

shaun-technovation:
On the chapter selection screen, this will allow participants who are already assigned to the only chapter available in their area to unselect their chapter.


viviancan (approved):

viviancan:
LGTM! Thx for handling this one 💯

PR #5766, opened 2025-08-06
Only display ambassadors from the current season

shaun-technovation:
This will display chapter ambassadors and club ambassadors for the current season:

- When viewing a chapter or club in the admin area
- When an ambassador is setting their primary contact


viviancan (approved):
Nice update, lgtm 💯

Issue #5480, opened 2025-03-27
Rebrand connect with mentors page


dboyer:
Confirmed as part of #5445 that we need to keep the connect with mentors page. That page will need to be rebranded.

dboyer:
Confirming that I see the rebranded connect with mentors page on production

PR #5777, opened 2025-08-08
Rebrand connect with mentors

viviancan:
Refs #5480 

This PR rebrands the mentor "connect with mentors" view. I removed the rebrand directory and moved the rebranded versions of the partials to `mentor_searches` directory. I could not reuse the student mentor search view since it is in vue.


app/views/mentor_searches/mentor/_mentor_preview.html.erb
```diff
@@ -0,0 +1,57 @@
+<div class="shadow-lg border border-gray-300 rounded w-full mb-4 flex flex-col">
```
viviancan, line 1: This is the only partial that was rebranded/updated. The other partials were just moved to the parent directory.

PR #5781, opened 2025-08-12
Add URL for other development platform

shaun-technovation:
This will add a URL field for when "Other" is specified as the coding language.


viviancan (approved):
Did not try locally, but LGTM

PR #5784, opened 2025-08-14
Allow mentors and unaffiliated participants to update their chapterable

shaun-technovation:
This will:

- Allow mentors to update their chapterable
- Allow unaffiliated mentors and students to update their chapterable
- Combine the "location and chapterable" partial so both mentors and students can use it

### DIFF
--- a/app/views/mentor/profiles/_menu.en.html.erb
+++ b/app/views/mentor/profiles/_menu.en.html.erb
@@ -11,7 +11,9 @@
     <div class="tw-thick-rule"></div>
 
     <div class="tw-tab-content">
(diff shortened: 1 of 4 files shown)


viviancan (approved):
Everything worked as expected for a student and a mentor.

I did run into a little bit of weirdness with a mentor/ChA combo account. I'm not sure if it was just locally. If I changed to mentor mode and then tried to update my chapterable, the page would error out. I believe it was due to this line
`              <%= render "chapterable_account_assignments/#{current_account.scope_name}/chapters_found_intro_text" %>` where `scope_name` was `chapter_ambassador`. Not sure if you want to confirm locally or test on QA. Great work on this 💯

PR #5812, opened 2025-08-21
Rails 7 upgrade

shaun-technovation:
This is the initial work for the Rails 7 upgrade. The CircleCI build is passing and quite a few things work in the app, but I'm sure we'll uncover more things to address once this makes its way to Preview.

Here is an overview of the work...

**Step 1. Update the Rails gem and related gems**

I updated Rails to the latest 7.0 version. I also updated a few other gems in order for them to work with Rails 7.

I also manually updated `Gemfile.lock` to use version 1.3.4 of the `concurrent-ruby` gem in order to address this issue:

- https://stackoverflow.com/questions/79387275/getting-error-when-creating-new-rails-project-uninitialized-constant-activesupp

**Step 2. Address initial zeitwerk migration**

Documentation:

- https://guides.rubyonrails.org/v7.0/classic_to_zeitwerk_howto.html

Some changes made as per the documentation:

- Removed `config.autoloader = :classic` from `application.rb`
- Added custom acronyms
- Addressed namespace issues
- Moved some classes to their own files because with zeitwerk only one
  class can be defined per file
- Deleting unnecessary `require` calls

I also ran `$ bin/rails zeitwerk:check` and everything checks out:

> Hold on, I am eager loading the application.
> All is good!

**Step 3. Get Rails server up and running**

The server is up and running, quite a few things work, but I'm sure we'll uncover and fix issues as we go!

**Step 4. Replace Turbolinks with Turbo**

Turbolinks has been completely replaced with Turbo!

High level overview of the Turbo work:

- Integrating Turbo
- Removing Turbolinks
- Replacing Turbolink actions/hooks w/ Turbo actions/hooks

**Step 5. Update binstubs**

I updated our binstubs (in the ./bin `directory`), to do so I followed these instructions that were printed out at the command line:

- `bundle config --delete bin`    # Turn off Bundler's stub generator
- `rails app:update:bin`          # Use the new Rails executables
- `git add bin`                   # Add bin/ to source control

**Step 6. Update configuration settings**

To update our `environments/*.rb` files, I made a new Rails 7 project and compared/merged the differences so that our files are closer aligned to those from a new Rails 7 project.

More details on new configuration settings can be found here:

- https://guides.rubyonrails.org/configuring.html

**Step 7. Address tests**

Tests and testing libraries have been updated work work w/ Rails 7.

----

**Misc.**

"Content Encoding Error" was being caused by the Bullet gem, specifically the setting to display the footer, so I disabled that setting for now.

Datagrid and the "Wrong number of arguments" error, this was caused b/c `datagrid-filter` takes keyword arguments and we were passing a hash, I updated the code to pass keyword arguments and it's working now.

Relevant datagrid documentation:

- https://rubydoc.info/gems/datagrid/Datagrid/FormBuilder

Updating our cookie consent banner to v2.0, more details in their README:

- https://github.com/IMMOSQUARE/immosquare-cookies?tab=readme-ov-file#-upgrading-to-v20


viviancan (approved):
Amazing work on this Shaun 💯 I think we should get it on preview to start testing!

PR #5846, opened 2025-09-22
Convert data terms agreement from Vue to Rails

shaun-technovation:
This will:

- Convert the data terms agreement from Vue to Rails
- Remove all of the Vue code for data terms
- Change the Rails actions from edit/update to new/create
  - Since every season a participant has to agree to the data terms, i.e. a new agreement is created every season


viviancan (approved):
I think this needs to be rebased 👍 I reviewed the files related to the commit and it lgtm.

PR #5848, opened 2025-09-23
Remove duplicate records on ambassador participants datagrid

shaun-technovation:
Some big changes here to address the duplicates, and to also make the seasons filter based on the assignments table vs the accounts table. The scopes were cleaned


viviancan (approved):
LGTM 💯

PR #5851, opened 2025-09-25
Fix CircleCI

shaun-technovation:
This is a temporary, band-aid type fix to get CircleCI to pass.

I found the pinned version of Chrome we need via the University of Chicago: the fix here is to add steps to manually download and install that version to get CircleCI to pass.

Relevant Stack Overflow post:

- https://stackoverflow.com/a/78001106

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  browser-tools: circleci/browser-tools@1.4.3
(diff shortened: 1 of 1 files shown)


viviancan (approved):
Nice find 💯 Thanks for digging into! Looks good to get circle ci to pass and merge the pending PRs!

Issue #5857, opened 2025-09-30
View teams and submissions by chapter or club


dboyer:
Goal: Admins can quickly view teams and submissions filtered by chapter or club to assist with tracking metrics around chapter or club participation.

PT/VET are interested in having the following filters:

- [x] Add a chapter and club filter to /admin/teams#/
- [x] Add a chapter and club filter to /admin/team_submissions#/
- [x] Add a teams count to the season stats box on the bottom of the chapter or club admin page (i.e. /admin/chapters/273#/)

We have discussed how some teams will appear in the counts and filters for multiple clubs and chapters if they contain students from multiple clubs and chapters and PT is fine with that.

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/gh/Iridescent-CM/technovation-app/5857?planning-poker) with Zenhub @shaun-technovation @viviancan

viviancan:
This is on QA.

I added a few additional front end columns/text to help with the seasonality testing.


<img width="1596" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNzNEQ3c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--bb5146a1a4210502594cdda82c1f00e8921a0e99/Screenshot%202025-10-17%20at%201.25.19%E2%80%AFPM.png" alt="Screenshot 2025-10-17 at 1.25.19 PM.png" />


Here is a screenshot of detroit club with no season filter. I found it helpful to include the season column and student name column. 
Expected behavior 
- When you filter by a chapterable and season then only teams/submissions should show for students who are assigned during the season selected. If they were part of a different team in a different season that team should not display.
- When you filter by a chapterable and no season then all team/submissions that have a student assigned to that chapter should display. The main thing to check for is that it is specifically the team they were during the same season they were assigned to the chapterable
- If a team has student in multiple chapterables then the team should show up when you filter by each of those chapterables


### I added the season to the team page
[...]

PR #5881, opened 2025-10-10
View/filter teams and submissions by chapter or club + team and submission count on chapter/club details page

viviancan:
Refs #5857 

This PR adds the following:
- Ability for admin to filter Teams by chapter/club on teams datagrid
- Ability for admin to filter Submissions by chapter/club on submission datagrid
- Team and submission count on chapter/club detailed page view
- A few UI updates to expose more season-related data for teams and submissions

The biggest logic change is the update to the `by_chapterable` scope, which now accepts an optional `season` parameter.

This allows admin to filter teams and submissions by:
- a specific season
- or no season - Which only shows teams/submissions where their chapterable account assignment and team membership occurred in the same season

Without this parameter, filtering could return teams/submissions from other seasons for students who changed chapters over time.

Expected behavior (I'm using chapter and team as an example here)
- When you filter by a chapter and a season, then a team with a student who was part of that chapter in that season should be displayed
- When you filter by a chapter and no season, then a team is only returned when a student's chapter affiliation and team membership are the same (in the same season)

For example
- Student A on Team 1 affiliated with Chapter 1 in 2025
- Later, Student A on Team 2 affiliated with Chapter 2 in 2026

Filtering by Chapter 2 should only return Team 2. Without the updated logic, Team 1 would also be included because the scope did not check the chapter and team season matched

### DIFF
--- a/spec/controllers/ambassador/teams_controller_spec.rb
+++ b/spec/controllers/ambassador/teams_controller_spec.rb
@@ -70,7 +70,9 @@ RSpec.describe Ambassador::TeamsController do
           before do
             brazil_student.chapterable_assignments.create(
               account: brazil_student.account,
-              chapterable: brazil_chapter
+              chapterable: brazil_chapter,
+              season: Season.current.year,
+              primary: true
             )
 
             brazil_team.students << brazil_student


shaun-technovation (approved):
👍 LGTM! I know these datagrids are tricky!

PR #5900, opened 2025-10-23
Update CircleCI and specs

shaun-technovation:
A few updates here:

- Removing our pinned version of Chrome
- Installing the latest version of Chrome (and FIrefox)
- Updating our Capybara testing library
- A bunch of updates to our tests

For the tests, here's an annoying error (that we might run into more of b/c they happen very randomly, I fixed a bunch of them though):

 ```
Selenium::WebDriver::Error::UnknownError:
unknown error: unhandled inspector error: {"code":-32000,"message":"Node with given id does not belong to the document"}
(Session info: chrome=141.0.7390.122)
````

To solve this error, I used this trick by setting an expectation on the current path from this comment:

- SeleniumHQ/selenium#15401 (comment)

For other errors, I added more expectations about the content being displayed on the page.

For the most part, the tests are working as expected, the last 20+ test runs have been successful. But I have a feeling we'll still run into some flaky tests.

### DIFF
--- a/app/controllers/admin/season_schedule_settings_controller.rb
+++ b/app/controllers/admin/season_schedule_settings_controller.rb
@@ -6,7 +6,7 @@ module Admin
 
     def update
       SeasonToggles.configure(season_toggle_params)
-      redirect_to edit_admin_season_schedule_settings_path,
+      redirect_to admin_dashboard_path,
         success: "Season schedule saved!"
     end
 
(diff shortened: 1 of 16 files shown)

### CODE REVIEW

File: app/controllers/admin/season_schedule_settings_controller.rb [hunk 1]
```diff
@@ -6,7 +6,7 @@ def show
 
     def update
       SeasonToggles.configure(season_toggle_params)
-      redirect_to edit_admin_season_schedule_settings_path,
+      redirect_to admin_dashboard_path,
```
shaun-technovation: This is the only non-test, CircleCI, gem-related change.

I had to change the redirect in order for the `has_current_path` trick to work. Before it was basically redirecting back to the form and the URL wouldn't change.

When we get around to moving the admin content & setting from Vue to Rails we can update this.

Issue #5898, opened 2025-10-23
Add migration to drop divisions_regional_pitch_events table


viviancan:

dboyer:
@viviancan Need to run this on QA

viviancan:
Confirmed that this has been run on QA already

Issue #5897, opened 2025-10-23
Update RPE and division relationship + add migration to populate division id on events table


viviancan:

dboyer:
Moving this to confirmed on preview because RPE creation and editing seem to be working as expected

Issue #5890, opened 2025-10-15
Let ambassadors edit & delete events in Rails


dboyer:
Depends on #5887.

Let ambassadors edit events via the event details page (/chapter_ambassador/events/523).

dboyer:
Confirmed on QA that I could edit all event fields and delete an event. Leaving this in QA until PT completes testing.

Issue #5887, opened 2025-10-15
Move event creation to Rails


dboyer:
ChAs should be able to create an event with the following fields:

- Event name
- Division
- City
- Venue address
- Date
- Event time
- Cap on number of teams in attendance
- Event URL

PR #5901, opened 2025-10-23
Added Rails ChA Regional Pitch Events CRUD functionality + update division relationship 

viviancan:
Refs #5887, #5890, #5897 

This PR adds the following:
- ChA can create, edit, delete RPEs (in Rails)
- Updates the relationship between rpe and division from `HABTM` to `belongs_to` 
- Adds migration to populate division ids from `divisions_regional_pitch_events` to `regional_pitch_events` 
- Basic validation for team capacity limit 

I made a few vue updates just to allow the vue process to continue working since we are still building out the managing rpe functionality. All of vue related files will be deleted in another ticket. I also added #5898 to drop the `divisions_regional_pitch_events` table.

Also a SeasonToggles test is failing and I am unsure if its related to the circleci issues.

### DIFF
--- a/spec/system/chapter_ambassador/regional_pitch_event_teams_spec.rb
+++ b/spec/system/chapter_ambassador/regional_pitch_event_teams_spec.rb
@@ -1,6 +1,6 @@
 require "rails_helper"
 
-RSpec.describe "Regional Pitch Event Teams", :js do
+RSpec.xdescribe "Regional Pitch Event Teams", :js do
   let(:chapter_ambassador) { FactoryBot.create(:chapter_ambassador, :approved) }
 
   before do
--- a/spec/system/chapter_ambassador/regional_pitch_events_spec.rb
+++ b/spec/system/chapter_ambassador/regional_pitch_events_spec.rb
@@ -1,6 +1,6 @@
 require "rails_helper"
 
-RSpec.describe "Regional Pitch Events", :js do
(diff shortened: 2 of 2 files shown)


app/controllers/chapter_ambassador/regional_pitch_events_controller.rb
```diff
@@ -83,59 +66,37 @@ def update
...
 
     private
 
     def pitch_event_params
       params.require(:regional_pitch_event).permit(
         :name,
+        :event_date,
+        :start_time,
+        :end_time,
         :starts_at,
```
viviancan: These old params for vue form date/time will be removed in cleanup. So will division_ids. I left these here so the vue form also work while we complete the rest of the functionality

app/models/regional_pitch_event.rb
```diff
@@ -198,6 +210,15 @@ def at_team_capacity?
     end
   end
 
+  def capacity_enabled?
+    capacity_enabled == "1" ||
+      (capacity.present? && capacity.to_i > 0)
+  end
+
+  def clear_capacity_if_capacity_disabled
+    self.capacity = nil unless capacity_enabled?
```
shaun-technovation: Maybe it's just me, but the "if capacity disabled" and "unless capacity enabled" is a bit too much mental work for me; maybe just simplifythe naming by removing the "if capacity disabled" part. Image if all of our methods were named "if this...." as a part of their method names! 🤯

PR #5902, opened 2025-10-24
Add migration to drop `divisions_regional_pitch_events` table

viviancan:
Refs #5898 

Add migration to drop `divisions_regional_pitch_events` table


shaun-technovation (approved):
Nice! 👍

PR #5915, opened 2025-10-31
Bifurcate AI-related judging question

shaun-technovation:
This will update the "judge question" service to return a different question based on if the submission being judged used AI or not (based on the `ai_usage` question).


viviancan (approved):
LGTM 💯

Issue #5910, opened 2025-10-29
Move adding and removing a single team to an RPE from Vue to Rails


dboyer:
Based on the work completed in #5889, this will include:

- [x] Potentially using Hotwire to load the teams in a grid in the event page
- [x] Keeping the same existing logic for what teams are shown (i.e. all eligible teams in the country)
- [x] Including a field to enter a team name that will auto-filter the list of teams

The option to multi-select teams to add to the event may be covered in a different issue.

viviancan:
This is on Preview

dboyer:
Tested on QA and confirmed I could add and remove a single team. I also received an email when a team was added to an event.

Leaving this in the QA column until PT tests it.

PR #5925, opened 2025-11-06
RPE - Add, remove, search teams with turbo/hotwire + stimulus js 

viviancan:
Refs: #5910 + #5927

This PR updates the RPE team management experience by adding Turbo Streams to the
codebase for the first time. Chapter Ambassadors can now add and remove teams from regional pitch events with real-time UI updates (no JS), which will replace the vue integration. When a user clicks add teams, the available teams list is loaded. This defaults to teams in the region. One thing to note is that I used the `AddTeamToRegionalEvent` service when adding a team to the event. This is slightly different than the current vue integration which uses `CreateEventAssignment`. I used this service because I did not want to make changes to `CreateEventAssignment` since the Vue integration still uses that. I thought it would be helpful for testing/reviewing. 

The implementation uses turbo frame replacement. When a team is added or removed, the entire "Available Teams" and "Registered Teams" frames are replaced with updated event and team data. All docs and threads I reviewed stated start simple (in this case with replacement) and if performance becomes a concern with large team lists, we can optimize by switching to row-level updates (using turbo_stream.append and turbo_stream.remove for individual team rows) without changing the controller logic or user experience.

This PR also adds Stimulus for the first time with a new search controller to handle a debounced form submission. The availble teams section is also paginated using the `will_paginate` gem (already in the project). We may need to review this in the future since this gem will no longer receive updates per their github. This pagination gem is used in mentor and team search. I looked at swapping it for Pagy, but we need to upgrade ruby first.

### DIFF
--- a/spec/features/admin/view_rpe_details_spec.rb
+++ b/spec/features/admin/view_rpe_details_spec.rb
@@ -47,7 +47,7 @@ RSpec.feature "Admin view rpe details", js: true do
     end
   end
 
-  scenario "Viewing a specific RPE displays RPE judge details" do
+  xscenario "Viewing a specific RPE displays RPE judge details" do
     rpe_judges = FactoryBot.create_list(:judge, 3)
     rpe_judges.each do |judge|
       judge.regional_pitch_events << @rpe


app/controllers/chapter_ambassador/event_teams_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+      AddTeamToRegionalEvent.call(@event, @team)
+      @event.reload
+      @available_teams = Team.available_for_event(@event, current_ambassador)
+
+      respond_to do |format|
+        format.html {
+          redirect_to chapter_ambassador_event_path(@event),
+            notice: "#{@team.name} added to #{@event.name}."
+        }
+        format.turbo_stream
```
viviancan: This will render the `create.turbo_stream.erb` template

app/controllers/chapter_ambassador/event_teams_controller.rb
```diff
@@ -0,0 +1,51 @@
+module ChapterAmbassador
+  class EventTeamsController < ChapterAmbassadorController
+    def create
+      @event = RegionalPitchEvent.find(params[:event_id])
+      @team = Team.find(params[:team_id])
+
+      AddTeamToRegionalEvent.call(@event, @team)
```
viviancan: This is where I used the `AddTeamToRegionalEvent` - I do think this service needs to be updated for the email or switched out with the `CreateEventAssignment` which will also need updates. `CreateEventAssignment` also handles bulk assignments.

app/views/admin/regional_pitch_events/_available_teams.html.erb
```diff
@@ -0,0 +1,25 @@
...
+          controller: "search",
+          turbo_frame: "available-teams-list",
+          turbo_action: "advance"
+        } do |f| %>
+
+        <%=  f.text_field :query,
+          value: params[:query],
+          placeholder: "Search teams by name",
+          data: {
+            action: "input->search#search",
```
viviancan: The text_field is linked to the stimulus controller through this data-action. It pretty much means on an input action use the search controller and search method.

app/controllers/chapter_ambassador/regional_pitch_events_controller.rb
```diff
@@ -81,6 +80,22 @@ def destroy
         success: "Regional pitch event was successfully deleted."
     end
 
+    def available_teams
+      @event = RegionalPitchEvent.in_region(current_ambassador)
+        .find(params[:id])
+      @available_teams = load_available_teams_for_event(@event)
+
+      if turbo_frame_request_id == "available-teams-frame"
```
viviancan: This conditional logic checks which frame made the request. This is needed because on the initial "add teams" click the entire frame needs be loaded vs on the search only the list needs to be loaded

app/javascript/controllers/search_controller.js
```diff
@@ -0,0 +1,14 @@
...
+    this.timeout = null;
+  }
+
+  search() {
+    clearTimeout(this.timeout);
+    this.timeout = setTimeout(() => {
+      this.element.requestSubmit();
+    }, 300);
+  }
+}
```
shaun-technovation: The naming of this controller seems a little generic. It would be nice to have more structure/organization for the Stimulus controllers too. Maybe this could be nested under a regional pitch events folder or something; and the controller could reference teams... just some random thoughts/ideas.

Issue #5924, opened 2025-11-06
Make the program links optional on the public info tab of the chapter onboarding


dboyer:
Currently, program links are required for a chapter to complete the "public info" onboarding step. It should be optional to add those links. 

<img width="541" height="317" alt="Image" src="https://github.com/user-attachments/assets/7a30bb49-ea12-4164-ab56-133de597d94d" />

dboyer:
Tested on prod and confirmed that a chapter without program links (/admin/chapters/262) was now marked as onboarded.

NOTE: I did have to go into the chapter's edit admin page and click save again to get the page to reload to show the chapter as onboarded. I'm not sure if the onboarded flag will be triggered if a chapter hadn't filled out the program links prior to us deploying this update.

PR #5947, opened 2025-11-17
Make chapter links optional

viviancan:
Refs #5924 

This makes chapter program links optional


shaun-technovation (approved):
Nice, LGTM!

PR #5951, opened 2025-11-21
CSV upload to bulk-add teams to an RPE

shaun-technovation:
This PR has the basic functionality to upload a CSV to add teams to an RPE.

One thing this PR doesn't have is a success/failure message after the CSV has been processed; the CSV will get processed in a background job, but there won't be a follow-up message after it gets processed.

### DIFF
--- a/app/controllers/concerns/bulk_add_teams_to_regional_pitch_event.rb
+++ b/app/controllers/concerns/bulk_add_teams_to_regional_pitch_event.rb
@@ -16,9 +16,9 @@ module BulkAddTeamsToRegionalPitchEvent
           end.compact
 
         AssignTeamsToRegionalPitchEventJob.perform_later(regional_pitch_event_id: @event.id, team_ids:)
-
-        redirect_to chapter_ambassador_event_path(@event), success: "Your CSV file was uploaded successfully! It will be processed soon!"
       end
+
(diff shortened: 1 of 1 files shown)


app/views/admin/regional_pitch_events/_bulk_add_teams.en.html.erb
```diff
@@ -0,0 +1,23 @@
...
+  data: {turbo: false},
+  html: { method: :post }) do |form|
+%>
+
+  <p style="margin: 16px 0 24px;">Instructions here...</p>
+
+  <div>
+    <%= form.file_field :csv_file, 
+      required: true,
+      style: "width: 100%; padding: 8px; border: 1px solid oklch(87.2% 0.01 258.338); border-radius: 0.15rem; font-size: 1.2rem;"
```
viviancan: I was able to upload different file types, maybe file type validation would be good here!

shaun-technovation: I addressed this so that the file picker should only allow CSVs to be selected now! 👍

shaun-technovation:
> I left a few comments, but it worked as expected! Great job on this 🎉

Nice, thanks for the feedback! I got all of it addressed, I'm sure there will be more once it gets to Preview! Going to put on Preview now!

Issue #5945, opened 2025-11-17
Limit SMS consent form access to certain countries


dboyer:
Related to #5918 

We will provide access to SMS consent forms to certain countries as a pilot. This issue covers setting up the implementation to support a per country model. 

US (~1600), Nigeria (~3300), Brazil (~1000),  India(~1200), Ghana (~100)

dboyer:
Tested on QA and confirmed I had the option to enter a phone number if I was in one of the specified countries but not for other countries.

Issue #5921, opened 2025-11-06
Students can edit parent phone number


dboyer:
Depends on #5919. After the student adds the phone number, they should see an option to edit it. This could be similar to editing a parent's email address and could potentially automatically trigger sending a new consent form to that phone number.

<img width="736" height="139" alt="Image" src="https://github.com/user-attachments/assets/15524abd-5eca-4d12-81d1-d0e4b348a23d" />

dboyer:
Tested on QA and confirmed that I could change the parent phone number and it sent a new consent form and the student was not marked as onboarded until the parent completed the new consent form.

The media consent form did not reset but I think that's okay.

PR #5974, opened 2025-12-09
Add parental consent sms tracking, restrict sms request to certain countries, update frontend parental consent phone number partials

viviancan:
Refs #5945, #5970, #5921

This PR adds the following
- Limits sms requests to certain countries. Added `in_parental_consent_sms_country?` method similar to bg checks
- Adds `ParentalConsentRequests` table to track requests sent. I am open to naming. As discussed in standup, unsure of best naming that encompasses both sms, and whatspp
- Adds association between `StudentProfile` and `ParentalConsentRequests`
- Updated frontend to display appropriate partials depending on sms record

### DIFF
--- a/spec/services/twilio/api_client_spec.rb
+++ b/spec/services/twilio/api_client_spec.rb
@@ -36,11 +36,11 @@ RSpec.describe Twilio::ApiClient do
     allow(messages).to receive(:create).and_return(response)
   end
 
-  describe "#send_parental_consent_sms" do
-    it "sends an SMS message" do
+  describe "#send_parental_consent_text_message" do
+    it "sends a parental consent text message" do
       expect(messages).to receive(:create)
 
-      twilio_api_client.send_parental_consent_sms(student_profile: student_profile)
+      twilio_api_client.send_parental_consent_text_message(account: student_profile.account)
     end
   end
 end


app/models/parental_consent_request.rb
```diff
@@ -0,0 +1,6 @@
+class ParentalConsentRequest < ActiveRecord::Base
```
viviancan: I'm still thinking about the best naming for this. What about `ParentalConsentMessage`?

db/migrate/20251204225514_create_parental_consent_requests.rb
```diff
@@ -0,0 +1,14 @@
+class CreateParentalConsentRequests < ActiveRecord::Migration[7.0]
+  def change
+    create_table :parental_consent_requests do |t|
+      t.references :student_profile, index: true, foreign_key: true
+      t.integer :season, null: false, limit: 2
+      t.integer :delivery_method, null: false
+      t.string :external_message_id
+      t.string :recipient
```
shaun-technovation: Nice constraints! Shouldn't a recipient and student profile also be required (not null)? 🤔

app/models/student_profile.rb
```diff
@@ -383,6 +385,11 @@ def can_be_marked_onboarded?
       account.terms_agreed_at?
   end
 
+  def in_parental_consent_sms_country?
+    country_codes = ENV.fetch("PARENTAL_CONSENT_SMS_COUNTRY_CODES", "").split(",")
```
shaun-technovation: Would be nice to align this ("sms") to the standard name we're referring to this as... request, message, text message, or whatever...

app/models/parental_consent_request.rb
shaun-technovation: What about Text Message? 🤔

viviancan: I am going with "TextMessage"! 👍

PR #5982, opened 2025-12-12
Add rake task to update all contacts in Salesforce

shaun-technovation:
I ended up going with the simplest approach to update all accounts in Salesforce.

This will add a new rake task that will update 500 accounts (non-admin and in the current season) at a time in Salesforce every 15 minutes.

This rake task will leverage our existing rake task that takes a list of account ids and updates those accounts in Salesforce.


viviancan (approved):
Looks good to me 💯

PR #5985, opened 2025-12-15
Notify ambassadors of results from bulk-adding teams (via an CSV) to an RPE

shaun-technovation:
This will add basic email functionality for when ambassadors upload a CSV to bulk-add teams to an RPE; after the CSV has been processed the ambassador will receive a basic email with the results.

### DIFF
--- a/app/mailers/ambassador_mailer.rb
+++ b/app/mailers/ambassador_mailer.rb
@@ -102,6 +102,18 @@ class AmbassadorMailer < ApplicationMailer
     end
   end
 
+  def assigned_teams_to_event(account:, event:, assignment_result:)
+    @name = account.first_name
+    @event = event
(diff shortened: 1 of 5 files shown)


app/mailers/ambassador_mailer.rb
```diff
@@ -102,6 +102,18 @@ def team_joined_event(account, event, team)
     end
   end
 
+  def assigned_teams_to_event(account:, event:, assignment_result:)
+    @name = account.first_name
+    @event = event
+    @assignment_result = assignment_result
+    @event_url = chapter_ambassador_regional_pitch_events_url
```
viviancan: Should the event url be the specific event url? Right now it goes to `chapter_ambassador/regional_pitch_events#/`

Issue #5942, opened 2025-11-12
Implement Twilio webhook


dboyer:
Update the webhook with Twilio to receive the status of the message and report it somehow to the student.

PR #5986, opened 2025-12-16
Add Twilio webhook functionality to track sms delivery status

viviancan:
Refs #5942 
This PR adds Twilio webhook functionality to track sms delivery status

Changes include
- New Twilio webhook endpoint 
- Twilio webhook job
- Database updates - Added twilio fields (status, error code, error message)
- Updated admin debugging tools to show sms delivery status

Note for future self/devs - The delivery status callback url is set on the twilio account dashboard under Develop > Integration. Attaching a screenshot of the menu

Screenshot 2025-12-16 at 2 20 08 PM

Also docs on the status callback webhook [are here ](https://www.twilio.com/docs/usage/webhooks/messaging-webhooks#outbound-message-status-callback) and [here](https://www.twilio.com/docs/messaging/guides/outbound-message-status-in-status-callbacks)

### DIFF
--- a/app/views/admin/participants/_parental_consent_text_messages.erb
+++ b/app/views/admin/participants/_parental_consent_text_messages.erb
@@ -1,17 +1,17 @@
 <div class="margin--t-large">
   <p class="padding--l-large">Text Messages</p>
   <ul>
-    <% profile.account.text_messages.parental_consent.current.order(created_at: :desc).each do |request| %>
+    <% profile.account.text_messages.parental_consent.current.order(created_at: :desc).each do |text_message| %>
       <li>
         Sent At:
-        <%= request.sent_at.strftime("%B %e, %Y %l:%M %p") %>
+        <%= text_message.sent_at.strftime("%B %e, %Y %l:%M %p") %>
         -
         Status:
-        <%= request.status %>
+        <%= text_message.status %>
         <br>
-        <% if request.failed? || request.undelivered? %>
+        <% if text_message.failed? || text_message.undelivered? %>
           Error:
-          <%= request.error_message %>
+          <%= text_message.error_message %>
         <% end %>
       </li>
     <% end %>


app/controllers/webhooks/twilio_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+  def create
+    ProcessTwilioWebhookPayloadJob.perform_later(webhook_payload_id: @webhook_payload.id)
+    head :ok
+  end
+
+  private
+
+  def verify_webhook_payload
+    api_auth_token = ENV.fetch("TWILIO_AUTH_TOKEN")
+    validator = Twilio::Security::RequestValidator.new(api_auth_token)
```
viviancan: Using Twilio's built in validator
`Then, on your end, if you want to verify the authenticity of the request, you can leverage the built-in request validation method provided by all of our SDKs:`

[Docs here](https://www.twilio.com/docs/usage/security#validating-requests)

shaun-technovation: We also have our own Webhook validator that can be used (reused) too - `VerifyWebhookPayload`... My first thought/reaction is to use that...? Thoughts? Instead of having service specify validators... 🤔

app/controllers/webhooks/base_controller.rb
```diff
@@ -9,7 +9,8 @@ def verify_webhook_payload
     raise NotImplementedError, "All webhook controllers must implement this functionality"
   end
 
-  def save_webhook_payload
-    @webhook_payload ||= WebhookPayload.create!(body: request.body.read)
+  def save_webhook_payload(payload = nil)
+    body = payload || request.body.read
```
shaun-technovation: Ah! A cleaner way would be to make the `payload` argument required that way there isn't a conditional here (and no defaulting of the argument, making all of it simpler); that will be a nice improvement to our webhook functionality! 👍

shaun-technovation (approved):
Awesome to see all of this coming together! 🎉

app/views/admin/participants/_parental_consent_text_messages.erb
```diff
@@ -0,0 +1,19 @@
+<div class="margin--t-large">
+  <p class="padding--l-large">Text Messages</p>
+  <ul>
+    <% profile.account.text_messages.parental_consent.current.order(created_at: :desc).each do |request| %>
```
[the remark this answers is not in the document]
viviancan: Ahh good point. I think this is from the very first attempt I made where I was calling them "requests" like a student requested parental consent. I will update 👍

PR #5988, opened 2025-12-17
Slack notifications for 3rd-party integrations

shaun-technovation:
This will setup the Noticed gem, from the [README](https://github.com/excid3/noticed):

> Noticed is a gem that allows your application to send notifications of varying types, over various mediums, to various recipients. Be it a Slack notification to your own team when some internal event occurs or a notification to your user, sent as a text message, email, and real-time UI element in the browser, Noticed supports all of the above (at the same time)!

First, do we want this gem/functionality? 🤔 (If not, I found a stand-alone, dedicated Slack gem we can use too!)

Using the Noticed gem, the changes in this PR would send a message to Slack whenever an error occurs for one of these integrations:

- Checkr
- Docusign
- Salesforce
- Twilio

### DIFF
--- a/app/notifiers/integration_error_notifier.rb
+++ b/app/notifiers/integration_error_notifier.rb
@@ -1,0 +1,14 @@
+class IntegrationErrorNotifier < Noticed::Event
+  bulk_deliver_by :slack do |config|
+    config.url = ENV.fetch("SLACK_URL_FOR_INTEGRATION_ERROR_NOTIFIER")
+    config.json = -> {
+      {
+        text: params[:error_message]
+      }
+    }
+    config.if = -> {
+      Rails.env.production? &&
+        ENV.fetch("SLACK_URL_FOR_INTEGRATION_ERROR_NOTIFIER").present?
+    }
+  end
+end
--- a/app/services/checkr_api_client/api_client.rb
+++ b/app/services/checkr_api_client/api_client.rb
@@ -5,7 +5,8 @@ module CheckrApiClient
       client_id: ENV.fetch("CHECKR_API_KEY"),
       base_url: ENV.fetch("CHECKR_API_BASE", "https://api.checkr.com"),
       logger: Rails.logger,
-      error_notifier: Airbrake
+      error_notifier: Airbrake,
+      secondary_error_notifier: IntegrationErrorNotifier
     )
 
       @connection = Faraday.new(
@@ -17,6 +18,7 @@ module CheckrApiClient
 
       @logger = logger
       @error_notifier = error_notifier
+      @secondary_error_notifier = secondary_error_notifier
       @candidate = candidate
       @candidate_id = candidate.background_check.candidate_id
     end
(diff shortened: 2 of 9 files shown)


app/notifiers/integration_error_notifier.rb
```diff
@@ -0,0 +1,13 @@
+class IntegrationErrorNotifier < Noticed::Event
```
shaun-technovation: This is the notifier I created for integration errors. Right now it only sends notifications via Slack, but in this file is where we would define other notifiers.

app/notifiers/integration_error_notifier.rb
```diff
@@ -0,0 +1,14 @@
...
+  bulk_deliver_by :slack do |config|
+    config.url = ENV.fetch("SLACK_URL_FOR_INTEGRATION_ERROR_NOTIFIER")
+    config.json = -> {
+      {
+        text: params[:error_message]
+      }
+    }
+    config.if = -> {
+      Rails.env.production? &&
+        ENV.fetch("SLACK_URL_FOR_INTEGRATION_ERROR_NOTIFIER").present?
```
shaun-technovation: This will only send a notification if the Slack URL is setup and it's a production environment.

app/services/checkr_api_client/api_client.rb
```diff
@@ -66,17 +68,19 @@ def retrieve_invitation(invitation_id)
...
         Result.new(success?: true, payload: invitation_response_body)
       else
-        error = "[CHECKR] Error requesting invitation for #{invitation_id} - #{invitation_response_body[:error]}"
-        logger.error(error)
-        error_notifier.notify(error)
+        error_message = "[CHECKR] Error requesting invitation for #{invitation_id} - #{invitation_response_body[:error]}"
+
+        logger.error(error_message)
+        error_notifier.notify(error_message)
+        secondary_error_notifier.with(error_message:).deliver
```
shaun-technovation: 📌 This is how the new notifier is called, we pass in the `error_message`...

app/services/checkr_api_client/api_client.rb
```diff
@@ -5,7 +5,8 @@ def initialize(
       client_id: ENV.fetch("CHECKR_API_KEY"),
       base_url: ENV.fetch("CHECKR_API_BASE", "https://api.checkr.com"),
       logger: Rails.logger,
-      error_notifier: Airbrake
+      error_notifier: Airbrake,
+      secondary_error_notifier: IntegrationErrorNotifier
```
shaun-technovation: All of our API clients now have "secondary" error notifier (open to renaming that). Also, we might be able to move the Airbrake notifier into the new "integration error" notifier (then we'd only have one notifier), we'd need to write a custom notifier for Airbrake though since it's not supported by default.

viviancan (approved):
Nice addition, I think we should try it out for sure!! I think it will be super helpful and I love the flexibility of the gem!

PR #5991, opened 2025-12-22
Allow admins to void documents more than 120 days old

shaun-technovation:
This will make it so that when documents are more than 120 days old they can be marked as voided on our platform by an admin; before it would require one of us to manually mark them as voided.

Before this change if an admin tried to void one of these documents it would not be voided on our platform b/c voiding on our platform requires a successful "void" response from Docusign, but these documents weren't getting voided in Docusign b/c after 120 days the documents expire and are automatically marked as voided in Docsign; this would not return a successful response and hence they wouldn't get marked as voided on our platform.

The change here will mark documents as voided on our platform (when an admin tries to void one as before) if the document is more than 120 days old; which should match the voided status in Docusign too.

From Docusign:

> By default, envelopes expire 120 days after sending. When an envelope
> expires, the status changes to Voided and it can no longer be viewed or
> signed by recipients.

https://support.docusign.com/s/document-item?language=en_US#d10061e238&bundleId=gbo1643332197980&topicId=kyt1578456367748.html&_LANG=enus

### DIFF
--- a/app/services/docusign/api_client.rb
+++ b/app/services/docusign/api_client.rb
@@ -51,8 +51,9 @@ module Docusign
 
       if response.success? || document.expired?
         document.update(
-          active: false,
-          voided_at: Time.now
+          status: "voided",
+          voided_at: Time.now,
+          active: false
         )
 
         Result.new(success?: true)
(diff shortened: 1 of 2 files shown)


viviancan (approved):
Did not try locally, but LGTM! Nice fix and catch on the status 💯

Issue #5995, opened 2026-01-07
Display parent phone number on admin page


dboyer:
If a student adds a phone number for the parent consent, we should display that phone number on the student's admin page as we do for email.

<img width="444" height="285" alt="Image" src="https://github.com/user-attachments/assets/b7d9d5f5-079b-4d33-91e0-fc39a68aff32" />

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/gh/Iridescent-CM/technovation-app/5995?planning-poker) with Zenhub @shaun-technovation @viviancan

Issue #5994, opened 2026-01-07
Display phone number on parental consent tab


dboyer:
If the student has added a phone number on the parental consent tab, we should display it after the consent is signed similarly to how we display the parent email address.

<img width="855" height="458" alt="Image" src="https://github.com/user-attachments/assets/10171eb7-31f2-4b2f-ae43-5df58d8ae1e6" />

viviancan:
Included in this [PR](https://github.com/Iridescent-CM/technovation-app/pull/6000)

Issue #5922, opened 2025-11-06
Send a copy of signed forms via text if parents sign via text


dboyer:
After guardians sign the forms, they currently receive a copy via email. If parents sign via the link in a text message, they should receive a link to the forms in a text message as well.

Need to explore if these documents are generated on the fly or if there is a static evergreen link we can use.

PR #6000, opened 2026-01-09
Added logic to send signed consent form text message & link after consent forms are signed via text message

viviancan:
Refs #5922 

Also includes two frontend changes #5994, #5995

This PR updates the following:
After parent/guardian signs the consent forms via a text message link, they will receive a text message confirmation with a link to view their signed consent forms

- Added `source` param to track whether the consent request came from text message
- Parent confirmations now sent via text if they signed from a text message link. The original email confirmation stays the same if they signed via email link
- Created a new view to display signed consent forms
- Moved text message and email notification logic to controllers (away from model callbacks) so that the source param can be checked
- Added `SendSignedConsentTextMessageJob` job to send the signed consent text message and Updated twilio api client 
- A few front end updates

Flow
- Parent/guardian gets text with link `?source=text_message`
- Source param passed through hidden form field with they sign
- Controller checks for the `text_message` param and sends email or text message notification
- Text message contains link to `/signed_consents/:token` to view what they signed

### DIFF
--- a/spec/controllers/parental_consents_controller_spec.rb
+++ b/spec/controllers/parental_consents_controller_spec.rb
@@ -48,8 +48,7 @@ RSpec.describe ParentalConsentsController do
                               )}
 
       mail = ActionMailer::Base.deliveries[-2]
-      expect(mail).to be_present,
-                      "no notice to student about parental consent was sent"
+      expect(mail).to be_present, "no notice to student about parental consent was sent"
 
       expect(mail.to).to eq([student.email])
       expect(mail.subject).to eq(


app/controllers/parental_consents_controller.rb
```diff
@@ -42,7 +42,13 @@ def update
     @parental_consent = student.parental_consent
 
     if @parental_consent.update(parental_consent_params)
-      redirect_to edit_media_consent_path(token: token)
+      if params[:source] == "text_message" && student.parent_guardian_phone_number.present?
+        SendSignedConsentTextMessageJob.perform_later(account_id: student.account.id, consent_type: :parental)
```
shaun-technovation: Hmm, it might be more inline/consistent w/ our existing patterns to have a separate job for the parental consent and media consent...I'm not sure we combine them elsewhere...? 🤔 Like below the mailer is "parental consent finished" (not just "consent finished"...)

app/models/parental_consent.rb
```diff
@@ -95,9 +95,9 @@ def after_signed_parent_actions
       )
     end
 
-    ParentMailer.confirm_parental_consent_finished(student_profile.id).deliver_later
-
-    if Rails.env.production? && (student_profile.division.junior? || student_profile.division.senior?)
+    if Rails.env.production? &&
+       (student_profile.division.junior? || student_profile.division.senior?) &&
+       student_profile.parent_guardian_email.present?
```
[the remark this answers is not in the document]
shaun-technovation: Nice, good catch!

app/controllers/parental_consents_controller.rb
```diff
@@ -59,6 +65,7 @@ def student(opts = {})
   def parental_consent_params
     params.require(:parental_consent).permit(
       :student_profile_consent_token,
+      :source,
```
shaun-technovation: I'm wondering if a more intuitive name (if I'm understanding "source" correctly) would be "delivery method" (seems like it would be text message or email?) 🤔

spec/controllers/parental_consents_controller_spec.rb
```diff
@@ -116,19 +116,46 @@
...
         ParentalConsent.nonvoid.count
-      }
+      })
 
       expect(response).to redirect_to(
         parental_consent_path(student.parental_consent)
       )
     end
+
+    context "when the parent/guardian signs the parental consent via a text message link" do
```
shaun-technovation: 👍

PR #6027, opened 2026-02-03
Add a maximum amount of text messages allowed for parental consent forms 

viviancan:
Refs #6027

This PR adds a maximum amount of text messages allowed for parental consent forms.  

This PR adds the following:
- Env variable `MAXIMUM_NUMBER_OF_PARENTAL_CONSENT_TEXT_MESSAGES`
- A resend button for students who entered a phone number
- Limits the total number of text messages that can be send set by the new env variable. This value is 5, which is 1 original request + 4 additional requests (resends)
- Added limiting methods to `StudentProfile` model
- Added tests

**A note about what counts towards the limit
Only WhatsApp messages count toward the limit because every send attempt starts with WhatsApp first. If WhatsApp fails, the system automatically falls back to SMS. This fallback should not count against the user's limit since they only clicked send once.


app/models/student_profile.rb
```diff
@@ -388,6 +388,21 @@ def in_parental_consent_text_message_country?
...
+  end
+
+  def parental_consent_text_message_limit_reached?
+    parent_guardian_phone_number.present? &&
+      parental_consent_text_messages_remaining == 0
+  end
+
+  def parental_consent_text_messages_remaining
+    ENV.fetch("MAXIMUM_NUMBER_OF_PARENTAL_CONSENT_TEXT_MESSAGES").to_i - account.text_messages.current.parental_consent.whatsapp.count
+  end
```
shaun-technovation: Given that these methods^ start with or contain "parental consent" it might make more sense to move them to the parental consent model. It's seems like parental consent logic might make more sense there vs the student profile. In most apps (ours included) the main/core models tend to get overloaded with agrab bag of functionality 😅... Anywho, definitely not a blocker 😄

shaun-technovation (approved):
LGTM!

I was pondering over the naming of "...text_messages" count vs "text_message_request" count; like are the messages actually being sent, or are they just requests; and they can "request" a text message be sent up to five times...? Or maybe adding "sent" (or something) to perhaps give it more clarity? 🤷

Issue #6025, opened 2026-02-01
RPE and SMS testing feedback


dboyer:
Running list of feedback from testing RPEs.

- [x] In the section where ChAs add judges to an event, add a note stating "Clicking the "Add" button will send an email to the judge with details about this event." (VC)

<img width="365" height="276" alt="Image" src="https://github.com/user-attachments/assets/61a9ce83-8e9d-43a5-958e-1ea00ed69ce3" />

- [x] On the event page a ChA sees, add a note next to official event saying "Technovation staff will mark events as official if all event requirements are met." (Vc)

<img width="433" height="213" alt="Image" src="https://github.com/user-attachments/assets/d85bec49-0d85-4455-8154-2a041e6cd8b9" />

- [x] For the downloads available to ChAs from the page for a single event (i.e. /chapter_ambassador/events/523#/) - change file names for judge and team downloads to have Teams and Judges as the first word in the file name instead of the last (vc)

- [ ] For teams download for ChAs (i.e. downloaded from /chapter_ambassador/events/523#/), remove state and country columns and add a link to the submissions URL (i.e. /chapter_ambassador/team_submissions/us-project-9-by-us-team-9#/)

- [ ] Add a chapter filter to the ChA view of the events admin page (/chapter_ambassador/events_list#/)

- [x] Error when clicking the judge name on the events page for a ChA. Looks like there's a dot instead of a slash in the URL- /chapter_ambassador/participants.6412

<img width="269" height="331" alt="Image" src="https://github.com/user-attachments/assets/43dea048-97c0-4ff5-a70d-86316a7b20a0" />

- [ ] Judge details page is still erroring out even with the URL fix from item above

- [x] Adjust text where students enter a parent phone number. I've included the adjusted text below.

Parental consent forms are automatically sent to an email address. Add your parent or guardian's phone number here if you would like the forms to be sent via WhatsApp or text message. Please get permission from your parent or guardian to share their phone number before you add it here. 

By providing the phone number, you agree that your parent or guardian may receive text messages from Technovation. We will automatically try to send the message via WhatsApp first and then use SMS if WhatsApp is not available. Note: Standard messaging rates apply.

viviancan:
Addressed the url error in #5999
> Error when clicking the judge name on the events page for a ChA. Looks like there's a dot instead of a slash in the URL- /chapter_ambassador/participants.6412

We still need to review the query/scope because now I believe it is erroring out due to the by_chapterable scope

Issue #5999, opened 2026-01-07
Add field to invite a judge to an event via email


dboyer:
In the current RPE system, ChAs can add a person's email address to invite them to sign up as a judge for an RPE. We should include this functionality on the new RPE page - maybe as a separate "invite" section.

<img width="1143" height="178" alt="Image" src="https://github.com/user-attachments/assets/c608c41e-b499-4b82-aefb-b3b06cee9571" />

<img width="1143" height="220" alt="Image" src="https://github.com/user-attachments/assets/564773c7-d9d8-43a0-b969-4ddfdae0e0b7" />

viviancan:
This is on QA

dboyer:
Tested on QA and confirmed:

- I see the option to invite a judge
- I clicked and was able to enter an email address. I received an invite to that email address.
- The "invited judge" appeared on the list of judges assigned to the event with a status of "must sign up"
- When I added an email address that was already in the system, I received a notification that the account was already in the system
- When I completed all the judge onboarding tasks, the judge was updated to "ready" status

PR #6030, opened 2026-02-09
Add "Invite New Judge" to RPE details page with modal

viviancan:
Refs #5999 

This PR adds the "Invite New Judge" button to the RPE details page which opens the form in the new modal integration

Includes the following:
- Added `EventJudgeInvitationsController` with `new` and `create` actions
- Used the new SweetAlert + Turbo work completed in assigning judges to teams in event
- Updated the `registered_judges` partial to use `judge_list` which is a combination of judges and user invitations 
- Addressed the URL issue when clicking on judge name. The path contained a `.`(listed in #6025). We still need to look at the query because it is still erroring out. I beleive its due to the `by_chapterable` scope. I will add it to 6025

### DIFF
--- a/app/views/admin/regional_pitch_events/_registered_judge.html.erb
+++ b/app/views/admin/regional_pitch_events/_registered_judge.html.erb
@@ -1,11 +1,11 @@
 <tr id="<%= dom_id(judge, :registered) %>">
   <td>
-    <% if judge.is_a?(JudgeProfile) %>
+    <% if judge.pending_registration? %>
+      <%= judge.name.presence || "Invited Judge" %>
+    <% else %>
       <%= link_to judge.name,
       send("#{current_scope}_participant_path", judge.account.id),
       data: { turbo: false } %>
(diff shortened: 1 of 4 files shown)


app/views/admin/regional_pitch_events/_registered_judge.html.erb
```diff
@@ -1,8 +1,12 @@
 <tr id="<%= dom_id(judge, :registered) %>">
   <td>
-    <%= link_to judge.name,
-      send("#{current_scope}_participants_path", judge.account.id),
+    <% if judge.is_a?(JudgeProfile) %>
```
shaun-technovation: This would be nice to put at the judge level and name it something like `registered?` or `on_platform?` or maube the inverse `invited?` or `pending_registration?`.... 🤔

viviancan: Great idea 👍

PR #6033, opened 2026-02-11
Proof of concept for consent forms in Spanish

shaun-technovation:


app/controllers/application_controller.rb
```diff
@@ -136,4 +138,14 @@ def current_scope
...
   end
+
+  private
+
+  def set_locale_from_url_parameter
+    I18n.locale = params[:lang].presence_in(I18n.available_locales.map(&:to_s)) || :en
+  end
+
+  def default_url_options
+    {lang: I18n.locale}
```
shaun-technovation: This will make it so that all pages will have the lang URL parameter if it's present.

This was copied straight from the Rails Guide on I18n:
https://guides.rubyonrails.org/i18n.html#setting-the-locale-from-url-params

PR #6037, opened 2026-02-14
Rake task to remove accounts from the unaffiliated list

shaun-technovation:
This will add a new rake task that will remove accounts (by email address) from the unaffiliated list.

This rake task is needed b/c we have another rake task that assigns students to chapters, but that rake task doesn't remove them from the unaffiliated list.

```
 bundle exec rails remove_accounts_from_unaffilated_list[email address here]
```

### DIFF
--- a/lib/tasks/assign_students_to_chapters.rake
+++ b/lib/tasks/assign_students_to_chapters.rake
@@ -12,7 +12,9 @@ task assign_students_to_chapters: :environment do |task, args|
       puts "#{email_address} is not a student"
     elsif !account.current_season?
       puts "#{email_address} is not registered to the current season"
-    elsif account.chapterable_assignments.present?
+    elsif account.current_chapterable_assignments.present?
+      account.update_column(:no_chapterable_selected, nil)
+
       puts "#{email_address} is already assigned to a chapter"
     else
       account.student_profile.chapterable_assignments.create(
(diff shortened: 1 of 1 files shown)


lib/tasks/remove_accounts_from_unaffilated_list.rake
```diff
@@ -0,0 +1,18 @@
...
+task remove_accounts_from_unaffilated_list: :environment do |task, args|
+  puts "Removing accounts from the unaffilated list"
+
+  args.extras.each do |email_address|
+    account = Account.find_by(email: email_address)
+
+    if account.blank?
+      puts "#{email_address} could not be found"
+    elsif account.no_chapterable_selected?
+      account.update_column(:no_chapterable_selected, nil)
```
shaun-technovation: `update_column` won't send a "you've been assigned to xyz chapter" email, but `update` would.... I guess we can decide if we want emails sent out or not...

lib/tasks/assign_students_to_chapters.rake
```diff
@@ -22,6 +24,8 @@ task assign_students_to_chapters: :environment do |task, args|
         primary: true
       )
 
+      account.update_column(:no_chapterable_selected, nil)
+
```
shaun-technovation: I made a couple of changes after you looked at this @viviancan ^^^^

Removing students from the unaffiliated list if they're already assigned to a chapter and after they've been assigned to a chapter

Looking at the current season's chapter assignments (instead of all of them)

Issue #6039, opened 2026-02-18
RPE content setting for adding teams not enforced on new events page


dboyer:
The content settings were as follows.

<img width="444" height="200" alt="Image" src="https://github.com/user-attachments/assets/f5b9c850-e5d6-42cf-a3d5-db91d9445453" />

As a ChA, I was able to create events. In both the new and old events interface, I was able to create events as expected. However, in the new interface (/chapter_ambassador/events_list#/), I was able to add teams to an event even thought at setting was not turned on. In the old interface, I received a message that adding teams was not available. 

This setting also applies to adding judges. PT also confirmed that it should apply to bulk uploads of judges and teams.

PR #6051, opened 2026-02-20
Enforce RPE content and settings for updated RPE process

viviancan:
Refs #6039 

This PR enforces the RPE content and settings for adding teams/judges and creating events


app/controllers/chapter_ambassador/regional_pitch_events_controller.rb
```diff
@@ -6,6 +6,7 @@ class RegionalPitchEventsController < ChapterAmbassadorController
     include RegionalPitchEvents::BulkAddJudgesToRegionalPitchEvent
     include RegionalPitchEvents::BulkAddTeamsToRegionalPitchEvent
 
+    before_action :require_create_regional_pitch_event_enabled, only: [:new, :create]
```
shaun-technovation: Hmm, one "require" is an `include`/concern and the other isn't; seems like they both could be a concern to be consistent... 🤔

viviancan: Will update!

PR #6060, opened 2026-03-03
Add safe navigation to handle nil country in geocoding results

viviancan:
This method errors out if the country is blank. This most recently occurred with the `Jammu & Kashmir` location. This change will allow an error to be returned to the user instead of just the form. 
Screenshot 2026-03-03 at 11 54 08 AM


shaun-technovation (approved):
👍

Issue #6058, opened 2026-03-02
Additional columns on RPE page for ChAs


dboyer:
On the ChA RPE page, it would be helpful for ChAs to more easily see at a glance which teams have judges assigned and vis versa. We can add two columns to help with this.

- [x] On the teams grid, add an "assigned judges" column with a count of how many judges were assigned

<img width="1092" height="346" alt="Image" src="https://github.com/user-attachments/assets/481b97ce-8742-4a36-9457-3c1b3ad58065" />

- [x] On the judges grid, add an "assigned teams" column with a count of how many teams were assigned

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/gh/Iridescent-CM/technovation-app/6058?planning-poker) with Zenhub @shaun-technovation @viviancan

dboyer:
Confirming on QA:

- The judge and team count columns are visible on the admin RPE pages
- The same columns are visible on the ChA RPE pages

@viviancan I no longer see the button to manage judges. Is there some place else where I should do that?

<img width="1092" height="361" alt="Image" src="https://github.com/user-attachments/assets/e73c9dc2-cf16-4807-85ac-27a4d602474e" />

viviancan:
This is in code review, but these are things to test once on QA
- [x] Below threshold, no assignments → no count columns
- [x] Below threshold, no assignments → add teams to reach threshold → both columns appear
- [x] At/above threshold, no assignments → both columns visible
- [x] At/above threshold, no assignments → assign 1 judge to 1 team → both columns visible, counts update
- [x] At/above threshold, has assignments → remove team with no assignment (still at threshold) → both columns
visible, counts unchanged
- [x] At/above threshold, has assignments → remove team with an assignment (still at threshold) → both columns
visible, that team's row gone 
- [x] At/above threshold, has assignments → remove team, drops below threshold → both columns still visible
- [x] Below threshold, has assignments → remove team with assignments → both columns disappear
- [x] At/above threshold, no assignments → remove team, drops below threshold → both columns disappear

PR #6072, opened 2026-03-06
Add assigned judges and teams count columns to RPE show page for ChAs

viviancan:
Refs #6058 

This PR adds assigned judges and teams count columns to RPE show page for ChAs. It also uses turbo to update the "count" when adding or removing judges to a team in an event.


shaun-technovation (approved):
Cool, didn't try locally but LGTM!

PR #6117, opened 2026-03-31
Bug fix - Destroy judge team assignments when removing judge from event

viviancan:
Refs #6108 

When a judge was removed from an RPE, their team assignments in that event were not destroyed


shaun-technovation (approved):
👍

PR #6124, opened 2026-04-02
Allow judges to flag specific content items when recusing

viviancan:
Refs #6095 

This PR allows judges to flag specific content items when recusing

It includes the following:
- Adds `judge_recusal_flagged_contents` join table to track which specific content items a judge flags when recusing for "content doesn't belong to the team"
- Shows multi-select checkboxes in the recusal popup when that reason is selected. At least one content item must be checked to submit
- Flagged content are displayed in the admin portal


shaun-technovation (approved):
👍

Issue #6118, opened 2026-04-01
Allow RPEs to be set as closed


dboyer:
Request from a ChA

Some events should be "closed" meaning they do not appear on student dashboards as an option where the student can select to attend the event. For closed events, only teams assigned by the ChA see event details on their dashboard. 

Ideally, the chapter would like this to be a chapter setting, meaning some chapters could have the default be closed events and other chapters could have the default be open events. For the sake of making this available sooner rather than later, I think we could have just a check box on the details page for that event where the ChA checks the box if they would like it closed.

This task will cover:

- Setting a new flag in the database (name TBD) for the event
- Giving ChAs a box to select on the event admin page
- Adjusting the student dashboard to only show open/selectable events

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/gh/Iridescent-CM/technovation-app/6118?planning-poker) with Zenhub @shaun-technovation @viviancan

dboyer:
Chapter team said they thought "public/private" would be useful wording in many languages since open/closed could potentially be interpreted as the event is full rather than invite only. Can we switch to public/private rather than open/closed?

They also thought some help text would be useful. Could we add this near the checkbox on the event admin page?

Events are public by default and can be selected by teams from their dashboard. Please uncheck this box if you would like the event to be private. Only teams you add to the event will see event details.

viviancan:
this is on QA. I also added "event visibility" filter and datagrid on the events datagrid

dboyer:
Tested on QA and confirmed:

- As an admin, I see "event visibility" as an additional sortable column and as a filter. 
- I logged in as a student eligible for a senior division event and saw two events on my dashboard
- As a ChA, I set the status of one event to private
- As the student, I now saw only one event on the dashboard for me to select
- As the ChA, I could add teams to the private event
- As the student added to the private event, I saw that event on my dashboard
- As the ChA, I see "event visibility" as an additional sortable column and as a filter.

PR #6125, opened 2026-04-06
Allow RPEs to be set as "private" - they cannot be selected and do not appear on list of available events for students

viviancan:
Refs #6118 

Allow RPEs to set as "closed" - they cannot be selected and do not appear on list of available events for students

This PR adds the following:
- Adds `selectable` boolean to `RegionalPitchEvent` - default is true
- Closed events are excluded from the list of events available to students to choose from
- ChAs can update this value on the event form
- Added this to datagrids

Naming note
-  I started with "selectability" and different labels related to that, but I was unsure if it would translate globally. Ended up going with "availability" and open/closed (based on the ticket description). I am open to suggestions or waiting to update based on testing feedback

### DIFF
--- a/app/controllers/chapter_ambassador/regional_pitch_events_controller.rb
+++ b/app/controllers/chapter_ambassador/regional_pitch_events_controller.rb
@@ -172,6 +172,7 @@ module ChapterAmbassador
         :event_link,
         :capacity_enabled,
         :capacity,
+        :selectable,
         :division_id,
         division_ids: []
       )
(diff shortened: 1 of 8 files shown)


viviancan:
There were some text requests from VET so changing this back to a WIP

viviancan:
This is ready for review. Updated naming from open/closed to public/private

shaun-technovation (approved):
LGTM! 👍

PR #6126, opened 2026-04-07
Send judges an email when bulk-added to an event

shaun-technovation:
This will make it so that when judges get added to an event via bulk-uploading, the judges  will receive an event invite email.


viviancan (approved):
Did not try locally, but lgtm 👍

PR #6131, opened 2026-04-09
Add deleted scores count for judges

shaun-technovation:
Changes here are to add a "deleted scores count" for judges.


viviancan (approved):
Works as expected!! Lgtm!

Issue #6132, opened 2026-04-13
Students should be able to add Thunkable AI URLs


dboyer:
Currently if a student indicates they used Thunkable, they can only enter a URL that starts with https://x.thunkable.com/projectPage/....

They should also be able to enter a URL that starts with https://ai.thunkable.com/....

dboyer:
Tested on QA and confirmed:

- I could submit a link beginning with https://ai.thunkable.com
- I could also still submit a link beginning with https://x.thunkable.com/projectPage/

PR #6133, opened 2026-04-14
Add ai.thunkable.com as a valid Thunkable URL

viviancan:
refs #6132 
Add ai.thunkable.com as a valid Thunkable URL


shaun-technovation (approved):
👍 Thanks for taking care of this one!

PR #6138, opened 2026-04-15
Add "judge info" onboarding step to profile debugging section

shaun-technovation:
A pretty simple update here to add the new "judge info" onboarding step to the profile debugging section for both admins and chapter ambassadors.


viviancan (approved):
lgtm 👍

PR #6146, opened 2026-04-16
Rails 7.2 upgrade

shaun-technovation:


shaun-technovation:
Closing, given our current circumstances, I don't think we have enough time to ensure a successful (bug free) deployment for this.

PR #6147, opened 2026-04-17
Prevent judge-mentor accounts from being assigned to an RPE (via CSV bulk upload) if the mentor is on a team

shaun-technovation:
This should prevent judge/mentor combo accounts assigned to a team from being added to an RPE via CSV upload.


viviancan (approved):
LGTM!!

PR #6159, opened 2026-04-20
Only display distinct judges on judges datagrd

shaun-technovation:
This will add a `.distinct` clause to this query so only distinct judges will appear on the judges datagrid. This will fix the issue of the same judge appearing multiple times.

Most of our other datagrids are using `distinct` too, so whatever issue is happening here is happening in other datagrids too.


shaun-technovation:
I pulled down production data and verified this fix, since it's just a one line change, I'm going to get this one on QA so it can go out w our next deploy. 👍

Issue #6134, opened 2026-04-14
Mentors matched with teams should not appear as a judge option for RPEs


dboyer:
Mentors matched with a team should not appear in the list of judges that can be added to an RPE. This issue covers adding the judges individually. A separate ticket will cover updating the judge bulk upload to not include mentors matched with a team.

dboyer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Technovation-605271851956b1000ec2f24d/issues/gh/Iridescent-CM/technovation-app/6134?planning-poker) with Zenhub @shaun-technovation @viviancan

dboyer:
Confirmed on QA that a judge who was also a mentor matched with a team did not appear on the list of judges to be added, even when I searched specifically for their name. I was able to add other judges.

PR #6160, opened 2026-04-20
Exclude mentors matched with a team from RPE available judge list

viviancan:
Refs #6134 

Exclude mentors matched with a team from RPE available judge list


shaun-technovation (approved):
Feel free to merge whatever in, given our current circumstances.

Issue #6161, opened 2026-04-21
Bug: Error updating student profile if there is a parent guardian phone number


viviancan:

dboyer:
Tested on QA:

- Logged in as an admin and successfully edited data for a student with a parent phone number
- Parental and media consent status visible in onboarding section for student with a parent phone number

PR #6162, opened 2026-04-21
Fix student profile update bug due to parent guardian phone number

viviancan:
Refs #6161

This bug recently came to our attention when a student profile could not be updated due to the parent guardian phone number validation. Updated validaton to only run when phone number actually changed. Also added "onboarded" status to the student profile debugging view.


shaun-technovation (approved):
Feel free to merge whatever in, given our current circumstances.

PR #6167, opened 2026-04-23
Hide chapter filter on RPE datagrid for ChAs

shaun-technovation:
This will hide the chapter filter on the Events datagrid for ChAs, making it only available for admins


shaun-technovation:
I'm going to merge this one into QA.

PR #6173, opened 2026-04-28
Salesforce updates

shaun-technovation:
This will add phone number and contest rank to Salesforce.


shaun-technovation:
I don't have Salesforce hooked up locally, I'm going to get this one merged in so it can be tested on QA.

